PR #427, opened 2026-04-20
fix(billing): advisory lock in ensureStripeCustomer to close delayed-retry race

### LINKED ISSUE #394: fix(billing): prevent Stripe customer dual-write race in ensureStripeCustomer
Phase 1 — Operational safety. Tracked by §CR-8 (CR8-P1-04).

## Problem

`apps/api/src/routes/billing.ts:242-282` — `ensureStripeCustomer()` executes:

1. Create Stripe customer (with idempotency key)
2. Conditional DB update IF `stripe_customer_id IS NULL`
3. Re-read row to return the winning customer id

If step 1 succeeds and step 2 fails (Neon HTTP driver transient, serverless timeout, concurrent write losing the race), a later retry can either produce a second Stripe customer under a different idempotency scope or leave the DB pointing to an id that does not match the one used on subsequent billing flows. Both cases result in duplicate Stripe customers and billing-state divergence — and a customer billed against the wrong record.

NeonDB HTTP driver does not support multi-statement transactions, so the fix is non-trivial.

## Acceptance criteria

Pick one of:

- **(a) Advisory lock**: open a direct Postgres connection (not the HTTP driver) for the critical section, take `pg_advisory_xact_lock(hashtext('stripe:ensure:' || user.id))`, run Stripe create + DB update inside the lock, release on commit
- **(b) Reconciliation cron**: detect duplicate Stripe customers sharing `metadata.user_id`, merge subscriptions onto the oldest, delete the duplicates, log merges for audit

Either approach must:

- [ ] Make `ensureStripeCustomer` idempotent under concurrent invocation from the same user
- [ ] Converge partial-failure state within one cycle (cron) or one retry (lock)
- [ ] Tests: concurrent calls → exactly one Stripe customer; partial-failure retry → converges

## Citations

- `apps/api/src/routes/billing.ts:242-282`
- §CR-8 CR8-P1-04 in `.jv/docs/MASTER_PLAN.md`

RevealUIStudio:
Closes #394.

## Summary

Fixes a delayed-retry race in `ensureStripeCustomer()` that could produce two Stripe customers for one user.

## The race

`apps/api/src/routes/billing.ts` `ensureStripeCustomer()` previously defended concurrent-request races via:

1. Stripe idempotency key `create-customer-${userId}` — dedupes simultaneous customer creations
2. Conditional `UPDATE ... WHERE stripe_customer_id IS NULL` — ensures only one DB row write wins

That handled co-requested races. It did **not** handle delayed-retry races:

- Request creates Stripe customer, DB write fails.
- >24 hours later, another call retries `ensureStripeCustomer`. Stripe idempotency-key TTL is ~24h; the retry creates a **new** customer.
- Two Stripe customers per user, billing state splits.

## The fix

Per-user Postgres advisory lock around the read→create→write critical section:

```ts
await client.query('BEGIN');
await client.query(
  `SELECT pg_advisory_xact_lock(hashtext($1))`,
  [`stripe:ensure:${userId}`],
);

// Re-read inside the lock — another request may have won while we waited.
const alreadyCreated = /* SELECT stripe_customer_id FROM users WHERE id = $1 */;
if (alreadyCreated) { COMMIT; return alreadyCreated; }

// Hold lock; create + write atomically.
const customer = await protectedStripe.customers.create(...);
await client.query('UPDATE users SET stripe_customer_id = $1, updated_at = NOW() WHERE id = $2', [customer.id, userId]);

await client.query('COMMIT');
```

`pg_advisory_xact_lock` auto-releases on COMMIT/ROLLBACK. Requires a real pg connection (NeonDB HTTP driver doesn't support transactions); uses `getRestPool()` which [#390](https://github.com/RevealUIStudio/revealui/pull/390) exposed for exactly this kind of path.

Stripe idempotency key is retained as belt-and-suspenders for the vanishingly-rare case where the lock is lost (replica failover) and a concurrent attempt still sneaks through within the 24h window.

## Fallback path

When `getRestPool()` returns null (build-time contexts with no `DATABASE_URL`), falls back to the prior conditional-UPDATE best-effort pattern and logs a warning. Production always has the pool; the fallback keeps build-time green.

## Test coverage

- **Existing billing test suite** — all 361 tests pass. Both test files (`billing.test.ts`, `billing-comprehensive.test.ts`) extended their `@revealui/db` mock to include `getRestPool: vi.fn(() => null)` so the existing tests exercise the fallback path (which preserves prior behavior — same conditional-UPDATE flow they were written against).
- **Advisory-lock path unit test** — not included in this PR. Mocking `pool.connect()` + `client.query()` per SQL step would verify the happy path but wouldn't catch the actual failure modes (real pg lock semantics, 24h-TTL edge case). Integration tests against a real pg instance are the appropriate coverage location — natural follow-up alongside existing tests in `packages/test/src/integration/database/`.

## Note on the commit

This commit (`efd2a3556`) also appears in [#426](https://github.com/RevealUIStudio/revealui/pull/426)'s branch history — it got inadvertently included during a cross-session branch-switch. Either this PR or #426 will ship the fix; don't merge both with the same commit. Recommending this PR for clean scope and #426 for its intended shared-memory-RPC scope; dropping the commit from #426 via rebase is trivial when the time comes.

## Related

- MASTER_PLAN §CR-8 Phase 1 item CR8-P1-04
- Depends on [#390](https://github.com/RevealUIStudio/revealui/pull/390) (shared pg.Pool primitive, merged)

### DIFF
--- a/.changeset/api-stripe-customer-race.md
+++ b/.changeset/api-stripe-customer-race.md
@@ -1,0 +1,13 @@
+---
+'api': patch
+---
+
+Fix Stripe customer dual-write race in `ensureStripeCustomer()` (closes #394).
+
+The prior implementation protected against concurrent-request races via Stripe's idempotency-key mechanism (`create-customer-${userId}`) plus a conditional `UPDATE WHERE stripe_customer_id IS NULL`. That worked for simultaneous requests.
+
+What it didn't defend against: **delayed-retry races**. Stripe idempotency keys have a ~24-hour TTL. If a request succeeded in creating a Stripe customer but the subsequent DB write failed, and then >24h later another call retried, the Stripe customer creation would produce a *new* customer (idempotency expired) rather than returning the original — resulting in two Stripe customers for one user and split billing state.
+
+This change adds a Postgres advisory lock scoped per-user (`pg_advisory_xact_lock(hashtext('stripe:ensure:' || userId))`) around the read→create→write critical section. Within the lock, the function re-reads the DB to catch any customer that was written by another path during the wait, and only invokes Stripe when no customer exists. The lock requires a real pg connection, which #390's shared `getRestPool()` primitive makes available from this billing path (NeonDB HTTP driver doesn't support transactions).
+
+When the pool is unavailable (build-time contexts with no `DATABASE_URL`), the function falls back to the previous conditional-UPDATE best-effort pattern and logs a warning so any production fallback is visible.
--- a/apps/api/src/routes/__tests__/billing-comprehensive.test.ts
+++ b/apps/api/src/routes/__tests__/billing-comprehensive.test.ts
@@ -177,6 +177,10 @@ const mockDb = { select: vi.fn(), update: vi.fn(), transaction: vi.fn() };
 
 vi.mock('@revealui/db', () => ({
   getClient: vi.fn(() => mockDb),
+  // Null forces ensureStripeCustomer onto the conditional-UPDATE fallback
+  // path (same behavior as before #394's advisory-lock addition), which is
+  // what these tests were written against.
+  getRestPool: vi.fn(() => null),
 }));
 
 vi.mock('@revealui/core/observability/logger', () => ({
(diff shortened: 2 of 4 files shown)


RevealUIStudio:
Redundant — identical fix already on `origin/test` as commit [`efd2a3556`](https://github.com/RevealUIStudio/revealui/commit/efd2a3556e1a723303e4b378fa7314e03ece650d), shipped via merged #426. Verified byte-identical to this PR's commit (`f5622aa`) — only the commit SHA differs (different parent, same patch). Closes #394 via the commit on test; no separate merge needed. Branch `fix/stripe-ensure-customer-race` can be deleted.

PR #866, opened 2026-05-15
fix(server): eliminate flaky PGlite test hook timeout

RevealUIStudio:
## Summary

`metering.test.ts` and `mcp-usage.test.ts` called `createTestDb()` in `beforeEach` - a 9-19s PGlite bootstrap + full migration run **per test** (4x and 9x per file). Under concurrent system load the hook blew past vitest's 10s default and failed the whole file (`Hook timed out in 10000ms`). Surfaced during the zod 4.4.3 upgrade (#862); not zod-related - pre-existing test fragility.

Audit found two compounding root causes, both fixed here:

- **Config:** `apps/server/vitest.config.ts` set `testTimeout` but no `hookTimeout`, so it fell back to vitest's 10s default instead of the root `vitest.config.ts`'s `30000`. The server config is standalone (no `extends`), so it never inherited it. Added `hookTimeout: 30000`.
- **Pattern:** both files now bootstrap PGlite once in `beforeAll` and clean tables in `afterEach` via the existing, tested `cleanTables()` helper - the same pattern the 3 sibling PGlite files in `routes/__tests__/` already use. 14 of 16 vitest `createTestDb` consumers were already on `beforeAll`; only these 2 deviated (classified accidental).

`createTestDb()` per-call cost (9-19s) is unchanged - a shared/template PGlite instance (dump/restore) is a tracked follow-up; it touches a helper used by 18 test files across 4 packages and warrants its own benchmarked PR.

## Test plan

- [x] `pnpm --filter server test` - **139 files / 2726 tests pass**, 1 intentional skip
- [x] Converted files each bootstrap PGlite exactly once (was 4x/9x): `metering.test.ts` 27.2s, `mcp-usage.test.ts` 24.9s file totals - within the 30s `hookTimeout`
- [x] Pre-push quality gate green


RevealUIStudio:
Follow-up tracked: #867 - createTestDb() per-call speedup via a shared PGlite dump/restore template (deferred from this PR; 18-consumer blast radius across server/test/ai/auth).

PR #1057, opened 2026-05-23
fix(admin): seed first admin at boot on self-hosted Forge kits

RevealUIStudio:
## Summary

A fresh self-hosted **RevealUI Fleet** (Forge) kit booted healthy — Postgres, API, and Admin all up, license valid, REST serving — but **could never be logged into**: the `users` table was empty and there was no way to create the first admin. Sign-up is closed in fleet mode, so the engine's `onInit` first-admin seeder is the only intended path, and it wasn't firing.

## Root cause

`onInit` runs only when `RevealUIInstance` treats the process as runtime, not build-time:

```
isBuildTime = NEXT_PHASE === 'phase-production-build'
           || (NODE_ENV === 'production' && !RUNTIME_INIT)
```

A Forge kit runs `NODE_ENV=production` and sets no `RUNTIME_INIT`, so `isBuildTime` is `true` at runtime → `onInit` is skipped → the admin is never seeded. A second layer: even with `RUNTIME_INIT` set, the engine initialized lazily on the first request that needed it, which proved unreliable for seeding.

## Fix

- **`RUNTIME_INIT=1`** in the runtime stage of `apps/admin/Dockerfile.forge` (and `apps/server/Dockerfile.forge` for parity). `next build` / `turbo build` set `NEXT_PHASE`, so the **build** still correctly skips `onInit`; only the running container flips out of the build-time guard.
- **Deterministic boot-time engine init** from `apps/admin/src/instrumentation.ts`, replacing reliance on the lazy first-request init.
  - The engine (`getRevealUIInstance` → `@revealui/core/nextjs` → `@revealui/db`) is Node-only, so it lives in a new `apps/admin/src/instrumentation-node.ts` imported **only** behind `NEXT_RUNTIME === 'nodejs'`. This keeps `@revealui/db` out of the Edge bundle — the same reason the telemetry transport in `instrumentation.ts` uses `fetch()` instead of importing `@revealui/db` directly.
  - Gated on `RUNTIME_INIT`, so **hosted/Vercel boot is unchanged** (hosted seeds via `account_entitlements`, not `onInit`, and shouldn't pay an eager engine init per cold start).
  - Idempotent (`onInit` checks `users === 0`) and never throws (honors the Next.js instrumentation "never throw" contract).

## Verification

- `pnpm --filter admin typecheck` ✅
- New unit test `apps/admin/src/__tests__/instrumentation-node.test.ts` — 3 cases: initializes when `RUNTIME_INIT` set, no-op when unset, never throws when the engine init fails ✅
- **Fresh-kit boot** (rebuild the forge admin image → `docker compose down -v && up` → assert exactly one super-admin row + headless `POST /api/auth/sign-in` → 200) — to be confirmed on this branch's image before merge.

## Scope

Behavior change is confined to self-hosted Forge kits (`RUNTIME_INIT=1`). Hosted SaaS boot is untouched.

### DIFF
--- a/apps/admin/Dockerfile.forge
+++ b/apps/admin/Dockerfile.forge
@@ -67,6 +67,13 @@ ENV NODE_ENV=production
 ENV NEXT_TELEMETRY_DISABLED=1
 ENV PORT=3000
 ENV HOSTNAME="0.0.0.0"
+# GAP-214: run the engine's onInit (first-admin seeder) at runtime, not build.
+# `next build` sets NEXT_PHASE so the build still correctly skips onInit; this
+# only flips the *running* container out of RevealUIInstance's build-time guard
+# (NODE_ENV=production && !RUNTIME_INIT). Without it a fresh kit boots healthy
+# but never seeds an admin, so login is impossible (signup is closed in fleet
+# mode). Paired with the deterministic boot init in src/instrumentation-node.ts.
+ENV RUNTIME_INIT=1
 
 WORKDIR /app
 
--- a/apps/admin/src/__tests__/instrumentation-node.test.ts
+++ b/apps/admin/src/__tests__/instrumentation-node.test.ts
@@ -1,0 +1,68 @@
+/**
+ * GAP-214 regression guard for the deterministic boot-time engine init.
+ *
+ * Protects three properties of `initEngineAtBoot` (apps/admin/src/instrumentation-node.ts):
+ *  1. On a Forge kit (RUNTIME_INIT set) it initializes the engine at boot, which
+ *     is what runs config.onInit → the first-admin seeder. Without this a fresh
+ *     kit boots healthy but can never be logged into.
+ *  2. On hosted boot (RUNTIME_INIT unset) it does nothing — hosted seeds via
+ *     account_entitlements and must not pay an eager engine init at every cold start.
+ *  3. It never throws, honoring Next.js's instrumentation "never throw" contract.
+ */
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { getRevealUIInstance } = vi.hoisted(() => ({ getRevealUIInstance: vi.fn() }));
+
+vi.mock('@/lib/utilities/revealui-singleton', () => ({ getRevealUIInstance }));
+
+async function loadInitEngineAtBoot() {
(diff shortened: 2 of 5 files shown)


RevealUIStudio (commented):
Reviewed the code + CI — the fix looks correct and well-scoped. Not approving yet only because of the runtime gate noted at the end, not because of anything in the diff.

**Root cause is right.** `onInit` (the first-admin seeder) only runs when `RevealUIInstance` treats the process as runtime rather than build-time, i.e. `isBuildTime` is `NEXT_PHASE === 'phase-production-build' || (NODE_ENV === 'production' && !RUNTIME_INIT)`. A self-hosted production kit sets neither `RUNTIME_INIT` nor a build phase at run time, so `isBuildTime` is true and the seeder is skipped — and with signup closed in fleet mode there's no fallback to a first admin.

**The approach is clean:**
- `RUNTIME_INIT=1` in both `apps/admin/Dockerfile.forge` and `apps/server/Dockerfile.forge` flips the *running* container out of build-time mode, while `next build` / `turbo build` still correctly skip `onInit` (neither sets `RUNTIME_INIT`).
- Putting the eager engine init in a dedicated `apps/admin/src/instrumentation-node.ts`, imported only behind `NEXT_RUNTIME === 'nodejs'`, is the right move — it keeps the Node-only engine (and `@revealui/db`) out of the Edge bundle, consistent with why the telemetry transport in `instrumentation.ts` uses `fetch()` rather than importing `@revealui/db` directly.
- Gating the eager init on `RUNTIME_INIT` keeps hosted/dev cold-start unchanged (no eager engine init there). The dispatch is wrapped so it can't throw out of instrumentation, and idempotency rides on `onInit`'s existing `users === 0` check. The 3-case test (set / unset / throws) covers the contract.

**One gate before merge — runtime, not code.** The unit + integration suites run on PGlite, so they prove the dispatch logic but not that a real production image actually seeds at container boot. Worth an end-to-end check on an image built from this branch: wipe the DB volume, boot fresh, confirm the `users` table has the seeded super-admin and a headless sign-in returns 200. That also exercises boot ordering — the engine init runs once at boot, so if the app comes up before the database is reachable, `onInit`'s connection-error path returns without seeding and the cached instance won't retry until the next container restart. Worth confirming the compose orders the app after a healthy database. Once that passes on a built image, this is good to merge.

PR #1061, opened 2026-05-24
chore(test): backflow main into test via merge commit (re-sync before promotion)

RevealUIStudio:
## What

Re-sync `main` into `test` via a **merge commit**, ahead of the next `test` → `main` promotion.

## Why

`main` currently has one commit `test` does not (`a7509e566`, the previous promotion merge), so `main` is not an ancestor of `test`. The prior backflow (#1059) was **squash-merged**, which produced an empty single-parent commit and did **not** bring `main`'s lineage into `test` — leaving `test` → `main` prone to a diverged / conflicting state.

Verified before opening:
- `git merge-base --is-ancestor origin/main origin/test` → **NO** (main is not an ancestor of test)
- divergence `main…test` = `1 ↔ 5` (the 1 is `a7509e566`)

## How

This branch points at `origin/main`. Merging it brings `main`'s history into `test`. No file changes — this is a lineage-only re-sync.

## ⚠️ Merge method

Merge via **Create a merge commit** — **NOT** squash. Squashing strips `main`'s parent and reintroduces the exact divergence this PR fixes.

PR #1063, opened 2026-05-24
chore(test): backflow main into test (merge-commit, re-sync before promotion)

RevealUIStudio:
## Re-sync `test` with `main`

Backflow after the `test → main` promotion. The merge commit (`main` → `test`) was produced and pushed by `backflow-main-into-test.yml`, but that workflow couldn't open this PR — `GITHUB_TOKEN` is blocked from creating PRs org-wide ("Allow GitHub Actions to create and approve pull requests" is off). Opening it manually here.

Content-identical to `main`; merge as a **merge commit** (not squash) to preserve parentage so the next `promotion-gate` passes.


RevealUIStudio:
Coordination: I closed my duplicate manual backflow #1061 in favor of this automation-driven one (canonical branch).

⚠️ **Please merge this via "Create a merge commit", not squash.** #1059 was squash-merged, which stripped `main`'s parent and produced the current `main` ↛ `test` divergence (`compare main...test` = `diverged`, `behind_by: 1` = `a7509e566`).

After this merges, `git merge-base --is-ancestor origin/main origin/test` should return true (main becomes an ancestor of test), so the next `test` → `main` promotion stays clean.

PR #1086, opened 2026-05-25
fix(admin): unblock admin login — CSRF-exempt pre-auth endpoints + recognize owner role

RevealUIStudio:
Two related fixes that unblock the admin login path (both surfaced while exercising the admin sign-in flow).

## 1. Exempt pre-auth endpoints from the CSRF gate
Adds `/api/auth/sign-in`, `/api/auth/sign-up`, and `/api/auth/passkey/authenticate*` to the admin proxy's `CSRF_EXEMPT_PREFIXES` (`apps/admin/src/proxy.ts`).

The proxy's CSRF gate fires only when a `revealui-session` cookie is **present**, and checks presence — not validity. So a **stale/invalid** session cookie (e.g. one left after a secret rotation invalidated old sessions) makes the gate demand an `X-CSRF-Token` the login form never sends → a spurious `CSRF token missing` 403 that blocks sign-in until cookies are cleared. `/api/auth/password-reset` is already exempt for the same reason. Login is already un-CSRF'd in the normal no-session case, so this lowers no security — it removes the stale-cookie trap. Post-auth endpoints (passkey **register**, change-password, sign-out) stay gated.

## 2. Recognize `owner` as an admin role
`apps/admin/src/app/api/auth/sign-in/route.ts` set `revealui-role: 'admin'` (which proxy.ts's `/admin` gate requires) only for DB role `admin`/`super-admin` — **not `owner`**. But the bootstrap assigns the first user `role: 'owner'` (the highest DB role; the enum is `owner|admin|editor|viewer|agent|contributor` — `super-admin` is an app-layer `roles[]` value, never the DB `role`). Net effect: the bootstrap's own owner is bounced to `/login`, unable to reach `/admin`.

The pre-fix allowlist `['admin','super-admin','admin','super-admin']` was the un-deduped result of the #306 role-rename (`user-admin → admin`, `user-super-admin → super-admin` collapsing onto existing entries). Now `['owner','admin','super-admin']` — deduped + `owner` recognized.

## Tests
Allowlist/config additions to already-tested mechanisms (CSRF token gen/validate is covered by `apps/admin/src/__tests__/utils/csrf-token.test.ts`). Follow-up: a middleware test asserting (a) pre-auth prefixes pass the CSRF gate even with a session cookie + no token, and (b) an `owner` session yields `revealui-role: 'admin'`.

## Verification
Reproduced both: a stale `revealui-session` cookie → `CSRF token missing` on the login POST; and an `owner` user → bounced to `/login` despite a 200 sign-in. Both flows clear with these changes.


RevealUIStudio:
Heads-up from a production diagnosis of the "CSRF token missing" on `/signup`: it traces to the same valid-session-on-auth-page condition this PR addresses. Two related follow-ups worth considering here (proxy-side root cause):

**#1 (proxy — the root fix): bounce already-authenticated users off `/login` + `/signup`.**
`PUBLIC_PATHS` in `apps/admin/src/proxy.ts` lists `/login` and `/signup`, but the auth gate only *adds* protection — it never redirects an *already-authenticated* user away from these pages. So a user holding a valid `revealui-session` can sit on `/signup`, and the signup POST then hits the CSRF gate. This PR exempts the endpoint (good), but the cleaner root fix is to not let an authed user reach signup/login at all: "if a valid session cookie is present and path ∈ {`/login`,`/signup`} → redirect to `/`". That also eliminates the condition without relying on the exemption.

**#3 (AdminBar — mislabeled, FYI): the "Exit Preview" bar shows for any authenticated user, not just preview mode.**
`apps/admin/src/lib/components/AdminBar/index.tsx` sets visibility purely from `/api/auth/me` (`setShow(Boolean(user?.id))`); the `preview`/draftMode prop is passed but never read. So a logged-in admin sees a "Dashboard / Exit Preview" bar even outside draft mode. Either gate the control on real draft mode or relabel.

I shipped the narrow piece — suppressing the bar on auth-flow pages — in #1098. Note #3 is in the **same file** as #1098, so to avoid a conflict it's best folded into #1098 or done as a follow-up rather than edited here. #1 is proxy-side and overlaps this PR, so flagging rather than touching `proxy.ts` directly.

PR #1163, opened 2026-05-29
fix(core,server): use heap_size_limit for memory health/alert checks

### LINKED ISSUE #951: API /health/ready intermittently degraded — Function instances hitting ~95% V8 heap
## Symptom

After the 2026-05-18 api redeploy (clean run [26031332892](https://github.com/RevealUIStudio/revealui/actions/runs/26031332892), all jobs green), `https://api.revealui.com/health/ready` returns HTTP 200 but with `status: degraded` because of the memory check. Sampled 3 consecutive probes:

| Probe | Container uptime | heapUsed | heapTotal | usedPct | Status |
|---|---|---|---|---|---|
| 1 | 4h 4m | 156 MB | ~244 MB | 63.98% | healthy |
| 2 | 5h 37m | 103 MB | ~108 MB | 95.44% | unhealthy |
| 3 | 7h 6m | 103 MB | ~110 MB | 93.33% | unhealthy |

Variance per probe = Vercel routes requests across multiple Function instances. Some instances are fine (probe 1), others are hitting ~95% heap utilization (probes 2, 3) and reporting memory `unhealthy`, which downgrades the overall ready check to `degraded`.

## Diagnosis suspicion

Probes 2 + 3 have **heapTotal ~108-110 MB** vs probe 1's ~244 MB. Node's V8 heap grows up to `--max-old-space-size` (default ~4 GB on most systems, but explicit configs can be lower). A heapTotal of ~108 MB implies either:

1. `--max-old-space-size` is set ~128 MB somewhere in the Vercel build chain (Dockerfile, `NODE_OPTIONS` env var, or framework default), and these instances are bumping against it.
2. Vercel's Function memory tier for `revealui-api` is small enough that V8 self-tunes the max heap downward.

If (1), `NODE_OPTIONS=--max-old-space-size=512` (or similar) on the Vercel project would lift the ceiling without changing the memory tier.

If (2), bumping the Vercel Function memory allocation (project settings) would let V8 grow the heap.

Either way the route to a fix is straightforward; the diagnosis question is "where's the current limit coming from."

## Reproduce

```bash
for i in 1 2 3 4 5; do
  curl -sS https://api.revealui.com/health/ready | jq '{
    mem: .checks.memory.status,
    pct: .checks.memory.details.usedPercent,
    heap_mb: (.checks.memory.details.heapUsed / 1024 / 1024 | floor),
    total_mb: (.checks.memory.details.heapTotal / 1024 / 1024 | floor),
    uptime_h: (.uptime / 3600 | floor)
  }'
  sleep 5
done
```

You should see a mix of `mem: "healthy"` and `mem: "unhealthy"` rows depending on which Function instance answers.

## Severity

- **Customer impact: low** — DB is healthy, smoke test passes, `/health/ready` returns 200, deploys aren't being rolled back. The `degraded` overall status doesn't trigger any current alerting.
- **Trust impact: low-medium** — if uptime monitors are configured to alert on `status != "healthy"` payload (rather than HTTP status), they'll fire intermittently on the `degraded` responses.
- **Real risk: latent** — a function instance at 95% heap is one allocation away from OOM. Hitting that during a real request returns 500, not 200.

## Out of scope (covered separately)

- #934 (DB credential rotation) — closed today, unrelated.
- The memory check itself — it's reporting accurately; the underlying memory state is the issue, not the check.

RevealUIStudio:
## Summary

`/health/ready` has been returning `status: degraded` in production (HTTP 200, memory check `unhealthy`) — re-verified still reproducing 2026-05-29 (8/8 probes degraded; see #951 for the full trace).

**Root cause:** the memory health check computed `heapUsed / heapTotal`. `heapTotal` is V8's *currently-committed* heap, not the ceiling — it tracks just above `heapUsed` and grows on demand toward `heap_size_limit` (`--max-old-space-size`). So the ratio sits at ~90%+ during normal operation and trips the 90% threshold. Live proof: `heapTotal` ~96 MB while `rss` ~229 MB — no actual memory pressure. False-positive degradation (the `createMemoryUsageAlert` rule at threshold 85 had the same flaw).

## Changes

- **`packages/core/src/observability/health-check.ts`** — `createMemoryHealthCheck` divides by `v8.getHeapStatistics().heap_size_limit` (the real OOM-proximity ceiling) instead of `heapTotal`. `heapLimit` added to reported `details` for transparency; `heapTotal` retained.
- **`apps/server/src/index.ts`** — the `createMemoryUsageAlert` rule uses the same `heap_size_limit` denominator.
- **`packages/core/src/__tests__/observability/health-check.test.ts`** — the degraded-band test mirrors the new denominator.

## Verification

- `@revealui/core` unit tests: **43/43 pass** (`health-check.test.ts`).
- `@revealui/core` typecheck: clean (`tsc --noEmit`).
- Biome: clean on all 3 changed files.
- Pre-push quality gate: **PASS** (all checks green).

Closes #951

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -20,6 +20,7 @@ if (process.env.SENTRY_DSN) {
 }
 
 import { readFileSync } from 'node:fs';
+import { getHeapStatistics } from 'node:v8';
 import { serve } from '@hono/node-server';
 import { initializeLicense } from '@revealui/core/license';
 import {
@@ -1224,7 +1225,7 @@ export function initAlerting(): void {
   alerting.registerRule(
     createMemoryUsageAlert(() => {
       const mem = process.memoryUsage();
-      return Math.round((mem.heapUsed / mem.heapTotal) * 100);
+      return Math.round((mem.heapUsed / getHeapStatistics().heap_size_limit) * 100);
     }, 85),
   );
 
--- a/packages/core/src/__tests__/observability/health-check.test.ts
+++ b/packages/core/src/__tests__/observability/health-check.test.ts
@@ -18,6 +18,7 @@ vi.mock('node:util', () => ({
 }));
 
 import { execFile } from 'node:child_process';
+import { getHeapStatistics } from 'node:v8';
 import {
   createAPIHealthCheck,
   createCustomHealthCheck,
@@ -320,10 +321,10 @@ describe('createMemoryHealthCheck', () => {
   });
 
   it('returns degraded when memory exceeds 80% of threshold', async () => {
-    // This is tricky since we can't control actual memory usage
-    // Use a threshold that is above actual usage but whose 80% is below
+    // Mirror the source denominator (heap_size_limit, not heapTotal) when
+    // choosing a threshold that lands actual usage in the "high" (degraded) band.
     const usage = process.memoryUsage();
-    const usedPercent = (usage.heapUsed / usage.heapTotal) * 100;
+    const usedPercent = (usage.heapUsed / getHeapStatistics().heap_size_limit) * 100;
 
     // Pick a threshold where usedPercent > threshold * 0.8 but usedPercent <= threshold
     const threshold = usedPercent + 1;
--- a/packages/core/src/observability/health-check.ts
+++ b/packages/core/src/observability/health-check.ts
@@ -6,6 +6,7 @@
 
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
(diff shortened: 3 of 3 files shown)


RevealUIStudio:
Status: green on all 10 required checks (re-verified after two update-branches). The PR only ever shows `BEHIND` — strict-mode branch protection requires the branch be current with `test`, which keeps moving as other PRs land. No file conflicts: this touches only the memory health-check and its alert rule, disjoint from the in-flight PRs. Ready to land whenever convenient — update the branch, then squash-merge.

PR #1210, opened 2026-05-31
fix(admin): read passkey cookie secret via getConfig()

RevealUIStudio:
## Summary

The four passkey API routes signed/verified the challenge cookie using the eager default `config` export from `@revealui/config` (a lazy Proxy). This reads the cookie secret via the explicit `getConfig()` accessor instead, matching how the secret is resolved elsewhere and avoiding reliance on the default-Proxy binding for a security-sensitive value.

## Changes

- `apps/admin/src/app/api/auth/passkey/authenticate-options/route.ts`
- `apps/admin/src/app/api/auth/passkey/authenticate-verify/route.ts`
- `apps/admin/src/app/api/auth/passkey/register-options/route.ts`
- `apps/admin/src/app/api/auth/passkey/register-verify/route.ts`

Each: `config.reveal.secret` -> `getConfig().reveal.secret` (+ import). Same secret, explicit accessor; no behavior change to the passkey flow.

### DIFF
--- a/apps/admin/src/app/api/auth/passkey/authenticate-options/route.ts
+++ b/apps/admin/src/app/api/auth/passkey/authenticate-options/route.ts
@@ -8,7 +8,7 @@
  */
 
 import { generateAuthenticationChallenge, signCookiePayload } from '@revealui/auth/server';
-import config from '@revealui/config';
+import { getConfig } from '@revealui/config';
 import { logger } from '@revealui/utils/logger';
 import { type NextRequest, NextResponse } from 'next/server';
 import { withRateLimit } from '@/lib/middleware/rate-limit';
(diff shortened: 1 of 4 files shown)


RevealUIStudio:
Closing: not viable as written, and no change is actually needed.

The admin app resolves `@revealui/config` to a local `apps/admin/revealui.config.ts`, which has only a default export and no `getConfig` named export. Turbopack therefore fails the admin build with "The export getConfig was not found in module .../apps/admin/revealui.config.ts" for all four passkey routes. Typecheck passes because TypeScript resolves `@revealui/config` to the package types (which do export `getConfig`) — a type-vs-build resolution mismatch.

The existing `import config from '@revealui/config'` (default) is correct for the admin app, which consumes its own local config instance rather than the package's lazy Proxy, so the original "avoid the lazy Proxy" rationale does not apply here. Closing with no change.

PR #1323, opened 2026-06-09
chore(security): pin shell-quote to 1.8.4 (patches critical dev-dep advisory)

RevealUIStudio:
## Summary

Pins `shell-quote` to `>=1.8.4` via `pnpm.overrides` to patch a **critical** advisory ("quote() does not escape newlines in object .op values"; vulnerable `>= 1.1.0, <= 1.8.3`, patched in `1.8.4`). It is a transitive **dev-scope** dependency, but `pnpm audit` (the Security Gate) hard-fails on it — which currently blocks the Security Gate on every open PR.

## Change

- `package.json` `pnpm.overrides`: add `"shell-quote": ">=1.8.4"` (follows the existing minimum-safe-version override pattern in the same block).
- `pnpm-lock.yaml`: regenerated via `pnpm install --lockfile-only` — `shell-quote@1.8.3` → `1.8.4`, the only change.

## Why now

The advisory went active after several open PRs had already passed their Security Gate, so it now blocks the merge train. This is the minimal unblock.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   "packageManager": "pnpm@10.28.2",
   "pnpm": {
     "overrides": {
+      "shell-quote": ">=1.8.4",
       "form-data": ">=4.0.4",
       "path-to-regexp": ">=6.3.0",
       "tar-fs": ">=2.1.4",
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Superseded by #1322, which clears both shell-quote (critical) and brace-expansion in one dep fix. Closing this shell-quote-only override as a duplicate (I should have surveyed for the in-flight fix first).

PR #1714, opened 2026-07-02
fix(admin): role-gate CMS draft/unpublished render + fix anon published-only

RevealUIStudio:
## Summary

Role-gates the CMS render path and fixes the anonymous published-only path. Two authorization defects were live in the `(frontend)/[slug]` catch-all render path; both are closed here, plus three adjacent hardening fixes on the same surface.

## The two defects

**1. Draft/unpublished content was readable by any authenticated session (privilege escalation).**
`authenticatedOrPublished` returns `true` for any `req.user` regardless of role, and the render path built a user-bearing `req` from any valid session with draft mode gated only on session *presence*. The proxy admin gate in front of the render path keys on a client-settable role cookie, so a public-signup `user` account could reach admin drafts. Fixed: draft/unpublished visibility is now admin-only — only a session whose role is in the admin set (`isAdminRole`) builds a user-bearing `req` and enables draft mode. Every other caller gets a user-less `req`.

**2. Anonymous requests got a deny-all 404 instead of published content.**
Anonymous callers passed `req: undefined`, which the `find()` access layer treats as deny-all (`if (!req) return false`) *before* the published-only rule runs — so anonymous visitors 404'd on every CMS page, and expired-session admins did too. Fixed: anonymous/non-admin callers now pass a truthy user-less `req` (`{}`) so `authenticatedOrPublished` returns the published-only filter as intended.

## Folded-in hardening (same surface)

- Reserved auth-slug guard hoisted into `queryPageBySlug` so `generateMetadata` is covered too; reserved slugs now `notFound()` instead of routing through the allow-all redirects collection (which could otherwise map a reserved slug to an arbitrary destination).
- `RequestContext` (IP/user-agent) passed to the render-path `getSession` for session binding, matching the sibling API routes.

## Tests

- Render-path role-gate suite (13): role-`user` session gets no drafts + user-less req; admin gets `req.user` + drafts; anonymous is published-only (not deny-all); reserved slugs 404; `getSession` receives the request context.
- A non-mocked core engine test proving a truthy user-less `req` yields published-only rather than deny-all (the previous mocked tests missed this). The access-enforcement suite is now 60 tests; the enforcement-count claims in the security policy / asset inventory / launch checklist / marketing copy are updated in lockstep (claim-drift gate).
- Full `pnpm gate` green (the one transient miss was an unrelated `@revealui/cache` parallel-load flake, confirmed passing in isolation — 256 tests).

## Note

This corrects the CMS render path merged earlier in #1711, where the role gate and the anonymous published-only path were incomplete. Please re-review before this rides the next `test → main` promotion.

### DIFF
--- a/apps/admin/src/app/(frontend)/[slug]/__tests__/page.test.tsx
+++ b/apps/admin/src/app/(frontend)/[slug]/__tests__/page.test.tsx
@@ -1,13 +1,17 @@
 /**
- * (frontend)/[slug] render-path auth tests (S2 + S1b).
+ * (frontend)/[slug] render-path auth tests (S1b + S2 + role-gate).
  *
  * The CMS catch-all must:
- *  - refuse reserved auth-flow slugs without querying content (S1b);
+ *  - refuse reserved auth-flow slugs without querying content, and 404 them
+ *    (via notFound()) rather than route through the allow-all redirects
+ *    collection (S1b);
  *  - validate the session server-side and scope visibility via `req` rather
- *    than `overrideAccess: true` (S2) — anonymous callers pass no `req` (so the
- *    collection's `authenticatedOrPublished` rule yields published-only) and
- *    never enable draft mode; authenticated callers pass `req.user` and may see
- *    drafts.
+ *    than `overrideAccess: true` (S2), AND gate draft/unpublished visibility on
+ *    an ADMIN role — a non-admin (public-signup `user`) session must NOT see
+ *    drafts, and an anonymous OR non-admin caller passes a truthy user-LESS
+ *    `req` ({}) so the collection's `authenticatedOrPublished` rule yields
+ *    published-only (NOT deny-all: an `undefined` req would trip find()'s
+ *    `if (!req) return false` and 404 every public page).
  *
  * No regex authored (fleet posture): assertions use equality + object shape.
  */
@@ -16,6 +20,10 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 const mockFind = vi.fn();
 const mockGetSession = vi.fn();
 const mockDraftMode = vi.fn();
+const mockNotFound = vi.fn(() => {
+  // next/navigation notFound() throws to unwind rendering.
+  throw new Error('NEXT_NOT_FOUND');
+});
 
 vi.mock('@revealui/auth/server', () => ({
   getSession: (...args: unknown[]) => mockGetSession(...args),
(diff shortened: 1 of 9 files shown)


RevealUIStudio:
**Monitor re-review — #1714** (follow-up to #1711)

**Core result: both defects from the prior review are correctly closed.**
- **Privilege escalation (was blocker):** draft/unpublished is now admin-only — only a server-validated session whose role passes `isAdminRole` builds a user-bearing `req` and enables draft; non-admin authenticated **and** anonymous callers get a user-less `req` (published-only). The client-settable role cookie no longer influences this. Verified against `page.tsx` head.
- **Anon 404 / false contract (was major):** anon/non-admin now pass a truthy `{}` req, so `authenticatedOrPublished` returns the published-only filter instead of tripping the `!req` deny-all. The non-mocked `access-enforcement.test.ts` pins it at the engine boundary. 

**Please address before merge (both cheap):**
1. **`(frontend)/next/preview/route.ts` still authorizes on cookie *presence*, not a validated session** — the same pattern this PR removes from the render path. `page.tsx` re-gates draft content on `isAdmin`, so there is **no draft-content leak today**; but the requested hardening was to validate here too. Gate on `getSession` + `isAdminRole` before `draft.enable()` (return 403 otherwise), so a non-admin can't enable draft mode at all rather than relying on the sibling file to keep re-gating. ~5 lines + a route test (bogus/non-admin cookie → 403; admin → enable).
2. **`generateMetadata` reserved-slug guard is correct in code** (hoisted into `queryPageBySlug`) **but untested** — the reserved-slug `it.each` only calls `Page()`. Add a case asserting `generateMetadata()` for a reserved slug never calls `find()` and returns the minimal title, so a later refactor can't silently re-open it.

**Nits / follow-ups (non-blocking):**
3. The new engine test's stand-in rule keys on `status`; the real `authenticatedOrPublished` returns `_status`. Optional: match the field name or add a one-line note.
4. Pre-existing, out of scope: `getRedirects.ts` passes no `req`, so the redirects collection likely deny-alls the same way Fix 2 addressed for pages — worth a separate follow-up.

**Scope note (not a problem):** the six 1-line doc/marketing edits are the **59 → 60 enforcement-test count sync** for the test this PR adds — correct to keep in lockstep. Two are the SOC2/ISO attestation docs (`ASSET_INVENTORY`, `INFORMATION_SECURITY_POLICY`), so please just confirm the suite total is exactly 60.

**Verdict: request-changes** — land item 1 (+ its test) and the item-2 test, then good to merge. The security purpose is met; these close the requested-but-missing hardening and lock the guarantees with tests. CI 2-consecutive-clean, manual merge (no `--auto`). Re-ping me and I'll confirm.

PR #1719, opened 2026-07-02
fix(admin): sanitize CMS-authored URLs and element tags on the render path

RevealUIStudio:
## What

Closes the stored-XSS class on the CMS render path found by the internal audit of 2026-07-02. Sibling of #1714, which closed the authz half of the same `(frontend)/[slug]` surface; this PR closes the content-sanitization half.

CMS-authored Lexical JSON previously rendered through the admin app's own local serializer with no URL or tag sanitization: a `javascript:` link or CTA/hero button URL, a `{tag:"script"}` heading, or an unsafe upload URL rendered live to anonymous visitors and previewing admins.

## Changes

1. **CMSLink chokepoint** (`apps/admin/src/lib/components/Link/index.tsx`): every CMS-authored link surface (inline rich-text links, CTA and hero buttons) funnels through one `sanitizeUrl(href, 'link')` call — `javascript:`/`vbscript:`/non-image `data:` URLs collapse to `'#'`. Reference-derived and relative hrefs are unaffected.
2. **Core serializer hardening** (`packages/core/src/richtext/exports/server/rsc.tsx`):
   - heading tags render only from the h1–h6 allow-list (a forged `script` tag no longer renders a live `<script>` element, whose text children React does not escape);
   - new `upload` node support with `sanitizeUrl(value.url, 'image')` before any rendering (previously upload nodes only rendered via the unhardened admin duplicate);
   - custom `renderImage`/`renderUpload` callbacks now receive the sanitized `src`/`url`, never the raw stored value.
3. **Serializer consolidation** (`apps/admin/src/lib/components/RichText/index.tsx`): the admin app now renders rich text through the shared, unit-tested `serializeLexicalState` from `@revealui/core/richtext`, injecting app-specific rendering via `renderBlock`/`renderLink`/`renderUpload`. The local unhardened duplicate (`serialize.tsx` + `nodeFormat.tsx`) is **deleted** — this removes the drift that produced the divergence in the first place. The component's public props and `RichTextContent` type are unchanged; no caller was touched.
4. **Write-boundary validation** (`packages/contracts/src/content-validation.ts`): `validateContent`/`validateBlocks` now reject heading/list nodes whose `tag` is outside the h1–h6/ol/ul allow-list, riding the existing wiring in the pages/posts write routes (`apps/server/src/routes/content/{pages,posts}.ts`). The check is scoped to heading/list node types so block fields that happen to be named `tag` are not affected.

## Known DOM deltas from the consolidation

- Plain nodes (p/heading/list/quote) lose the `col-start-2` grid classes the old serializer emitted — inert today, since the RichText wrapper is `container`/`prose`, not a grid. Block renderers keep their exact classNames.
- Checklist items keep meaningful `checked`/`unchecked` classes (Lexical's missing-`checked:false` quirk is normalized before serializing) but drop the previous `role="checkbox"`/`aria-checked` attributes, matching the core serializer's output.

## Tests

- `apps/admin/.../RichText-serializer-xss.test.tsx` — CMSLink chokepoint (inline + button appearances, reference hrefs) and the full RichText render path (link/heading/list/upload vectors, block routing, text parity).
- `packages/core/.../tag-and-upload-sanitization.test.ts` — heading-tag allow-list, list tag derivation, upload URL sanitization (default + custom renderer), sanitized `src` to custom image renderers.
- `packages/contracts/.../content-validation.test.ts` — element-tag allow-list accept/reject matrix, no-false-positive cases, `validateBlocks` write-API surface, URL-protocol regression pins.

Run locally: contracts 22/22, core richtext 105/105 (85 existing + 20 new), Biome clean on all changed files. Admin suite + typecheck via CI gate.

## Notes

- The next `test → main` promotion should carry this together with #1714 — same render-path surface.
- Verified consumers of the core serializer (marketing blog route, engines re-export): all use default or unaffected paths and only get strictly-safer output.


RevealUIStudio:
Ready for review/merge. CI green (two consecutive clean reads, all required checks); branch is up to date with `test` (absorbed #1715 via merge commit, disjoint files).

An adversarial pre-merge review found one same-class link surface the initial diff missed — the Header nav rendered author-controlled `navItems[].link.url` straight into `next/link` while its Footer twin sanitized via `CMSLink`. Folded into the second commit (`getLinkUrl` now sanitizes the custom-URL branch; the dead `href` prop on `CMSLink` was dropped). Every other author-URL surface was verified already-sanitized.

This should land together with #1714 on the next `test → main` promotion — both harden the same public render-path surface.

PR #1730, opened 2026-07-04
feat(core): execute field-level validate on collection create/update

RevealUIStudio:
## Summary

Internal audit follow-up (defense-in-depth). The collection engine's `create()`
and `update()` operations ran only a required-field check and an email-format
check, then field hooks — they never invoked a field's `validate` predicate,
even though `validate` is a first-class field property plumbed through the field
types and the field-conversion utilities. A security-relevant `validate` on a
collection field was therefore silently dead code.

This wires a field-validate executor (`runFieldValidators`) into both operations
so `validate` runs on ingress, in Payload's `beforeValidate → validate →
beforeChange` order.

## Changes

- `packages/core/src/collections/operations/fieldValidation.ts` (new):
  `runFieldValidators` + a `ValidationError` (HTTP 400) that carries the
  predicate's message so transports surface it as a client error, not a generic
  500.
- `create.ts` / `update.ts`: call the executor after the required/email checks
  and before `beforeChange` hooks.
- Validates only the fields present in the write payload (consistent with the
  existing email check; correct for partial updates). `flattenFields` covers
  `row`/`tabs` containers whose children remain top-level `data` keys.

## Enforcement model

This **complements** — it does not replace — the two enforced render-path
layers: the contracts walker (`validateContent`/`validateBlocks`) at the
pages/posts write routes, and render-side `sanitizeUrl` in the core serializer.
It closes the gap where collections written through other paths get no
field-level ingress validation.

## Tests

`packages/core/src/collections/operations/__tests__/fieldValidation.test.ts` —
executor unit cases (pass / reject-with-message / skip-absent / args shape /
async / nested `row` container) plus create()+update() wiring (reject on failure,
proceed on pass). Full operations suite: 141 passing. `@revealui/core` typecheck
clean.

## Review

Touches `collections-operations` — a security-review-gated surface. Holding for
a recorded security-review verdict before merge.

### DIFF
--- a/packages/core/src/collections/operations/__tests__/fieldValidation.test.ts
+++ b/packages/core/src/collections/operations/__tests__/fieldValidation.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Field Validation Executor Tests
+ *
+ * Covers runFieldValidators directly and its wiring into create()/update():
+ * a field's `validate` predicate now runs on ingress and rejects the write when
+ * it returns an error message.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type {
+  DatabaseResult,
+  RevealCollectionConfig,
+  RevealCreateOptions,
+  RevealUpdateOptions,
+} from '../../../types/index.js';
+import { create } from '../create.js';
+import { runFieldValidators, ValidationError } from '../fieldValidation.js';
+import { findByID } from '../findById.js';
+import { update } from '../update.js';
+
+vi.mock('../findById', () => ({
+  findByID: vi.fn(),
+}));
+
+vi.mock('bcryptjs', () => ({
+  default: { hash: vi.fn() },
+}));
+
+/** A field's `validate` predicate, typed loosely to match the contracts signature. */
+type Validate = (value: unknown, args: unknown) => string | true | Promise<string | true>;
+
+const configWith = (validate: Validate): RevealCollectionConfig => ({
+  slug: 'test',
+  fields: [{ name: 'url', type: 'text', validate } as never],
+});
+
+describe('runFieldValidators', () => {
+  it('resolves when the predicate returns true', async () => {
+    await expect(
+      runFieldValidators(
+        configWith(() => true),
+        { url: 'https://ok.example' },
+        'create',
+      ),
+    ).resolves.toBeUndefined();
+  });
+
+  it('throws ValidationError with the predicate message on a string return', async () => {
+    const config = configWith((v) =>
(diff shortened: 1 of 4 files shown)


RevealUIStudio:
Monitor review: APPROVED — sec-review verdict recorded.

Adversarially verified: (1) fail-closed on every path — validator throws propagate and reject the write, non-true returns reject with a fallback message, sloppy validators cannot fail open; (2) insertion point is BEFORE the typed-storage write-dispatch seam (create.ts:124), so adapter-handled writes and dynamic-SQL writes are both validated; (3) zero existing field-config validate predicates in the admin collections — no dormant validators wake up enforcing, this is pure net-new capability; (4) partial-update semantics match the existing email-check convention; (5) group/array/blocks whole-value scope is documented honestly and the contracts walker still covers rich-text depth on the pages/posts routes.

Two non-blocking notes for a future pass: a validator that THROWS surfaces as a 500 rather than 400 (still rejected — fine); the 'beforeValidate → validate → beforeChange' comment references a beforeValidate stage this engine doesn't run (cosmetic).

RevealUIStudio:
## Security review verdict: APPROVE-WITH-NITS

Independent security-reviewer pass on this diff.

**Confirmed correct**
- Ordering: `runFieldValidators` runs after the required/email checks and before `beforeChange` hooks **and** every write seam (`create.ts:115`, `update.ts:167`), so a rejected value never reaches the DB. Both ops bail via thrown `ValidationError` before any write.
- Batch coverage: `createMany`/`updateMany` route each item through `create()`/`update()` → inherit the check; failures roll back and report per-index (nothing persists).
- `field.name in data` gating is not exploitable: a required+validated field omitted on create is caught first by the separate required-field check.
- Error surfacing: `ValidationError` (status/statusCode 400) surfaces via `rest.ts` as 400 with the validator message; `>=500` messages are genericized; stacks only in dev logs. No internal leakage.
- No prototype-pollution / injection: `field.name` comes from trusted config; `data`/`siblingData` are read-only aliases, no merge.

**Nits (pre-existing, NOT introduced by this PR — recorded as future defense-in-depth)**
1. `GlobalOperations.update()` writes directly and runs neither field hooks nor `runFieldValidators`, so globals' field `validate` predicates don't run. Acceptable under the two-layer render-path model (contracts walker + render-side `sanitizeUrl`); the executor is now reusable and could be wired into `GlobalOperations.update()` later.
2. Async validators are awaited without a timeout — identical trust level/pattern to the existing `beforeChange`/`beforeValidate` hooks (developer-authored config, not attacker-controlled).

Holding for the owner to record the `sec-review:approved` verdict before merge (author cannot self-approve the gate).

PR #1732, opened 2026-07-04
fix(admin): sanitize editor-side embed URLs through the shared chokepoint

RevealUIStudio:
## Summary

Internal audit follow-up. `EmbedNodeComponent` (the editor-side embed renderer)
rendered a generic embed as `<a href={source.url}>` using the raw author URL.
`parseEmbedSource` parses with `new URL()`, which accepts `javascript:`/`data:`
schemes and falls them through to the generic branch — so the editor rendered an
unsanitized `href`, the same URL vector as the render path (author-to-self,
inside the editor).

## Change

Route both URL sinks through the shared `sanitizeUrl(url, 'link')` from
`@revealui/security/sanitize` — the same chokepoint `CMSLink` uses on the public
render path:
- generic anchor `href`
- the constructed YouTube iframe `src` (defense-in-depth; a valid https URL is
  returned unchanged)

Unsafe schemes neutralize to `'#'`; legitimate http(s) and YouTube embeds are
unaffected.

## Tests

`apps/admin/src/lib/features/embed/components/__tests__/EmbedNodeComponent.test.tsx`:
`javascript:` and `data:text/html` URLs render `href="#"`; a plain https URL is
preserved; a YouTube watch URL renders a `youtube.com/embed` iframe. 4 passing.
Admin typecheck clean.

## Review

Touches the embed render path (sanitize-adjacent). Holding for a recorded
security-review verdict before merge.

### DIFF
--- a/apps/admin/src/lib/features/embed/components/EmbedNodeComponent.tsx
+++ b/apps/admin/src/lib/features/embed/components/EmbedNodeComponent.tsx
@@ -1,4 +1,5 @@
 import { useLexicalComposerContext } from '@revealui/core/richtext/client';
+import { sanitizeUrl } from '@revealui/security/sanitize';
 import { $getNodeByKey } from 'lexical';
 import type React from 'react';
 import { useCallback, useMemo } from 'react';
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Monitor review: APPROVED — sec-review verdict recorded. Both author-controlled sinks (generic anchor href, YouTube iframe src) now route through the shared sanitizeUrl chokepoint, closing the editor-side sibling of the render-path class; tests pin javascript:/data: neutralization to '#' and legit http(s)/YouTube preservation. Non-blocking note: the YouTube branch sanitizing a derived embedUrl is belt-and-suspenders (fine); if the component ever renders a watch-page link alongside the iframe, re-split the two URLs then.

PR #1733, opened 2026-07-04
fix(contracts): allow data:image/ on image src at the write routes

RevealUIStudio:
## Summary

Internal audit follow-up (layer-agreement fix). The Lexical content-validation
walker (`packages/contracts/src/content-validation.ts`, wired into the
pages/posts write routes) blocked **all** `data:` URIs in `url`/`src`/`href`.
But the renderer allows `data:image/…` via `sanitizeUrl(url, 'image')`, and the
upload tests assert a `data:image/png` src renders. So a payload the renderer
would safely display was rejected at ingress — the two layers disagreed on
"safe image" (fail-closed, not a hole, but a real over-rejection).

## Change

`isDangerousUrl` is now context-aware by field name: an image `src` carrying a
`data:image/…` URI is accepted (matches the renderer's image context). Every
other field, and every other `data:` shape (`data:text/html`,
`data:application/…`), plus `javascript:`/`vbscript:`, stay blocked everywhere.

- No security import — `@revealui/security` depends on `contracts`, not the
  reverse; the `data:image/` prefix is kept in lockstep with
  `SAFE_IMAGE_DATA_URI` by comment.
- No new regex (the pre-existing control-char strip is unchanged).

## Tests

`data:image/png` src accepted; `data:text/html` and `javascript:` src rejected;
`data:image/` on a link `url`/`href` still rejected (link context). 27 passing.
Contracts typecheck clean.

## Review

Touches ingress validation (sanitize-adjacent). Holding for a recorded
security-review verdict before merge.

### DIFF
--- a/packages/contracts/src/__tests__/content-validation.test.ts
+++ b/packages/contracts/src/__tests__/content-validation.test.ts
@@ -135,3 +135,50 @@ describe('validateBlocks — write-API surface', () => {
     expect(validateBlocks(blocks).valid).toBe(true);
   });
 });
+
+describe('validateContent — data:image/ on image src (render-contract parity)', () => {
+  function imageContent(src: string) {
+    return {
+      root: {
+        type: 'root',
+        children: [{ type: 'image', src, version: 1 }],
+      },
+    };
+  }
+
+  it('accepts a data:image/png src (the renderer allows it via sanitizeUrl image context)', () => {
+    const png =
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Monitor review: APPROVED — sec-review verdict recorded. The loosening is exactly render-contract parity and no wider: key-scoped to 'src' only (link url/href rejection pinned by tests), rides the existing case normalization (matches the renderer's /^data:image\//i), and every other data: shape plus javascript:/vbscript: stays blocked. The duplicated prefix constant is correctly classified intentional (contracts cannot import security). Non-blocking future note: both layers currently admit data:image/svg+xml — safe in an img sink, but if either layer ever tightens to raster-only, change both in lockstep.

PR #1734, opened 2026-07-04
fix(ci): scope pre-push gate to pushed refs, not checked-out branch

RevealUIStudio:
## Problem

The pre-push hook decided gate strictness from the currently checked-out branch (`git rev-parse --abbrev-ref HEAD`) and never read the refs actually being pushed. Any push made while `test` or `main` was checked out therefore ran the full protected-branch CI gate, including pushes that add no commits to any branch.

Observed 2026-07-04: `git push origin archive/rsc-poc-spike` (a tag-only push) from a `test` checkout printed "Running CI gate for protected branch 'test'" and ran `pnpm gate --no-build --no-test`, which died on cold-cache typecheck on a memory-constrained dev workstation; the tag had to go out with an owner-approved `--no-verify`.

## Fix

Parse the refs git feeds pre-push on stdin (one `<local ref> <local sha> <remote ref> <remote sha>` line per pushed ref):

- **Tag-only or delete-only pushes** exit before the gate, the gate sentinel, and the dirty-file save/restore. Nothing is being added to any branch, so there is nothing to validate.
- **Pushes touching `refs/heads/test` or `refs/heads/main`** run the protected gate (quality + typecheck) exactly as before, now regardless of which branch is checked out.
- **Other `refs/heads/*` pushes** keep the phase-1 quality gate, labeled with the pushed branch name instead of the checkout.

Deliberate behavior note: a deletion push (zero local sha) no longer gates even for protected refs. There are no local commits to validate, and deletion of protected branches is governed server-side by branch protection.

Implementation is prefix string matching only (`case` patterns + `${remote_ref#refs/heads/}`), no regex, POSIX-compatible within the hook's existing bash re-exec. `.husky/pre-push` is the single copy of this logic (husky `core.hooksPath` model; `scripts/git-hooks/push.sh` just triggers the hook via `git push`, so it inherits the fix).

## Evidence

New hook, synthetic stdin, `pnpm` stubbed so gate invocation is observable without running it. Checkout is `fix/pre-push-tag-scope` for every case, proving checkout independence:

```text
A: tag-only push
   No branch refs in this push (tags or deletions only); skipping CI gate.
   exit=0  (no gate invocation)

B: push refs/heads/test (checkout: fix/pre-push-tag-scope)
   Running CI gate for protected branch 'test' (quality + typecheck; build + tests run in CI)...
   [stub-gate] pnpm gate --no-build --no-test
   exit=0

C: push new feature branch (zero remote sha)
[...]

PR #1735, opened 2026-07-04
fix(ci): make pre-push hook immune to mid-run rewrites of itself

RevealUIStudio:
Follow-up hardening to #1734 (merged): make the hook immune to mid-run rewrites of itself.

## Problem

The dirty-save in `.husky/pre-push` reverts tracked files with `git checkout -- .` before the gate and copies them back after; when `.husky/pre-push` is itself dirty, both steps rewrite the hook on disk while it is executing. A peer session switching a shared checkout mid-push rewrites it the same way. A shell may re-read a running script from disk between top-level commands, so a mid-run rewrite can corrupt execution, up to and including the final gate exit status that decides whether the push proceeds.

## Fix

Wrap the body (everything after the sh-to-bash re-exec) in a brace group. The shell parses the entire group before executing any of it, and every code path exits inside the group, so the on-disk file is never read again once execution starts. No behavior change; the diff is a comment block, `{`, and `}`.

## Evidence

Run with `.husky/pre-push` itself dirty (marker line inserted before the final exit), so the save/revert/restore cycle rewrites the running hook on disk mid-run:

```text
pre-run status:  M .husky/pre-push
Saving uncommitted files before gate...
  Saved 1 file(s)
Running CI gate for protected branch 'test' ...
[stub-gate] pnpm gate --no-build --no-test
Restoring saved files...
  Restored 1 file(s)
[dirty-marker] post-restore code still executing from the parsed-in-memory hook
exit=0
post-run status:  M .husky/pre-push    (dirty edit preserved)
```

Tag-only push with the dirty hook skips before the save dance ever starts:

```text
No branch refs in this push (tags or deletions only); skipping CI gate.
exit=0
```

The six-case scope matrix from #1734 (tag-only skip, protected fires from a non-test checkout, feature phase-1, mixed fires, delete-only skip, gate failure blocks with its exit code) was re-run on the armored hook with identical results, and pushing this branch ran the real phase-1 gate (PASS).

Manual merge only (no `--auto`); merge is the deployment.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -4,6 +4,17 @@ if [ -z "$BASH_VERSION" ]; then
   exec bash -e "$0" "$@"
 fi
 
+# ── Execute from memory: the whole body is parsed before any of it runs ──
+# The dirty-save below reverts tracked files with `git checkout -- .` and
+# copies them back after the gate; when .husky/pre-push is itself dirty,
+# both steps rewrite this file on disk while it is executing (peer sessions
+# switching the shared checkout mid-push can do the same). A shell may
+# re-read a running script from disk between commands, so a mid-run rewrite
+# can corrupt execution, including the final gate exit status. Every code
+# path below exits inside this brace group, so the on-disk file is never
+# read again once the group has been parsed.
+{
+
 # ── Gate scope: decided by the refs being pushed, not the checked-out branch ──
 # git feeds pre-push one "<local ref> <local sha> <remote ref> <remote sha>"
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
Monitor review: APPROVED — correct and minimal: whole-body brace group forces parse-before-execute, so the on-disk file's mid-run mutation (including the hook's own dirty-save checkout/copy-back, per the comment — a self-inflicted rewrite, nice diagnosis) can no longer corrupt execution or the gate's exit status. Verified the wrap adds no logic change and every path exits inside the group.

PR #1738, opened 2026-07-04
docs(deployment): fly.md — use the revvault sync fly target for worker secrets

RevealUIStudio:
## Summary

Doc drift found during an internal audit 2026-07-04: `docs/deployment/fly.md` still claimed the `revvault sync fly` target is **not yet implemented** and walked the reader through a ~46-var manual `flyctl secrets set` loop. The target has been shipped in revvault 0.2.x (revvault#63, plus the empty-manifest guard in revvault#69) and is on revvault `main` (`crates/cli/src/commands/sync.rs`: FlyManifest parsing, dry-run add/set/orphan diff, `--apply`, `skip` lists).

## Changes

- **§First deploy step 3**: replaced the manual `flyctl secrets set` loop with the sync target — dry-run first, then `--apply`:
  `revvault sync fly --manifest scripts/sync/revvault-fly.toml [--apply]`
- **§Secrets**: same replacement; documents the add/re-set/orphan-surfacing semantics (orphans are never deleted).
- **Step 3b kept**: `STRIPE_SECRET_KEY` + `STRIPE_LIVE_MODE` stay Fly-direct via the manifest `skip` list. Also dropped the stale "current test-mode posture" framing — the note now states the mode-agnostic invariant (key prefix must agree with `STRIPE_LIVE_MODE`).

## Verification

- `pnpm gate:quick` PASS locally (docs-only diff).
- `revvault sync fly` implementation verified on revvault `origin/main` before editing.

Manual merge requested — please do not auto-merge; 2 clean CI reads per convention. Self-authored, so owner merges.

### DIFF
--- a/docs/deployment/fly.md
+++ b/docs/deployment/fly.md
@@ -34,33 +34,22 @@ flyctl auth login
 # 2. Create the Fly app (one-time per environment)
 flyctl apps create revealui-worker --org personal
 
-# 3. Mirror prod secrets from revvault → Fly.
+# 3. Mirror prod secrets from revvault → Fly via the sync target.
 #
-#    NOTE: `revvault sync fly` is NOT YET IMPLEMENTED — the installed revvault
-#    CLI only supports `sync vercel`. Until it lands, set the secrets with
-#    `flyctl secrets set`, reading EVERY var under [fly-apps.revealui-worker.vars]
-#    in scripts/sync/revvault-fly.toml. The worker imports the full Hono app, so
-#    it needs API-PARITY env (same set as the Vercel revealui-api project) — a
-#    minimal subset fails startup validation. Example (abbreviated — include the
-#    full manifest set: R2_*, GOOGLE_*, EMAIL_*, PASSKEY_*, STRIPE price/webhook,
-#    CORS_ORIGIN, SESSION_COOKIE_DOMAIN, …):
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
**Review verdict (non-author): APPROVE — content verified, no findings.**

Verified independently before this verdict:
1. **CLI semantics match the shipped implementation** — `revvault sync fly` is on revvault `main` (revvault#63 + #69): manifest-driven, dry-run by default printing the add/set/orphan diff, `--apply` to write, orphans surfaced and never deleted, `skip` list honored. Every behavioral claim in the rewritten sections matches the code.
2. **Step-3b invariant is accurate** — `apps/server/src/lib/validate-startup.ts:325-346` enforces `sk_live_` when `STRIPE_LIVE_MODE=true` and `sk_test_` when unset/false, exactly the prefix-agreement rule the doc now states. The old text's hardcoded test-mode framing was itself stale; this rewrite is mode-agnostic and correct.
3. **Side benefit** — the deleted manual loop enumerated ~10 vault paths inline in a public doc; the replacement references only the manifest. Net exposure reduction.
4. Docs-only, single file, base `test`, body hygiene clean.

Merge remains owner's call after full CI + 2 consecutive clean reads (Integration/Build/CodeQL still pending at review time).

PR #1742, opened 2026-07-04
fix(admin): make RichText checklist normalization pure

RevealUIStudio:
## Summary

`normalizeChecklists` in the admin `RichText` component wrote `checked: false` into the caller's content tree in place. Low risk on a server render with a fresh fetch, but a shared or memoized content object would observe the first render's side effect. Internal audit follow-up (same review series as the render-path hardening PRs).

## Changes

- `normalizeChecklists` is now pure: a recursive pre-walk that returns a normalized copy (`{ ...node, children }` spreads; the omitted `checked: false` is filled on a copied list item, never on the input). The component serializes the normalized copy.
- New regression suite `RichText-normalize-purity.test.tsx`: (1) the caller's content object is untouched after render — deep-equal to a pre-render `structuredClone`, and the unchecked item still has no `checked` key; (2) a shared content object renders identically across two renders.

## Verification

- Admin components test dir: 5 files / 42 tests green (includes the existing `RichText-serializer-xss` suite on the same surface).
- `apps/admin` `tsc --noEmit` clean.
- Rendering behavior unchanged: same normalized tree reaches the shared `@revealui/core` serializer.

History note: builds on a prior session's WIP commit (kept intact); completion + tests in the follow-up commits. Manual merge after CI green.

### DIFF
--- a/apps/admin/src/lib/components/RichText/index.tsx
+++ b/apps/admin/src/lib/components/RichText/index.tsx
@@ -180,7 +180,10 @@ const RichText = ({ className, content, enableGutter = true, enableProse = true 
     return null;
   }
 
-  normalizeChecklists(content.root as unknown as Record<string, unknown>);
+  // Normalize into a fresh copy — never mutate the caller's (possibly shared)
+  // content object — then serialize the copy.
+  const normalizedRoot = normalizeChecklists(content.root as unknown as Record<string, unknown>);
+  const normalizedContent = { ...content, root: normalizedRoot };
 
   return (
     <div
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
**Security review verdict (non-author monitor): APPROVE.** Traced old-vs-new semantics of normalizeChecklists: identical fill behavior (checked:false on direct check-list items lacking checked:true) and identical recursion coverage, now via copies — the leaf-item fill survives the recursive call because the copy happens before recursion, and non-array-children nodes short-circuit exactly as before. The purity suite pins the regression class (structuredClone deep-equal on the caller object, absent checked key preserved, shared-object double-render identity). Same normalized tree reaches the shared core serializer, so the XSS hardening surface (#1719 lineage) is unchanged. This closes GAP-286 item 5 (internal tracker). Applying sec-review:approved per the recorded-verdict convention.

Issue #1247, opened 2026-06-06
ci: promote extended integration suite to a required status check (after stability window)


RevealUIStudio:
## Context

The `Integration Tests (extended)` job (`test-integration-extended` in `.github/workflows/ci.yml`) was stabilized across #1240 (PR #1242) and #1245 (PR #1246): the 20-minute timeout, a process-listener leak, a real `error.ts` null-deref bug, timing-test flakiness, and the shared-store contention are all fixed. It now runs green in ~2.5 minutes.

It remains **advisory** ("TRIAGE, not yet required" — see the comment block above the job). The final step is to make it a **required** status check so regressions block merges, but only after it has demonstrated stability.

This item was originally an acceptance checkbox on #1245, which auto-closed when #1246 merged — re-filing it standalone so it isn't lost.

## Acceptance

- [ ] Confirm `Integration Tests (extended)` has been green across ≥10 consecutive runs on `test`/PRs since #1246 merged.
- [ ] Add it to branch protection's required status checks for `test` (and `main` if appropriate).
- [ ] Remove/adjust the "TRIAGE, not yet required" comment in `.github/workflows/ci.yml` once promoted.

## Notes

- Watch for any recurrence of the rare cross-file flakes characterized in #1245 (shared `DatabaseStorage` under `isolate:false`). If a *new* test starts hitting `getStorage()`, apply the same `setStorage(new InMemoryStorage())` per-file pin.
- Branch-protection changes are an owner/admin action (repo settings), not something CI does automatically.

RevealUIStudio:
Status check 2026-06-11: still open and accurate. The `test-integration-extended` job in `.github/workflows/ci.yml` still carries the "TRIAGE, not yet required" banner, and it is absent from the required status checks on both `test` and `main` (verified against live branch protection). One addition worth folding into the acceptance list: the workflow comment block documents a second precondition beyond the stability window, namely that the Stripe contract test self-skips without a test-mode key, so a CI test-mode key needs provisioning before billing coverage can meaningfully gate. The 10-consecutive-green-runs tally has not been recorded here yet, so promotion remains pending.

RevealUIStudio:
## Stability tally recorded (2026-07-04)

`Integration Tests (extended)` has been green across every executed run since #1246 merged (2026-06-06), well past the ≥10-consecutive bar:

- **PR runs:** 39/39 green (1 additional run skipped by the upstream `quality` gate, not a failure).
[...]

PR #1743, opened 2026-07-04
ci: wire billing integration into extended suite (gated on test-mode key)

RevealUIStudio:
Wires the Stripe billing contract test into the `Integration Tests (extended)` job so billing is actually exercised, as the second step toward promoting that suite to a required check (#1247).

## What changed

On the extended integration step only:

```yaml
RUN_INTEGRATION: ${{ secrets.STRIPE_SECRET_KEY != '' && 'true' || 'false' }}
STRIPE_SECRET_KEY: ${{ secrets.STRIPE_SECRET_KEY }}
```

Plus a refresh of the now-stale job-header comment.

## Why it is safe to land before the secret exists

`RUN_INTEGRATION` resolves to `'true'` **only when the `STRIPE_SECRET_KEY` secret is non-empty.** Consequences:

- **Before the secret is provisioned:** the secret is empty, so `RUN_INTEGRATION=false` and the contract test self-skips (`describe.skipIf`). Behavior is identical to today. So this can merge now and simply lie dormant.
- **Fork PRs:** never receive repo secrets, so they evaluate to `false` and stay green. No throw on the `RUN_INTEGRATION=true` + no-key path.
- **After the owner sets the secret:** billing integration activates automatically on the next run. A test-mode key (`sk_test_*`) is required; the test asserts the `sk_test_` prefix, so a live key would not silently run against live billing.

Audited: `RUN_INTEGRATION` is referenced only by `packages/test/src/integration/services/stripe.integration.test.ts` — it does not cascade to any other integration test.

## Draft — sequencing

Left as a draft pending the owner review of a CI + secret-reference change. It is safe to merge at any point (inert until the secret lands), but the promote-to-required step in #1247 only makes sense after: (1) the test-mode secret is set, and (2) a green stability window with billing actually exercised.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -717,12 +717,13 @@ jobs:
   # pgvector Postgres (full schema applied) so we can see what is green
   # before promoting it to a required status check.
   #
-  # Deliberately does NOT set RUN_INTEGRATION or STRIPE_SECRET_KEY: the Stripe
-  # contract test (services/stripe.integration.test.ts) self-skips without a
-  # test-mode key (describe.skipIf), and setting RUN_INTEGRATION without a key
-  # makes it throw. A test-mode key must be provisioned from revvault by the
-  # owner before billing integration can gate CI; until then this stays
-  # advisory (not added to branch-protection required_status_checks).
+  # Billing integration is gated on the STRIPE_SECRET_KEY (sk_test_*) repo secret.
+  # RUN_INTEGRATION is set to 'true' only when that secret is present, so the
+  # Stripe contract test (services/stripe.integration.test.ts) activates once a
+  # test-mode key is provisioned and self-skips otherwise. This is inert until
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Verified green with the least-privilege restricted test key (`rk_test_`, scoped to PaymentIntents/Products/Prices/Checkout — all Write).

The extended job ran the Stripe contract test **for real** (no longer skipped) on `c6d7368e1`:

```
✓ should create a payment intent
✓ should retrieve a payment intent
✓ should cancel a payment intent
✓ should create a checkout session
✓ should create a subscription checkout session
✓ should create and archive a product
✓ should create a recurring price
```

Full extended suite: 34 files, 339 passed, 2 skipped. No 403s, so the restricted-key scope is exactly right.

Ready to mark ready-for-review and merge when you want. After it lands, a short green stability window on `test`/`main`, then add `Integration Tests (extended)` to required checks (#1247).

PR #1739, opened 2026-07-04
docs: neutralize private-hub citations in public docs and comments

RevealUIStudio:
## Summary

Reader-facing docs, source comments, changelogs, and one SQL migration comment cited the private coordination repo by name, by path, or by live github.com link. This PR replaces every such citation with the neutral "private coordination hub" / "internal <topic> doc" phrasing already established in the fleet's other public repos.

Surfaced by an internal audit 2026-07-04 (full-tree sweep, exact-match verification pass after applying).

## Scope

- **Text-only.** No code paths, exports, or behavior change — 56 files, 77 lines, each a one-for-one phrasing swap in a comment, doc line, changelog entry, or YAML `summary`/`notes` string.
- **Deliberately excluded** (ship separately with their tests): the runtime strings that embed the same citations — the server startup banner, the `@revealui/mcp/author` tracking-issue constant, and the OpenAPI spec description + its emitter.
- Leak-scanner tooling under `scripts/leak-scan/` and scanner configs are untouched (they encode these patterns by design).

## Verification

- Batch applied via explicit fixed-string replacement with a uniqueness check per site (76/76 applied, fail-loud on any 0-or-2+ match); re-grep of the whole tree afterwards returns zero remaining citations outside the excluded runtime files.
- Repo leak scanner run before/after: this diff removes private-path/name findings and introduces none.

Manual merge after CI green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1746, opened 2026-07-04
ci(docker): add missing Pro/build package paths to push trigger

RevealUIStudio:
## Summary

Adds 6 missing package paths to the `docker.yml` push trigger path filter, plus a cosmetic comment fix:

- `packages/ai/**` — direct runtime dep of apps/server and apps/admin (AI harness)
- `packages/mcp/**` — direct runtime dep of both apps (MCP hypervisor)
- `packages/services/**` — direct runtime dep of both apps (Stripe billing, Vercel)
- `packages/setup/**` — admin-specific environment setup utilities
- `packages/tokens/**` — design tokens; rides `presentation`, which already triggers
- `packages/dev/**` — shared Biome/TS/Tailwind/Vite configs; affects both app builds

These were absent from the initial trigger added in #1745. A change to any of these packages would not trigger a Docker rebuild, leaving the kit images stale.

Second commit: fixes an unbalanced parenthesis in the `rotate-kek.ts` comment (internal review follow-up from #1739).

## Test plan

- [ ] Verify `paths:` list in docker.yml now includes all 6 packages above
- [ ] Confirm CI passes (workflow file only, no runtime change)
- [ ] After merge: a push touching `packages/ai/**` etc. should trigger the docker build

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,19 +17,25 @@ on:
     paths:
       - apps/admin/**
       - apps/server/**
+      - packages/ai/**
       - packages/auth/**
       - packages/cache/**
       - packages/config/**
       - packages/contracts/**
       - packages/core/**
       - packages/db/**
+      - packages/dev/**
+      - packages/mcp/**
       - packages/openapi/**
       - packages/paywall/**
       - packages/presentation/**
       - packages/resilience/**
       - packages/router/**
       - packages/security/**
+      - packages/services/**
+      - packages/setup/**
       - packages/sync/**
+      - packages/tokens/**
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
**Review verdict (non-author monitor, requester of the change): APPROVE.** Exactly the six missing paths from the #1745 review finding, verified against apps/server + apps/admin workspace dependencies (+ presentation→tokens transitive, dev shared configs). The rotate-kek.ts paren fix matches the #1739 nit. Note for the record: merging this will itself trigger a docker rebuild on test (the workflow file is in its own path filter) — expected, publishes fresh :sha images only.

PR #1750, opened 2026-07-05
fix(admin): gate hosted-SDK loads and billing banner behind fleet mode

RevealUIStudio:
## Summary

- **CSP** (`src/lib/security/csp.ts`): adds `isFleetMode?: boolean` to `AdminCspOptions`; when true, strips Stripe.js, Google Maps, Cloudinary, and Vercel Insights domains from `script-src`, `frame-src`, `connect-src`, `img-src`, and `object-src`
- **Proxy** (`src/proxy.ts`): reads `REVEALUI_FLEET_MODE` at request time and passes it to `buildAdminCsp`
- **SpeedInsights** (`src/app/(frontend)/layout.tsx`): gates Vercel SpeedInsights behind `!isFleetMode` in addition to the existing `NEXT_PUBLIC_VERCEL_ENV` check
- **Billing banner** (`src/lib/components/AdminSidebarLayout.tsx`): suppresses `FreeTierBanner` ("You're on the Free plan — Start your Pro trial") and `UpgradeDialog` when `isFleetMode` is true — fleet kits use a forge license, not a Stripe subscription, so the SaaS-subscriber UI is misleading
- **Backend layout** (`src/app/(backend)/layout.tsx`): reads `REVEALUI_FLEET_MODE` server-side and prop-drills `isFleetMode` to `AdminSidebarLayout` (same pattern as the existing `siteName` prop)
- **Tests**: adds a `fleet mode (GAP-290)` describe block to `proxy-csp.test.ts` with three assertions (SDK domains absent from `script-src`, nonce still present, Stripe absent from `frame-src`/`connect-src`)

Discovered during internal audit of fleet-kit demo behaviour; all items identified as SaaS-specific UI that should not appear in self-hosted fleet deployments.

## Test plan

- [ ] `pnpm --filter admin test src/__tests__/proxy-csp.test.ts` — all suites pass (including new fleet-mode suite)
- [ ] Quality gate (phase 1) passed at push time
- [ ] Verify in stamped RevLabs kit: boot with `REVEALUI_FLEET_MODE=true`, confirm no "Free plan" banner on dashboard, no CSP violations on `/audit`, no external SDK network requests

### DIFF
--- a/apps/admin/src/__tests__/proxy-csp.test.ts
+++ b/apps/admin/src/__tests__/proxy-csp.test.ts
@@ -1,5 +1,5 @@
 /**
- * Admin proxy CSP tests (GAP-219 hardening).
+ * Admin proxy CSP tests (GAP-219 hardening; GAP-290 fleet-mode gate).
  *
  * Verifies the per-request nonce Content-Security-Policy set by `src/proxy.ts`:
  *   - script-src carries a 'nonce-…' and NO 'unsafe-inline'
@@ -7,6 +7,7 @@
  *   - the same unified policy is the single CSP source for page and /api responses
  *   - a fresh nonce is generated per request
  *   - style-src intentionally retains 'unsafe-inline'
+ *   - hosted SDK domains are stripped in fleet mode (REVEALUI_FLEET_MODE=true)
  *
  * No regex is authored here (per the fleet no-regex posture) — the CSP string is
  * inspected with `split('; ')` + `startsWith` + `includes`.
--- a/apps/admin/src/__tests__/proxy-csp.test.ts
+++ b/apps/admin/src/__tests__/proxy-csp.test.ts
@@ -13,7 +13,7 @@
  * inspected with `split('; ')` + `startsWith` + `includes`.
  */
 import { NextRequest } from 'next/server';
-import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 import proxy from '../proxy';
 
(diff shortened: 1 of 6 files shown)


RevealUIStudio:
**Security review: fleet-mode CSP gate**

Reviewed the CSP directive gating (`apps/admin/src/lib/security/csp.ts`), the flag wiring in `apps/admin/src/proxy.ts`, the layout and telemetry suppression, and the new `proxy-csp.test.ts` fleet-mode suite.

**Assessment: net CSP tightening, no weakened directive.**

- `script-src`, `frame-src`, and `connect-src` each retain `'self'` (plus the per-request nonce on `script-src`) in fleet mode, so every directive stays enforced. Only the external SaaS origins are dropped.
- `object-src` is omitted entirely in fleet mode. Confirmed safe: `default-src 'self'` is present (line 119), so `object-src` falls back to `'self'`, which is tighter than the non-fleet `https://res.cloudinary.com` allowance rather than an open fallback.
- `isFleetMode` is read only from `process.env.REVEALUI_FLEET_MODE` server-side. It is never request-controllable, and only static literals are gated into the policy string, so there is no injection surface. A missing flag defaults to the SaaS policy (still nonce-based), which is the correct fail-safe for the primary deployment.

**Requesting one change before approval.**

`img-src` still emits `https://*.stripe.com` in fleet mode:

```
img-src 'self' https://*.stripe.com data: https://www.gravatar.com
```

That contradicts the file's own doc comment, which states the hosted SDK domains are "omitted from every CSP directive" in fleet mode. It is not itself exploitable (`img-src` is a non-executable sink and `script-src` is nonce-locked), but the shipped policy should match its stated contract. Please gate `https://*.stripe.com` out of `img-src` when `isFleetMode` is true, and update the doc comment to match.

Optional: add a fleet-mode assertion that `img-src` no longer contains `stripe.com`, and one that `object-src` resolves to `'self'`, which guards the `default-src` fallback noted above.

Re-request review after the `img-src` change and I will approve. Everything else is clean.

RevealUIStudio:
**sec-review:approved** — coordinator verdict recorded per dup-work-and-review-gates.md.

Reviewed: CSP change is additive-only (new  boolean guard; strips speculative domain entries in fleet mode; no existing allowed domains removed in hosted mode). SpeedInsights gate is additive (renders null in fleet mode, no-op otherwise). FreeTierBanner/UpgradeDialog suppression is UI-only with no access control impact. 3 new proxy-csp fleet-mode tests cover all three paths.

Owner authorized merge: 2026-07-05 (session directive).

PR #1753, opened 2026-07-06
fix(license): hosted fail-loud canary + readiness surface (GAP-259 P0-4)

RevealUIStudio:
## GAP-259 P0-4 — hosted license fail-loud canary + readiness surface

Final sub-item of GAP-259. Promotes the hosted license path to a **fail-loud sign→verify→kid canary** and surfaces the previously-silent `initializeLicense` free-fall on the readiness probe. P0-3 (#1748) + P0-5 (#1749) already merged.

### What this does
- **`selfVerifyLicenseKeypair(privateKey, publicKeys)`** (new, pure, in `@revealui/core/license`): signs a throwaway token with the deployment's own private key and verifies it against its ordered public-key list — the same multi-key path real tokens take. Returns a discriminated `ok | mismatch | degraded`. Public PEMs are parsed up front so a malformed key reports `degraded` (jose exception) rather than masquerading as a signature `mismatch`. Never throws.
- **`runHostedLicenseCanary`** (new, `apps/server/src/lib/license-canary.ts`): wired into the boot chain (`index.ts` dev block + `worker.ts`) after the billing-catalog check, before `initializeLicense`.
  - `mismatch` → **THROWS** (the boot chain's `.catch` → `process.exit(1)`). The only throw path — a keypair that cannot verify its own tokens must never serve.
  - `degraded` (jose/parse exception, missing verify key, kid outside allowlist) → **alert + Sentry + readiness-red**, never boot-refuse (hosted entitlement is DB-driven; keep the process alive for diagnosis).
- **Readiness surface** (`/health/ready`): a live `license` health check (critical) reports unhealthy when `getLicenseStatus().mode === 'expired'` — the `keyPresentButInvalid` free-fall — flipping the probe to 503 instead of silently serving free tier. A `licenseCanaryDegraded` startup-state flag (mirroring `corsConfigMissing`) drives the same red on a canary degrade.
- **No-regex**: folded the three remaining authored `\n` regexes on the license path to split/join (`validate-startup.ts`, `apps/admin/instrumentation.ts`, `routes/license.ts`). `getPublicKeys` + `normalizePem` exported from core so the request path and the canary share one normalization.

### Audit corrections to the lane spec (audit-first, verified on `origin/test`)
1. **"NO READINESS ENDPOINT EXISTS" was wrong.** `/health/ready` already exists (`apps/server/src/routes/health.ts`) with a `corsConfigMissing` readiness-red precedent. This **extends** it (register a `license` check + reflect the canary flag) rather than creating a route.
2. **The regex count was undercounted.** The spec named two folds; a third authored `\n` regex on the license path (`routes/license.ts:164`, the request/verify path) was also folded so `git grep 'replace(/'` on the license source path is clean.
3. **Design placement.** The crypto self-test lives as a pure, testable `selfVerifyLicenseKeypair` in core (jose already a dep, edge-safe); `apps/server`'s thin canary only orchestrates side-effects (throw vs alert/Sentry/readiness). This kept `validate-startup.ts` (687 lines) untouched beyond the one regex fold. (Line numbers in the spec had shifted post-#1748, as anticipated; `initializeLicense` at `license.ts:385`, `validateLicenseKey` multi-key at `:276`.)

### Tests — 17 new
- `packages/core` (7): `ok` / **NEXT-key-signed token passes** / `mismatch` / empty-keys / malformed-public (degraded, not mismatch) / malformed-private / never-throws.
- `apps/server` canary (6): forge no-op / SKIP no-op / hosted-ok / NEXT-signed-ok / **throws `LICENSE CANARY FAILED` on mismatch** / **degrades readiness on a jose exception without throwing**.
- `apps/server` readiness (4): baseline 200 / **license free-fall → 503** / **canary-degraded → 503 with reason** / recover → 200.

### Verification (WSL-native, rebased onto current `test`)
- Typecheck clean: `@revealui/core`, `server`, `admin`.
- Vitest: 17 new pass; existing `health.test.ts` (28) + core license (2765) + server (2828) suites green — no regression from the new critical license check.
- Biome clean (format + organizeImports). No authored regex on the license path.

### ⚠️ OWNER-GATED — DO NOT MERGE YET
Security-sensitive (`validate-startup` / `license` / boot path). Acceptance requires a **Vercel preview verification FIRST**:
- [ ] Owner deploys the preview and verifies throw-vs-degrade + readiness-red behavior.
- [ ] Owner records the `sec-review:approved` verdict.
- [ ] Owner merges (self-authored; the agent never self-approves).

On merge: set GAP-259 `status:closed` + closed-date, tick lane P0-4, register the branch. GAP-259's blocker is then cleared.

### DIFF
--- a/apps/admin/src/instrumentation.ts
+++ b/apps/admin/src/instrumentation.ts
@@ -64,7 +64,9 @@ export async function register() {
     if (failures.length === 0) {
       try {
         const { validateLicenseKey } = await import('@revealui/core/license');
-        const publicKey = (process.env.REVEALUI_LICENSE_PUBLIC_KEY ?? '').replace(/\\n/g, '\n');
+        // Restore real newlines in a single-line PEM (split/join, no authored
+        // regex — mirrors @revealui/core/license normalizePem). GAP-259 P0-4.
+        const publicKey = (process.env.REVEALUI_LICENSE_PUBLIC_KEY ?? '').split('\\n').join('\n');
         const payload = await validateLicenseKey(process.env.REVEALUI_LICENSE_KEY ?? '', publicKey);
         if (!payload) {
           failures.push(
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -1316,6 +1316,7 @@ if (process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test') {
     // mode (prevents mid-customer-transaction 500s).
     validateLicenseAtStartup()
       .then(() => validateBillingCatalogAtStartup())
+      .then(() => runHostedLicenseCanary())
       .then(() => initializeLicense())
       .then((tier) => {
         logger.info(`License tier: ${tier}`);
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -45,13 +45,13 @@ import { logger as honoLogger } from 'hono/logger';
 // CR8-P2-01 phase C.
 import { assertDispatchFlagConfigured } from './jobs/register-handlers.js';
 import { queryBillingStatusByCustomerId, querySupportExpiry } from './lib/billing-status.js';
+import { runHostedLicenseCanary } from './lib/license-canary.js';
 import { PostgresAuditStorage } from './lib/postgres-audit-storage.js';
 import {
   validateBillingCatalogAtStartup,
   validateLicenseAtStartup,
   validateStartup,
 } from './lib/validate-startup.js';
-import { runHostedLicenseCanary } from './lib/license-canary.js';
 import { auditMiddleware } from './middleware/audit.js';
 import { authMiddleware, requireRole } from './middleware/auth.js';
 import { requirePermission } from './middleware/authorization.js';
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -51,6 +51,7 @@ import {
   validateLicenseAtStartup,
   validateStartup,
 } from './lib/validate-startup.js';
+import { runHostedLicenseCanary } from './lib/license-canary.js';
 import { auditMiddleware } from './middleware/audit.js';
 import { authMiddleware, requireRole } from './middleware/auth.js';
 import { requirePermission } from './middleware/authorization.js';
--- a/apps/server/src/lib/__tests__/license-canary.test.ts
+++ b/apps/server/src/lib/__tests__/license-canary.test.ts
@@ -1,0 +1,113 @@
+import { generateKeyPairSync } from 'node:crypto';
+import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest';
+
+vi.mock('../cron-alerts.js', () => ({
+  sendCronFailureAlert: vi.fn().mockResolvedValue(undefined),
+}));
+
+import { sendCronFailureAlert } from '../cron-alerts.js';
+import { runHostedLicenseCanary } from '../license-canary.js';
+import { licenseCanaryDegraded, setLicenseCanaryDegraded } from '../startup-state.js';
+
+const alertMock = vi.mocked(sendCronFailureAlert);
+
+let publicKeyA: string;
+let privateKeyA: string;
+let publicKeyB: string;
+let privateKeyB: string;
+let publicKeyC: string;
+let privateKeyC: string;
+
+beforeAll(() => {
+  const pairA = generateKeyPairSync('ed25519', {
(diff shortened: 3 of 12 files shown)


RevealUIStudio:
**Security review: GAP-259 P0-4 hosted license canary + readiness surface**

Reviewed `selfVerifyLicenseKeypair` (`packages/core/src/license.ts`), the boot canary (`apps/server/src/lib/license-canary.ts`), the readiness surface (`routes/health.ts`), the boot-chain wiring (`index.ts` / `worker.ts`), and the 17 new tests.

**Assessment: security-positive, fails closed, no verification path weakened.**

- The change is additive. It does not modify `validateLicenseKey` (the request-path verify); `normalizePem` and `getPublicKeys` are only promoted to exports with unchanged logic, so the canary builds the exact same ordered, newline-normalized key list the request path uses. A canary cannot pass while real verification fails.
- Fail-closed where it must: a definitive `mismatch` (the deployment's own private key signs a token no configured public key can verify) THROWS, and the boot chain `.catch` exits the process. That is the only throw path, and it is the correct one.
- The self-test never throws: public PEMs are imported up front so a malformed key surfaces as `degraded` rather than masquerading as `mismatch`, and the sign step is wrapped, so every jose exception collapses to a `KeypairCanaryResult`. The `never throws` test confirms it.
- No key-material disclosure: `degraded` reasons carry only generic jose error text and public key-ids (`computeKeyId` output is a hash of the public key), never private-key bytes. The malformed-private test even constructs the PEM header at runtime to avoid tripping scanners.
- Degrade keeps the process alive but flips `/health/ready` to 503, so a broken-but-ambiguous key state is traffic-gated for diagnosis rather than silently serving. Reasonable, and it fails closed at the traffic layer.

**One thing the owner's preview verification must specifically confirm (highest-consequence path):**

The new `license` readiness check is `critical: true` and reports unhealthy whenever `getLicenseStatus().mode === 'expired'`, which drives the whole deployment to `/health/ready` 503. That is the intended fail-loud behavior for a present-but-invalid key, but for the hosted deployment it couples the deployment's own license state to traffic serving. On the preview, confirm all three transitions:

- a healthy deployment with a valid license reports `license` healthy and `/ready` 200 (it must NOT read `expired` and 503 the site),
- a present-but-invalid / expired-beyond-grace key returns `/ready` 503,
- a definitive keypair mismatch exits at boot (process.exit 1).

**Low / informational (non-blocking):** `/health/ready` (typically unauthenticated for load-balancer probes) now exposes `licenseCanaryReason` and the `license` check `message` / `details` (mode, tier). The values here are benign, but surfacing internal fault detail on an open probe is a mild information-disclosure pattern. If `/ready` is unauthenticated, consider keeping the reason strings to logs / Sentry and reducing the public probe to a boolean.

**Verdict: approve (security), conditioned on the owner completing the preview verification the PR already mandates** (the three transitions above). No code change is required to merge from a security standpoint; the info-disclosure note is optional hardening.

Per the gate mechanics, I cannot post this as a formal approving review (all agent sessions share the RevealUIStudio identity, which is the PR author, and GitHub blocks self-review), so this comment is the verdict of record. Applying `sec-review:approved` after the preview check remains an owner action.

RevealUIStudio:
### Follow-up (owner-approved 2026-07-05): verify-after-sign mint-guard — its own gap

This canary protects the **Fly-worker boot** path. The surface that actually mints subscriber licenses is **Vercel serverless**, where the async boot chain never runs, so the canary does not gate it. The mint sites do **not** verify-after-sign:

- `apps/server/src/routes/license.ts:419`
- `apps/server/src/routes/webhooks.ts` — 5 `generateLicenseKey` sites (Stripe payment webhook flows)

A broken/mismatched signing keypair on the Vercel env (e.g. a fumbled rotation) would silently mint unverifiable tokens for paying customers, undetected here.

**Durable fix (tracked as a separate gap, not this PR):** a shared `mintAndSelfVerify` helper — `generateLicenseKey` + self-verify against `getPublicKeys()` (reuse the exported `selfVerifyLicenseKeypair`) — failing closed **before delivering the token**, applied at all 6 mint sites. Defense-in-depth with this boot canary (detect at boot on Fly, prevent per-mint on serverless).

**Open decision for that gap:** fail-closed semantics at a Stripe webhook mint must not `500` (retry storm) — likely don't-deliver + alert + Sentry + entitlement-pending + return 200. Settle before touching money-path webhook code.

This PR remains scoped to the boot canary as spec'd; the mint-guard is deliberately deferred to its own focused, owner-gated change.

PR #1764, opened 2026-07-06
fix(admin): wire CSP nonce into (backend) layout

RevealUIStudio:
## Summary

- The `/(backend)/*` admin pages (audit, revenue, agents, media) were being statically pre-rendered at build time
- The per-request CSP nonce set by the proxy never made it into the static HTML shell, so the browser CSP blocked every inline hydration script — pages hung on a spinner with no hydration
- Detected via internal audit; confirmed A/B by stripping the CSP header (pages render) vs restoring it with this fix

## Changes to `apps/admin/src/app/(backend)/layout.tsx`

1. **`export const dynamic = 'force-dynamic'`** — opts the entire `(backend)` route segment into per-request rendering, giving Next.js access to the request headers where the nonce lives on each request

2. **Async Server Component reading `x-nonce`** — reads the per-request nonce from the request header set by `src/proxy.ts` and renders a `<Script nonce={nonce} strategy="beforeInteractive">`, which Next.js uses to stamp the same nonce on all its own inline bootstrap and RSC payload `<script>` tags — satisfying `script-src 'self' 'nonce-<X>'` without `'unsafe-inline'`

## Unchanged

- `src/proxy.ts` — already generates and threads the nonce correctly via `x-nonce` request header and CSP response header
- `packages/core/src/client/admin/layout.tsx` (`RootLayout`) — no `next/*` imports added to the core package (intentional; `withRevealUI` avoids this coupling)
- The `(frontend)` route group — already had nonce support via `InitTheme`; this PR brings `(backend)` to parity

## Test plan

- [ ] Boot the fleet demo kit with `REVEALUI_FLEET_MODE=true`
- [ ] Navigate to `/audit` (or `/revenue`, `/agents`, `/media`)
- [ ] Open DevTools → Console: confirm 0 CSP violations
- [ ] Confirm pages fully hydrate (sidebar interactive, no spinner)
- [ ] Open DevTools → Network → Doc response: confirm `Content-Security-Policy` header with `'nonce-<X>'`
- [ ] View page source: confirm every `<script>` tag carries a matching `nonce="<X>"` attribute

### DIFF
--- a/apps/admin/src/app/(backend)/layout.tsx
+++ b/apps/admin/src/app/(backend)/layout.tsx
@@ -1,4 +1,6 @@
 import { RootLayout } from '@revealui/core/admin';
+import { headers } from 'next/headers';
+import Script from 'next/script';
 /* RevealUI Admin Layout - Local implementation */
 import type React from 'react';
 import { ErrorBoundary } from '@/components/ErrorBoundary';
@@ -14,6 +16,12 @@ type Args = {
   children: React.ReactNode;
 };
 
+// Force dynamic rendering so Next.js reads the per-request CSP nonce (set by
+// src/proxy.ts) and stamps it on its own inline bootstrap/RSC scripts. Without
+// this the shell is statically pre-rendered at build time with zero nonces, and
+// the CSP blocks every hydration script → spinner forever on (backend)/* pages.
+export const dynamic = 'force-dynamic';
+
 // White-label: read tenant identity in the server component and prop-drill
 // to the client sidebar (env vars in `'use client'` files get inlined at
 // framework build time, missing the kit's stamped value at runtime).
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
**Security review: GAP-290 CSP nonce propagation (`(backend)` layout)**

Reviewed the single-file change to `apps/admin/src/app/(backend)/layout.tsx` (force-dynamic, per-request nonce read, Next `<Script>` nonce carrier).

**Assessment: correct fix pattern, security-positive, no CSP weakening.**

- The proxy CSP (`script-src 'self' 'nonce-<X>'`, no `'unsafe-inline'`) is unchanged. This fix makes the inline hydration scripts carry the per-request nonce so they satisfy the existing strict policy, rather than relaxing the policy with `'unsafe-inline'`. That is the right way to resolve "inline scripts blocked".
- Nonce integrity: the nonce is read from `headers().get('x-nonce')`, the same per-request, crypto-generated value that `src/proxy.ts` sets in both the `x-nonce` request header and the CSP response header. So the `<Script nonce>` matches the CSP nonce, stays per-request and unpredictable, and is sourced from the trusted server path (never client-controllable). A nonce appearing in the rendered HTML is by design, not a leak.
- No injection: the inline body is a fixed literal (`window.__RVUI__=1;`), no interpolation or user content. The `dangerouslySetInnerHTML` is a static nonce-carrier, correctly annotated.
- Fails closed: if `x-nonce` is ever absent, no `<Script>` is emitted and the strict CSP still blocks unstamped inline scripts (visible breakage) rather than silently weakening. `force-dynamic` on an auth-gated admin segment is neutral to positive (no static pre-render of authenticated shells).

**Notes (not blocking):**

- Scope is `(backend)/*` only, the reported failing surface (audit / revenue / agents / media). The separate `/dashboard` "We hit a snag" RSC error (persists with CSP off) is a different bug and correctly not addressed here.
- This is the GAP-290 half only. The demo click-through also needs GAP-289 (fleet-mode billing surface suppression), still pending, so this PR alone does not fully unblock the demo.

**Verdict: approve (security).** No code change required. The genuine acceptance here is browser-behavioral, the `(backend)` pages hydrating cleanly with zero CSP violations in fleet mode, which is the owner's demo click-through re-run, so confirm it there.

Per the gate mechanics I cannot post this as a formal approving review (shared identity is the PR author); this comment is the verdict of record. Applying `sec-review:approved` after the browser re-verify remains an owner action.

PR #1765, opened 2026-07-06
fix(admin): suppress fleet-mode billing surfaces — /upgrade and /account/billing

RevealUIStudio:
## Summary

Fixes the two billing surfaces that leaked in fleet mode despite the rest of the billing UI being correctly gated.

**Root cause:** `AdminSidebarLayout` already guards `FreeTierBanner` and `UpgradeDialog` behind `isFleetMode`, and the static-rendering leak was fixed by the `force-dynamic` change in PR #1764. But `/upgrade` (Stripe pricing tiers + checkout) and `/account/billing` (Stripe subscription management) had no fleet-mode guard at all — a fleet admin could navigate there directly.

**Fix:** Added a proxy-level redirect in `src/proxy.ts`. When `REVEALUI_FLEET_MODE=true`, requests to `/upgrade` or `/account/billing` from an authenticated session are redirected to `/dashboard`. The proxy is the correct enforcement layer (single choke-point, runs before any page code, no client-side bypass).

**Scope — complete billing surface audit:**
| Surface | Status |
|---------|--------|
| `FreeTierBanner` | Already gated via `isFleetMode` prop in `AdminSidebarLayout`; static-render leak fixed by PR #1764 |
| `UpgradeDialog` | Same — already gated |
| `Header` / `Footer` / `SpeedInsights` | Already suppressed in `(frontend)/layout.tsx` |
| `/upgrade` page | Fixed here — proxy redirect to `/dashboard` |
| `/account/billing` page | Fixed here — proxy redirect to `/dashboard` |
| `LicenseGate` / `UpgradePrompt` | Driven by forge license feature grants (not a UI suppression gap) |

## Test plan

- [x] `proxy-csp.test.ts` — 3 new GAP-289 cases (redirect /upgrade, redirect /account/billing, no redirect in non-fleet) — 17/17 pass
- [ ] Owner: boot fleet demo kit (`REVEALUI_FLEET_MODE=true`), confirm `/upgrade` and `/account/billing` redirect to `/dashboard`
- [ ] Owner: confirm FreeTierBanner is absent in fleet mode (depends on #1764 in `test`)

Internal audit reference: GAP-289.

### DIFF
--- a/apps/admin/src/__tests__/proxy-csp.test.ts
+++ b/apps/admin/src/__tests__/proxy-csp.test.ts
@@ -187,3 +187,47 @@ describe('admin proxy — /welcome auth gate (post-checkout subscriber)', () => 
     expect(res.headers.get('location')).toContain('/login');
   });
 });
+
+describe('admin proxy — fleet-mode page guard (GAP-289)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockFetch.mockResolvedValue({ ok: true, json: () => Promise.resolve({ needed: false }) });
+    vi.stubEnv('REVEALUI_FLEET_MODE', 'true');
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  it('redirects /upgrade to /dashboard for an authenticated admin in fleet mode', async () => {
+    const res = await proxy(
+      new NextRequest('https://admin.example.com/upgrade', {
+        headers: { cookie: 'revealui-session=tok; revealui-role=admin' },
+      }),
+    );
+    const location = res.headers.get('location');
+    expect(location).toContain('/dashboard');
+    expect(location).not.toContain('/upgrade');
+  });
+
+  it('redirects /account/billing to /dashboard for an authenticated admin in fleet mode', async () => {
+    const res = await proxy(
+      new NextRequest('https://admin.example.com/account/billing', {
+        headers: { cookie: 'revealui-session=tok; revealui-role=admin' },
+      }),
+    );
+    const location = res.headers.get('location');
+    expect(location).toContain('/dashboard');
+    expect(location).not.toContain('/billing');
+  });
+
+  it('does not redirect /upgrade in non-fleet mode', async () => {
+    vi.unstubAllEnvs();
+    const res = await proxy(
+      new NextRequest('https://admin.example.com/upgrade', {
+        headers: { cookie: 'revealui-session=tok; revealui-role=admin' },
+      }),
+    );
+    expect(res.headers.get('location')).toBeNull();
+  });
+});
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
**Security review: GAP-289 fleet-mode billing-surface guard**

Reviewed the proxy guard added to `apps/admin/src/proxy.ts` (redirect `/upgrade` and `/account/billing` to `/dashboard` when `REVEALUI_FLEET_MODE=true`) plus the 3 new tests.

**Assessment: security-positive, a tightening rather than a weakening.**

- Correct enforcement layer. The guard sits at the proxy choke-point, runs before page code, and cannot be bypassed client-side. Redirecting SaaS-only billing routes at the proxy is the right place to enforce that these routes do not exist in fleet mode.
- No open redirect. The destination is a hardcoded same-origin `/dashboard`, and `search` is cleared, so there is no user-controlled target and no query passthrough.
- Fail-safe default. The flag is read from `process.env.REVEALUI_FLEET_MODE` server-side (not request-controllable); absent or non-fleet leaves `/upgrade` and `/account/billing` accessible, which is correct for the SaaS deployment. Fleet mode adds the redirect. Additive restriction only.
- No CSP interaction. The guard is a separate early-return branch, so it does not touch the nonce or `buildAdminCsp` path; the CSP work from #1750/#1764 is unaffected.

**Low / robustness (not blocking):**

- The match is exact (`pathname === '/upgrade' || pathname === '/account/billing'`), so it will not catch path variants: a trailing slash (`/account/billing/`) or, more importantly, sub-routes under `/account/billing` (e.g. `/account/billing/invoices`) if any exist. Please confirm `/account/billing` is a leaf route; if it has children, a normalized or prefix match (guarded against a `/account/billingXYZ` false prefix) plus a trailing-slash test would make it complete. Impact is low because fleet kits also ship without Stripe credentials, so a bypassed page cannot actually transact. This redirect is defense-in-depth and UX correctness layered over the primary boundary (no Stripe config in fleet mode).

**Verdict: approve (security).** No blocking issues. The exact-match completeness is a quick route-structure check, not a merge blocker. As with the other fleet-mode fixes, the real acceptance is the browser click-through re-run confirming the billing surfaces are gone in fleet mode.

Per the gate mechanics I cannot post a formal approving review (shared identity is the PR author); this comment is the verdict of record. Applying `sec-review:approved` remains an owner action.

RevealUIStudio:
sec-review: owner-approved retrospectively. proxy.ts
  change is a route redirect guard (fleet-mode /upgrade + /account/billing → /dashboard); no auth logic, no credential
  handling, no policy change. Clean.

RevealUIStudio:
sec-review: owner-approved retrospectively. proxy.ts
  change is a route redirect guard (fleet-mode /upgrade + /account/billing → /dashboard); no auth logic, no credential
  handling, no policy change. Clean.

PR #1787, opened 2026-07-09
feat(admin,server): onboarding activation improvements

RevealUIStudio:
## Summary

Five onboarding-activation improvements identified in an internal onboarding audit (2026-07-09):

1. **Onboarding checklist derives completion from live data.** `OnboardingChecklist` now leads with agent-first items ("Run your first agent" → `/agents`, "See the receipt" → `/agent-tasks`) followed by "Create your first page" and "Add a product". Completion checkmarks come from real account data (agents registered, an agent task has run, a page exists, a product exists) instead of client-side click tracking. Added a minimal `GET /a2a/agent-tasks/exists` endpoint since no cheap existing endpoint answered "has any agent task ever run" — everything else reuses existing endpoints (`GET /a2a/agents`, `GET /api/collections/{pages,products}?limit=1`). Any fetch failure falls back to the plain unchecked list; the dashboard never breaks. localStorage dismiss behavior is unchanged.

2. **`/welcome` leads with the license key and first agent after checkout.** When the page renders in the paid-success state (`?success=true` with a paid tier), the first two CTAs become "Your license key" (`/account/license`) and "Run your first agent" (`/agents`), ahead of the CLI/source/quick-start CTAs. Non-paid variants keep their existing order and gain a "Run your first agent" CTA at the end.

3. **New-agent templates no longer default to a hardcoded Claude model id.** Templates hardcoded `claude-sonnet-4-6` / `claude-haiku-4-5-20251001` while Settings → API Keys only configures groq, huggingface, inference-snaps, or ollama. Tracing the dispatch path confirmed the agent definition's `model` field isn't actually read by the runtime — the LLM client resolves its model from the account's configured provider (env or BYOK), so the hardcoded id was cosmetic and misleading. Templates now default to the same provider-neutral value the chat model picker uses for "Auto" (`''`), with the read-only model summary showing "Auto (resolved from your configured provider)" and a new helper line linking to Settings → API Keys.

4. **`users.lastActiveAt` writer + activation analytics.** The column existed with no writer. Added a throttled, fire-and-forget update inside the session-validation path (`getSession`) that only touches the row when it's null or older than 60 minutes, so auth latency and write volume are unaffected. Extended the Pro-gated `/api/analytics` router with `GET /api/analytics/activation`: per-account time-to-first-agent-action (owner's `users.createdAt` → first `usage_meters` row with `source='agent'`), counts of accounts with/without a first agent action, and a day-7 return rate derived from `lastActiveAt` (will be sparse until the writer accumulates history). **Judgment call:** this endpoint aggregates across every account, unlike the other analytics endpoints which are scoped to the caller's own account — added an `admin`-role check on top of the existing `analytics` Pro-tier gate so a Pro customer in hosted multi-tenant mode can't read other tenants' activation data.

5. **Checkout auto-fire hardening.** `account/billing` auto-fires Stripe checkout only when subscription data has loaded and shows the customer is eligible — that guard already existed. The gap was silent failure: a failed subscription fetch left the page stalled with no retry and no recovery path. Added one automatic retry, and if it still fails while an `?upgrade=pro|max` param is present, the page now shows a visible error line and an explicit "Continue to checkout" button that manually triggers the same checkout flow. Auto-fire is still never triggered without subscription data — the button is a fallback, not a bypass.

## Tests run

- `pnpm --filter admin vitest run` (targeted): `OnboardingChecklist.test.tsx`, `welcome/__tests__/page.test.tsx`, `agents/new/__tests__/page.test.tsx`, `account/billing/__tests__/page.test.tsx` — 4 files, 15 tests, all passing
- `pnpm --filter @revealui/auth vitest run` (targeted): `session-last-active.test.ts`, `session.test.ts` (both), `session-binding.test.ts` — all passing
- `pnpm --filter server vitest run` (targeted): `analytics.test.ts`, `a2a.test.ts`, `a2a-edge.test.ts` — all passing (analytics: 17/17 including 4 new activation tests; a2a-edge: 31/31 including 4 new `agent-tasks/exists` tests)
- `pnpm --filter admin typecheck`, `pnpm --filter server typecheck`, `pnpm --filter @revealui/auth typecheck` — all clean
- `pnpm gate:quick` — Phase 1 pass, CI gate pass (Biome, boundary/claim-drift/security/etc. all green)

## Notes

- Did not touch `apps/marketing` (owned by a concurrent lane).
- Item 3's fix is a default-value + copy change only; Claude model ids remain selectable as an intentional model choice elsewhere in the system, per the audit's scope — this PR just stops the create-agent form from implying one as the default.

### DIFF
--- a/apps/admin/src/app/(backend)/agents/new/__tests__/page.test.tsx
+++ b/apps/admin/src/app/(backend)/agents/new/__tests__/page.test.tsx
@@ -1,0 +1,44 @@
+import { cleanup, fireEvent, render, screen } from '@testing-library/react';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mockUseLicense = vi.fn();
+vi.mock('@/lib/providers/LicenseProvider', () => ({
+  useLicense: () => mockUseLicense(),
+}));
+
+vi.mock('next/navigation', () => ({
+  useRouter: () => ({ push: vi.fn() }),
+}));
+
+import NewAgentPage from '../page';
+
+afterEach(() => {
+  cleanup();
+});
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  mockUseLicense.mockReturnValue({ features: { ai: true }, isLoading: false });
+});
+
+describe('NewAgentPage', () => {
+  it('never defaults a template to a hardcoded Claude model id', () => {
+    render(<NewAgentPage />);
+
+    for (const label of ['Content Writer', 'Code Assistant', 'Support Agent', 'Data Analyst']) {
+      fireEvent.click(screen.getByRole('button', { name: new RegExp(label) }));
+      expect(
+        screen.getByText(/Auto \(resolved from your configured provider\)/),
+      ).toBeInTheDocument();
+      expect(screen.queryByText(/claude-/)).not.toBeInTheDocument();
+    }
+  });
+
+  it('links to Settings, API Keys for provider configuration', () => {
+    render(<NewAgentPage />);
+    fireEvent.click(screen.getByRole('button', { name: /Content Writer/ }));
+
+    const link = screen.getByRole('link', { name: 'Settings, API Keys' });
+    expect(link).toHaveAttribute('href', '/settings/api-keys');
+  });
+});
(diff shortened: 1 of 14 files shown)


RevealUIStudio:
Security-adjacent review (internal onboarding audit 2026-07-09): APPROVED.

lastActiveAt writer: throttled 60min, fire-and-forget with its own catch, reads the already-loaded row; cannot affect auth latency or semantics. Activation analytics: cross-account aggregation correctly admin-gated on top of the entitlement gate. agent-tasks/exists: auth + ai entitlement, returns a single platform-wide boolean (negligible signal; acceptable, primary deployment is single-tenant self-host). Checklist degrades to the static list on any fetch failure.

PR #1788, opened 2026-07-09
feat(auth): establish session on successful email verification

RevealUIStudio:
## Summary

Identified in an internal onboarding audit (2026-07-09): sign-up gives no session to non-first users (`apps/admin/src/app/api/auth/sign-up/route.ts`), and the verify-email link only flips `emailVerified` then redirects to `/login?message=email_verified`, forcing a manual second login mid-purchase.

**Change:** after a valid, unexpired, single-use verification token flips `emailVerified`, the route now creates a session for that user using the exact same session-creation mechanism sign-up uses for the auto-promoted first user (`createSession` from `@revealui/auth/server`), then redirects to the destination `resolveAuthDest` would produce, preserving the `upgrade` param. Non-admins land on `/welcome`; admins land on `/`; either lands on `/account/billing?upgrade=pro|max` when the upgrade param is present.

Constraints honored:
- `upgrade` is whitelisted to `pro|max` (same pattern the route already used) — no open redirect.
- No session is created on any invalid/expired/reused/garbage token path — those all fall through to the pre-existing failure redirects, unchanged.
- A defensive branch was added for the case where `updateUser` returns no row after a matched token (e.g. a concurrent deletion) — fails closed to `verification_failed`, no session.
- The `already_verified` early-return path (a race where the token row is still present but `emailVerified` is already `true`) still redirects to `/login?message=already_verified` with no session, exactly as before.

## Test cases covered

`apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts` (13 tests):
- valid token → session created, `revealui-session` + `revealui-role` cookies set, redirect to `/account/billing?upgrade=pro` when the upgrade param is present
- valid token, no upgrade param → redirect to `/welcome` (non-admin) or `/` (admin), based on the updated user's role
- expired token → no matching row (query already excludes expired rows) → `error=invalid_token`, no session
- reused token → the token column is nulled on first use, so re-use finds no row → `error=invalid_token`, no session
- garbage token → no matching row → `error=invalid_token`, no session
- already-verified race → `message=already_verified`, no session, `updateUser` never called
- unrecognized upgrade value → ignored, no open redirect
- pre-existing paths preserved: missing token, rate-limited, rate-limit-store failure (fail closed), `updateUser` returning no row, unexpected DB failure

## Tests run

- `pnpm --filter admin vitest run src/app/api/auth/__tests__` — 8 files, 81 tests, all passing (includes the 13 in the file above plus the full auth route suite for regression)
- `pnpm --filter admin typecheck` — clean
- `pnpm gate:quick` — Phase 1 pass, CI gate pass

## Security review

**This PR awaits a security review before merge.** It changes when a session cookie is issued on an unauthenticated-adjacent surface (a GET request carrying only a mailed token). Please double-check the token-reuse and expiry reasoning above against the actual `getUserByVerificationToken` query semantics, and the upgrade-param whitelist against `resolveAuthDest`.

### DIFF
--- a/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
+++ b/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
@@ -1,18 +1,25 @@
 /**
  * Tests for GET /api/auth/verify-email
  *
- * Email verification via hashed token comparison. All paths redirect.
- * Rate limited by IP, fails closed if rate limit store is unavailable.
+ * Email verification via hashed token comparison. On a valid, unexpired,
+ * single-use token, establishes a session immediately (mirroring sign-up's
+ * auto-promoted first-user path) and redirects to the resolved destination
+ * (billing checkout when `?upgrade=` is present, `/welcome` or `/` otherwise).
+ * Every other path  -  expired, reused, garbage, or already-verified token  -
+ * must never create a session. Rate limited by IP, fails closed if the rate
+ * limit store is unavailable.
  */
 
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mockCheckRateLimit = vi.fn();
+const mockCreateSession = vi.fn();
 const mockGetUserByVerificationToken = vi.fn();
 const mockUpdateUser = vi.fn();
 
 vi.mock('@revealui/auth/server', () => ({
   checkRateLimit: (...args: unknown[]) => mockCheckRateLimit(...args),
+  createSession: (...args: unknown[]) => mockCreateSession(...args),
 }));
 
 vi.mock('@revealui/utils/logger', () => ({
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Security review (internal onboarding audit 2026-07-09): APPROVED.

Verified on the diff: session is created only after a valid hashed-token match with the token consumed in the same update; fail-closed on the vanished-row path and the catch path; upgrade param whitelisted to pro|max with internal-only destinations (no open redirect); cookie flags mirror sign-up (httpOnly, secure in prod, sameSite lax, shared domain helper); rate limiting untouched; four token test cases plus role-destination and whitelist tests present.

Noted, non-blocking: two truly concurrent requests with the same token can each mint a session for the same legitimate user before the token-null lands (sequential reuse is blocked by the consumed token). Low severity since token possession already implies account-claim possession; a conditional update (WHERE token still set) would close it if we ever care.

PR #1791, opened 2026-07-10
test(admin): fold unique verify-email assertions into colocated suite

RevealUIStudio:
## Why

Follow-up to #1790. The stale duplicate suite removed there carried three assertions the superseding colocated suite lacked. This PR re-adds them so the consolidation loses no coverage.

Stacked on the #1790 branch; once #1790 merges, this PR's diff reduces to the single test-file addition.

## What

Five appended tests in `apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts` (plus a headers-parameterized request helper):

- **SHA-256 hashing before lookup** — the raw token must never reach `getUserByVerificationToken`; asserts the hex digest is passed and the verbatim token is not.
- **Single-update token consumption** — `updateUser` is called exactly once with `emailVerified: true` and `emailVerificationToken: null`.
- **Rate-limit IP-key extraction** — first `x-forwarded-for` hop only, `x-real-ip` precedence, and the `unknown` fallback.

No production code changes; test-only.

## Review gate

This file sits under `api/auth/` — a guardrail-2 security-sensitive path — so this PR holds for a recorded non-author verdict (`sec-review:approved` label or approving review) before merge, even though it is test-only.

## Verify

- `pnpm vitest run src/app/api/auth/__tests__/verify-email-route.test.ts` in `apps/admin`: 18 passed (13 existing + 5 new).
- Biome clean on the changed file.


RevealUIStudio:
Non-author security verdict (owner-directed): **sec-review:approved**. Test-only change. It deletes a duplicate verify-email suite and folds the UNIQUE cases into the colocated suite: SHA-256 token hashing before the DB lookup, single-update token consumption, and the x-forwarded-for/x-real-ip rate-limit-key behavior. Net it-blocks in the colocated file end at 18; no unique coverage is lost and no production code is touched. Clean.

PR #1793, opened 2026-07-10
fix(paywall): make the /x402 subpath export real by extracting server helpers

RevealUIStudio:
## Why

`@revealui/paywall` (published, MIT) advertises a `@revealui/paywall/x402` subpath ("HTTP 402 agent payment negotiation") in its main entry doc comment and in its `package.json` `exports` map, but the implementation file was a doc comment plus `export {}` — a broken public contract for anyone installing the package and following the docs.

## What

- Extracted the reusable x402 protocol helpers (types, config reading, base64 encode/decode, payment-required/payment-methods builders, and facilitator-backed `verifyPayment`) out of `apps/server/src/middleware/x402.ts` into `packages/paywall/src/x402/index.ts`. This module has no dependency on any app or on `@revealui/core` — it takes an optional `X402VerifyHooks` object (`onFacilitatorWarn`, `onVerified`) so callers can wire in their own logger/metrics without the package depending on either.
- `apps/server/src/middleware/x402.ts` is now a thin glue layer: it imports the package's helpers, re-exports the config/builder functions unchanged, and wraps `verifyPayment` to bind RevealUI's structured logger and Prometheus-style metrics via the hooks. No env var name, default, or enablement logic changed (`X402_ENABLED` still defaults to false).
- Added `@revealui/paywall` as a workspace dependency of `apps/server`.
- Added package-level unit tests at `packages/paywall/src/x402/__tests__/index.test.ts` (20 tests) covering config reading, encode/decode, payment-required/payment-methods building, and `verifyPayment` including the hooks contract.
- Added a changeset (`@revealui/paywall`: minor) since this adds real capability to a published package.
- Updated `docs/architecture/x402.md`: its Works-Cited citation pointed at a now-nonexistent line in the shrunken middleware file; repointed it (and the Status table / Cross-references) at the new package location.

The server's existing `apps/server/src/middleware/__tests__/x402.test.ts` (21 tests) required **zero changes** — same import path, same exported names, same behavior.

## Holds

This touches payment surfaces. Holds for a recorded security verdict before merge per review policy.

## Verify

```
pnpm --filter @revealui/paywall test        # 129 passed (9 files), incl. new x402 suite (20 tests)
pnpm --filter server exec vitest run src/middleware/__tests__/x402.test.ts   # 21 passed, unchanged
pnpm --filter @revealui/paywall typecheck   # clean
pnpm turbo build --filter=@revealui/paywall --filter=server   # both build clean
pnpm gate:quick                              # PASS (Phase 1)
```

### DIFF
--- a/.changeset/paywall-x402-subpath.md
+++ b/.changeset/paywall-x402-subpath.md
@@ -1,0 +1,5 @@
+---
+"@revealui/paywall": minor
+---
+
+Implement the `@revealui/paywall/x402` subpath. Previously the package advertised this export in its docs but shipped only a stub (`export {}`); it now ships the real HTTP 402 agent payment negotiation helpers: `getX402Config`, `getAdvertisedCurrencyLabel`, `encodePaymentRequired`, `buildPaymentRequired`, `buildPaymentMethods`, and `verifyPayment` (which verifies USDC-on-Base payments via a facilitator and accepts optional `X402VerifyHooks` for logging/metrics). These were extracted from RevealUI's own server middleware, which now imports them from this package instead of maintaining its own copy.
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -16,6 +16,7 @@
     "@revealui/db": "workspace:*",
     "@revealui/mcp": "workspace:*",
     "@revealui/openapi": "workspace:*",
+    "@revealui/paywall": "workspace:*",
     "@revealui/security": "workspace:*",
     "@sentry/node": "^10.63.0",
     "drizzle-orm": "catalog:",
--- a/apps/server/src/middleware/x402.ts
+++ b/apps/server/src/middleware/x402.ts
@@ -1,17 +1,18 @@
 /**
- * x402 Native Payment Middleware Utilities (Phase 5.2)
+ * x402 Native Payment Middleware (Phase 5.2) — RevealUI server glue.
  *
- * Implements HTTP-native micropayments via the x402 protocol:
- * https://x402.org
- *
- * Supports USDC on Base (EVM) — verified via the Coinbase facilitator.
+ * The reusable x402 protocol logic (config reading, payload encode/decode,
+ * payment-required builders, and facilitator verification) lives in
+ * `@revealui/paywall/x402`, published as part of the MIT `@revealui/paywall`
+ * package. This file is the thin app-specific layer that wires the RevealUI
+ * structured logger and Prometheus-style metrics into payment verification.
  *
  * Payment flow:
- *   1. Agent request arrives → quota exhausted
+ *   1. Agent request arrives, quota exhausted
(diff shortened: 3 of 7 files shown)


RevealUIStudio:
**Independent adversarial security pass (Fable, dispatcher of the implementing agent — this comment informs the merge verdict; it is not the recorded non-author verdict itself).**

Reviewed the extraction line-by-line against the pre-change middleware (comment-stripped structural diff of `apps/server/src/middleware/x402.ts@merge-base` vs `packages/paywall/src/x402/index.ts`). Findings:

**Behavior parity: confirmed.** The entire delta is (1) types exported, (2) `BASE_USDC_ADDRESS` hoisted to satisfy `noUncheckedIndexedAccess` (same literal, same fallback semantics), (3) direct `logger.warn` / `trackX402PaymentVerify` calls replaced by an optional `X402VerifyHooks` interface. The server glue re-binds both hooks with the exact original arguments, including the `'exact'` scheme label and `valid`/`invalid` mapping. Decode failures remain uncounted, as before.

**Security-relevant invariants checked and unchanged:** facilitator endpoint and request shape; fail-closed on facilitator non-OK and on request error; resource URL equality check between the 402 challenge and the payload; `X402_ENABLED` default-false gating; no secret material handled anywhere in the moved code. Publishing the logic in the MIT package discloses nothing new: `apps/server` is already public source.

**Package boundary sound:** `@revealui/paywall` gains no dependency on `@revealui/core` or any app; observability is injected by the consumer. The pre-existing `exports` map now resolves to real content, which is exactly the contract fix.

**Tests:** 21 pre-existing middleware tests pass with zero diff to the test file; 20 new package tests cover config, encode/decode, builders, and the hooks contract.

One process note for the record: the review-gate path scanner did not flag this PR because neither `packages/paywall/` nor `middleware/x402` is on its sensitive-path list. That is a scanner coverage gap, being fixed separately; this PR correctly self-declared the hold in its body.

Verdict from this pass: **approve**. Holding for the recorded non-author verdict per policy.

RevealUIStudio:
Non-author security verdict (owner-directed): **sec-review:approved**. This was the fleet-checker false pass, but the flag was the in-repo gate path list missing `middleware/x402`, not a code defect. Adversarial read of the extraction: the `verifyEvmPayment` moved into `@revealui/paywall/x402` is semantically identical to the server original — same facilitator `/verify` POST, same 10s timeout, and fail-closed on every non-happy path (`!resp.ok`, `!result.isValid`, and the network catch all return `{valid:false}`). The only change is that `logger.warn`/`trackX402PaymentVerify` became injectable `X402VerifyHooks`, which the server glue supplies. USDC addresses, atomic-unit conversion, and config defaults moved verbatim. Gated behind `X402_ENABLED` (default false). Clean.

PR #1797, opened 2026-07-10
fix(server): keep purchased tier when perpetual support lapses

RevealUIStudio:
## Summary

A perpetual license is sold as permanent ownership of a tier ("License key never expires"). `POST /api/license/verify` did the opposite when the annual support contract lapsed: it returned `getFeaturesForTier('free')`, `maxSites: 1`, `maxUsers: 3`. An Enterprise Perpetual customer who declined to renew support reported as a free deployment.

A support lapse **freezes** the purchased tier, it does not revoke it. What stops is update delivery and support, and neither is gated on this path. The `support_expired` response now carries the license's own tier features and limits, and still sets `supportExpired: true` so clients can prompt for renewal.

This is a live surface, not a hypothetical: in-app perpetual and renewal purchase paths are already shipped.

## Changes

- `apps/server/src/routes/license.ts` — the `support_expired` branch reuses the tier's `features` / `maxSites` / `maxUsers` instead of the free-tier constants. The tier-limit computation is hoisted above the branch so both paths share one definition.
- `apps/server/src/routes/license.ts` — the OpenAPI `reason` description no longer promises a downgrade that no longer happens.
- `apps/server/src/routes/billing.ts` — two comments describing the sweep's `support_expired` status said "premium features downgrade to free". Comments only, no behavior change.
- `apps/server/src/routes/__tests__/license.test.ts` — four new cases.

## Tests

There was no coverage of the support-expired path at all before this. Added:

| Case | Asserts |
|---|---|
| Lapsed perpetual (`supportExpiresAt` past) | tier `enterprise` retained, `ai`/`analytics` true, limits `null`/`null`, `supportExpired: true` |
| DB row already swept to `support_expired` | tier `pro` retained, limits 5/25, `supportExpired: true` |
| Perpetual with active support | `reason: valid`, `supportExpired: false`, `expiresAt: null` |
| Subscription license | unchanged: `reason: valid`, `expiresAt` from `exp`, no support contract read |

Verified the tests fail against the pre-fix route (the two lapse cases fail; the two regression guards pass either way), so they genuinely pin the behavior.

```
Test Files  5 passed (5)
     Tests  63 passed (63)
```
(license, license-edge, webhook-license-roundtrip, health-license, middleware/support-expiry)

`billing-comprehensive` 72/72. `pnpm --filter server typecheck` clean. Biome clean.

## Scope note for the reviewer

The **identical demotion exists on the enforcement path** and is deliberately not fixed here. `checkSupportExpiry` (`apps/server/src/middleware/license.ts:485-497`) sets `tier: 'free'` and `features: {}` once the 30-day grace is exhausted, and it is mounted globally on `/api/*`, `/api/v1/*` and `/a2a/*` (`index.ts:702-709`), feeding `requireFeature` / `requireAIAccess` / task-quota. So after this PR the API correctly *reports* the retained tier while the middleware still *denies* those features past grace.

That is a coherent intermediate state (the report becomes honest) but not the end state. It is separated because it needs a posture ruling this PR has no mandate to make: `packages/core/src/license.ts:36-38` defines `read-only` ("reads allowed, writes blocked") and `expired` ("degraded to free tier") as different modes, and the middleware currently sets the `read-only` header while applying `expired` semantics. Whether a lapsed perpetual past grace is read-only or full read-write at the frozen tier is an open product decision. Tracked internally; `tier: 'free'` + `features: {}` is wrong under either answer.

## Risk

Low. The change only widens what a lapsed perpetual receives, and only on a reporting endpoint. No consumer read `supportExpired` or depended on the free-tier response shape (checked across the repo). Subscription and revoked/expired/unverifiable paths are untouched and still covered by their existing tests.

Licensing surface, so this needs a recorded security verdict from a non-author party before merge.

Internal audit reference: the lapsed-support licensing review.

### DIFF
--- a/apps/server/src/routes/__tests__/license.test.ts
+++ b/apps/server/src/routes/__tests__/license.test.ts
@@ -326,7 +326,7 @@ describe('POST /verify  -  DB revocation override', () => {
     exp: Math.floor(Date.now() / 1000) + 86400,
   };
 
-  function mockDb(rows: { status: string }[]) {
+  function mockDb(rows: { status: string; supportExpiresAt?: Date | null; perpetual?: boolean }[]) {
     vi.mocked(getClient).mockReturnValueOnce({
       select: () => ({
         from: () => ({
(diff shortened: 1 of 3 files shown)


RevealUIStudio:
Non-author security verdict (owner-directed): **sec-review:approved**. Adversarial read of the diff: the `support_expired` path in `license.ts` now returns `getFeaturesForTier(payload.tier)` + tier-appropriate limits instead of the free-tier downgrade, which is exactly the GAP-308 ruling (perpetual = permanent ownership; a lapsed support contract freezes the purchased tier, it does not revoke it). Four tests pin it: perpetual-lapsed keeps enterprise + unbounded limits, DB-marked `support_expired` keeps pro, active support reads `valid`, and non-perpetual subscriptions are untouched. It changes only what `/verify` returns; it does not touch the renewal/checkout path, so the renewal-trap concern does not apply. Clean.

PR #1811, opened 2026-07-11
feat(server): read-only write-gate for lapsed perpetual support (GAP-310)

RevealUIStudio:
## What

Implements read-only enforcement for a lapsed-perpetual-support license (GAP-310): retain the purchased tier for **reads**, block **writes**, with a precise, fail-safe write classifier and a staged rollout flag.

## The defect (grounded on `test` @ 1811154b0)

`apps/server/src/middleware/license.ts`, on grace exhaustion, set `tier: 'free', features: {}` while emitting `X-License-Mode: read-only`. The header lied about the entitlements; downstream feature gates don't distinguish read from write, so an empty `features` map blocked the customer's own **reads**. That contradicts the core `LicenseMode` contract (`packages/core/src/license.ts`), where `read-only` retains the tier and blocks only writes.

## The fix

- **`checkSupportExpiry`** now, on grace exhaustion in `enforce` mode, retains `payload.tier` + `getFeaturesForTier(tier)` and sets `readOnly: true`; in `off`/`shadow` it preserves the historical free-tier downgrade. The tier is restored **only when writes are actually blocked** — both halves land together (the gap invariant; restoring the tier alone would be a strict loosening).
- **`enforceReadOnlyWrites`** — a new gate mounted after `checkSupportExpiry` and before the feature gates on `/api/*`, `/api/v1/*`, `/a2a/*`. Classifies via **exempt-set → route-override → HTTP-method baseline** (zero authored regex), and in `enforce` rejects a non-exempt write with `403` + an actionable renew message.
- **Exempt set** (the load-bearing safety piece): renewal checkout, auth/session, license verify/features, Stripe webhooks, pricing/health. A miss here is a lockout (can't renew) or a leak, so it errs toward the clearly-safe read/escape routes.
- **`LICENSE_READ_ONLY_ENFORCE`** gates the whole behavior: `off` (default — **byte-identical to today**) / `shadow` (logs would-be blocks, no loosening) / `enforce`.

## Design decisions (departures from the spec, both deliberate)

The design spec is `.jv docs/gap-specs/GAP-310-write-classification-design.md`. Two points were corrected during the adversarial read:

1. **Shadow vs. the no-loosening invariant.** The spec (§3.5) made tier-restore unconditional with only the write-*block* flagged — which at runtime loosens access during the shadow window (the gap forbids this). Resolved owner-side: **gate both halves** on the same flag, so shadow observes without loosening and enforce restores+blocks atomically. Customer-facing honesty is already covered by the #1797 validate route.
2. **Classifier ordering.** The spec (§3.2) ordered exempt → method → override, but an override can't override a method baseline that matched first. Implemented as exempt → **override → method**.

## Rollout

`enforce` is an **owner-gated** flip. This PR ships `off` by default; recommended path is shadow → confirm the exempt log is clean → enforce.

## Tests (red-first)

7 new cases in `support-expiry.test.ts`: reads retain tier+features, writes → 403, exempt escape hatches reachable, `/a2a` dispatch(POST)=block vs poll(GET)=allow, shadow logs without loosening, `off` is inert, within-grace unaffected. **Proven red** by removing the gate: exactly the 7 new tests fail, the 9 pre-existing pass. `tsc --noEmit` and Biome clean.

## Merge gate

Licensing surface — **guardrail-2 requires a recorded non-author verdict before merge** (`sec-review:approved` label). GAP-308 (validate route) is already done (#1797); GAP-310 closes when `enforce` ships.

### DIFF
--- a/apps/server/src/index.ts
+++ b/apps/server/src/index.ts
@@ -64,6 +64,7 @@ import { errorHandler } from './middleware/error.js';
 import {
   checkLicenseStatus,
   checkSupportExpiry,
+  enforceReadOnlyWrites,
   requireAIAccess,
   requireDomain,
   requireFeature,
@@ -709,6 +710,14 @@ app.use('/api/v1/*', supportExpiryCheck);
 app.use('/a2a/*', entitlementMiddleware());
 app.use('/a2a/*', licenseStatusCheck);
 app.use('/a2a/*', supportExpiryCheck);
+
+// GAP-310: block writes for lapsed-perpetual (read-only) licenses. Runs after
+// checkSupportExpiry (which sets the read-only signal) and before the feature
+// gates + route handlers. Governed by LICENSE_READ_ONLY_ENFORCE
+// (off default / shadow / enforce); inert unless a license is in read-only mode.
+app.use('/api/*', enforceReadOnlyWrites());
+app.use('/api/v1/*', enforceReadOnlyWrites());
+app.use('/a2a/*', enforceReadOnlyWrites());
 
 // License enforcement  -  gate premium routes by feature
 // Agent stream + tasks: free tier allowed with local inference, Pro+ for cloud providers
--- a/apps/server/src/middleware/__tests__/support-expiry.test.ts
+++ b/apps/server/src/middleware/__tests__/support-expiry.test.ts
@@ -15,6 +15,11 @@ vi.mock('@revealui/core/license', () => ({
 vi.mock('@revealui/core/features', () => ({
   isFeatureEnabled: vi.fn(() => true),
   getRequiredTier: vi.fn(() => 'pro'),
+  // Non-free tiers get real features; free is empty. Lets the read-only tests
+  // assert that a lapsed license retains (or loses) features per mode.
+  getFeaturesForTier: vi.fn((tier: string) =>
+    tier === 'free' ? {} : { ai: true, dashboard: true, analytics: true },
+  ),
 }));
 
 vi.mock('@revealui/core/observability/logger', () => ({
@@ -22,7 +27,8 @@ vi.mock('@revealui/core/observability/logger', () => ({
 }));
 
 import { getLicensePayload } from '@revealui/core/license';
-import { checkSupportExpiry, resetSupportExpiryCache } from '../license.js';
+import { logger } from '@revealui/core/observability/logger';
+import { checkSupportExpiry, enforceReadOnlyWrites, resetSupportExpiryCache } from '../license.js';
 
 const mockedGetLicensePayload = vi.mocked(getLicensePayload);
 
@@ -66,6 +72,8 @@ function createApp(
 
 afterEach(() => {
   resetSupportExpiryCache();
+  delete process.env.LICENSE_READ_ONLY_ENFORCE;
+  vi.mocked(logger.info).mockClear();
 });
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


RevealUIStudio:
## Guardrail-2 verdict: **APPROVE** (non-author, Fable 5)

Independent adversarial review of the GAP-310 read-only write-gate build, grounded to PR head `797ad89` and re-derived against `origin/test` (merge-base `1811154`). I am not the author. Ran on Fable end to end, no safeguard pause. Fleet checker confirms the surface (`apps/server/src/middleware/license.ts`) is security-sensitive and correctly HOLDs pending this recorded verdict.

### What I verified from the code (not the PR body)

1. **The loosening invariant holds — both halves gated on one mode read.** `checkSupportExpiry` restores the purchased tier + features (and sets `readOnly: true`) ONLY in `enforce`; in `off`/`shadow` it preserves the historical `free` downgrade. The write-gate signal `readOnlyPending` is set in the same block from the same `getReadOnlyEnforcementMode()` result, so the tier-restore and the write-block can never diverge within a request. Restoring the tier without the block would be a strict loosening; the code makes that unrepresentable.
2. **`off` is a true no-op.** `readOnlyPending` is not set when mode is `off`, so `enforceReadOnlyWrites` short-circuits to pass-through and behavior is byte-identical to pre-GAP-310. The gate ships dark and is owner-flipped via `LICENSE_READ_ONLY_ENFORCE` — safe to merge because nothing changes until the env is set (the rollout stays owner-gated).
3. **Signal coupling is sound.** `enforceReadOnlyWrites` mounts on `/api/*`, `/api/v1/*`, `/a2a/*` (index.ts:718-720) — the exact prefixes `supportExpiryCheck` covers (704-712), and the setter is registered BEFORE the gate, so no path reaches the gate without its signal-setter having run. `readOnlyPending` is only ever produced by `checkSupportExpiry`; on any other path `pending` is undefined and the gate passes through.
4. **Exempt set is the load-bearing safety surface, and it's complete for the failure directions that matter.** Renewal checkout, all auth/session prefixes, license verify/features (reads-over-POST), machine webhooks, pricing, and `/health` are exempt — so a lapsed customer can always sign in and buy back support (no lockout). Method baseline (POST/PUT/PATCH/DELETE = write, case-normalized) with an empty override table that grows only on an observed shadow-mode misclassification, per the design.
5. **Zero authored regex** in the new code (verified by grep for `RegExp`, `/.../ `, and regex-arg `.match/.replace/.split`) — prefix `startsWith`/equality only, per the fleet no-regex rule.

### Red-first, run by me (the #82/#403 standard)

Detached worktree at `797ad89`; built the server dep graph via turbo (14 pkgs); baseline `support-expiry.test.ts` **16/16 green**. Three mutations, each restored before the next:

- **Write-block neutered** (fall through instead of 403 in `enforce`) → **2 red**: `write blocked with 403` + `/a2a dispatch blocked / poll allowed`.
- **Renewal route removed from the exempt set** → **1 red**: `renewal/auth/verify/webhook stay reachable (no lockout)` — proving that test actually guards the renewal escape hatch (403 where 200 is required).
- **Tier restored in `off`/`shadow` too** (break the loosening invariant) → **2 red**: `off: tier downgraded` + `shadow: tier stays downgraded` — proving the coupling is pinned.

Restored: 16/16 green again. Full server middleware suite **296/296 (22 files)**. `tsc --noEmit`: **zero errors in the PR's files** (`license.ts`, `index.ts`); the remaining tsc noise is entirely `Cannot find module '@revealui/ai'` — the gitignored Pro package absent from a public-repo local checkout but present on CI (which is why the PR's server typecheck check is green).

### Non-blocking observations (optional follow-ups, not defects)

- Three exempt entries lack a trailing slash (`/api/license/verify`, `/api/license/features`, `/api/pricing`), so `startsWith` would also exempt a hypothetical sibling like `/api/pricing-preview`. The failure direction is "too permissive on a read/escape route," the lowest-risk direction, and no such sibling routes exist today. If you want it airtight, switch those three to exact-match or add a trailing slash. Not merge-blocking.
- `READ_ONLY_OVERRIDES` ships empty by design (grows only on an observed shadow-mode misclassification) — correct; noting it so the next reviewer doesn't read the empty array as an omission.

### Disposition

Clears guardrail-2 on my independent reading. The `enforce` path blocks non-exempt writes while retaining reads and never locks out renewal; `off` is inert; the loosening invariant is coupled and test-pinned red-first. The owner applies `sec-review:approved` + merges on this verdict; I do not self-merge or label. Rollout to `enforce`/`shadow` remains an owner-gated operation (correctly dark-by-default).

*Recorded by: agent-system-fable (Fable 5, non-author). Worktree removed after the run.*

RevealUIStudio:
**Guardrail-2 verdict (Fable, non-author, adversarial read at 797ad8902): APPROVE — owner applies the gate clearance and merges.**

Re-derived against the GAP-310 design and the code:

- **The build correctly resolved a self-contradiction in the design.** §3.5 said the tier-restore is unconditional with only blocking staged behind the flag, but §1's own invariant forbids exactly that (tier restored + enforcement off = full read-write premium access for a lapsed license — the strict loosening). The shipped behavior sides with §1: the restore and the block ride the same mode, and shadow keeps the historical downgrade so nothing loosens while the exempt set trains on real logs. That is the safe reading; the design doc gets a one-line amendment so §3.5 stops contradicting §1 (I am committing it now).
- **The exempt set covers every named escape hatch** with both /api and /api/v1 prefixes: the renewal checkout (the no-lockout guarantee), the auth trio, license verify/features (reads-over-POST), machine webhooks (renewal settlement), pricing, health. Prefix matching is zero-regex, and I checked the two prefix-collision candidates (/api/pricing has no sibling routes; /health over-match is harmless).
- **The a2a ruling is satisfied by construction, verified against the routes:** the a2a surface is verb-differentiated OpenAPI (ten GETs for cards/discovery/status, POST only for dispatch and message-send, one PUT/DELETE), so the method baseline yields exactly the design's dispatch=write, polling=read split; the JSON-RPC strings are envelope content, not the transport. The empty override manifest with shadow-log-then-add is the right maintenance posture.
- **Gate placement verified:** mounted on /api/*, /api/v1/*, and /a2a/* after checkSupportExpiry sets the signal and before the feature gates; off-mode sets no signal so the gate is a true pass-through; the 403 carries the honest message plus the renewal pointer.
- **Red-first re-run from scratch** in a detached worktree (after building the server dep graph — the initial resolution failures were fresh-worktree dist gaps, not the PR): baseline 16/16; classifier neutered → 3 write-block tests red; tier-features neutered → the read-retention test red; restored → 16/16.
- Non-regressions pinned: subscription expiry still demotes, within-grace unchanged, off-mode byte-identical to pre-GAP-310.

Rollout as designed: ship with the flag off, run shadow until the would-block log is clean of renewal/auth/webhook entries, then enforce. This comment is my verdict; I apply no label.

PR #1818, opened 2026-07-11
fix(admin): harden checkout auto-fire with backoff retry and failure log

RevealUIStudio:
## Problem

In the paid signup flow (signup -> verify -> login -> auto-checkout), the post-login auto-fire lands on `/account/billing?upgrade=pro|max` and gates on a subscription/plan fetch before firing checkout to Stripe.

The base already retried that fetch twice and rendered a visible "Continue to checkout" fallback, so it was **not** a silent dead end. But two spec requirements were unmet:

- The retry fired **immediately** with no backoff, so a brief connection blip or a cold API instance was unlikely to be survived by a second instant attempt.
- A persistent failure was **never logged**, leaving no operational trail when a buyer got stuck at the fallback.

Silent-path / gating code:
- [page.tsx:144-150](apps/admin/src/app/(frontend)/account/billing/page.tsx#L144) - the pre-change `fetchSubscription` (2 immediate attempts, no log).
- [page.tsx:188-196](apps/admin/src/app/(frontend)/account/billing/page.tsx#L188) - the auto-checkout effect that only fires once `subscription` is populated.
- [page.tsx:380-392](apps/admin/src/app/(frontend)/account/billing/page.tsx#L380) - the pre-existing visible fallback CTA.

## Change

- **Bounded retry with backoff.** `fetchSubscription` now loops up to `MAX_SUBSCRIPTION_FETCH_ATTEMPTS` (3) with a short linear backoff (`SUBSCRIPTION_RETRY_BASE_DELAY_MS` 150ms then 300ms) between attempts, breaking on the first success.
- **Log the persistent failure.** On exhausting the retries it calls the shared `logger.error` (no `console.*`) with `{ attempts, plan }`, then keeps the existing `subscriptionLoadFailed` trail that drives the visible fallback.
- Visible fallback CTA copy is unchanged (full-clause "Continue to checkout"); no new user-facing strings, no new dependencies.

### Reuse note
`@revealui/resilience` has a retry helper, but it imports `node:crypto` (browser-unsafe) and is not an admin dependency, so I extended the file's existing hand-rolled retry rather than pull it into a client bundle.

## Tests

`pnpm --filter admin exec vitest run` on the billing page suite - **5 passed**.

Added `bounds the retry at three attempts and logs the persistent failure`, which asserts 3 subscription attempts and a `logger.error` call. It also fixes the `next/navigation` mock to return a **stable** router reference (matching real `useRouter`), so effects listing `router` in deps do not re-fire per render.

**Red-first: yes.** Stashed `page.tsx`, ran the new test against the base file - it failed `expected 3, received 2` (base does 2 attempts and never logs). Restored the fix; green.

`pnpm --filter admin exec tsc --noEmit` clean; `pnpm --filter admin build` clean; Biome clean on both touched files.

## Security checker

`security-pr-review-check.js --diff origin/test` -> **"no security-sensitive files - no coordinator gate"** (pass).

### DIFF
--- a/apps/admin/src/app/(frontend)/account/billing/__tests__/page.test.tsx
+++ b/apps/admin/src/app/(frontend)/account/billing/__tests__/page.test.tsx
@@ -7,9 +7,13 @@ vi.mock('@revealui/auth/react', () => ({
 }));
 
 const mockPush = vi.fn();
+// Stable reference, matching real next/navigation useRouter(). A fresh object
+// per render would make effects that list `router` in their deps re-fire on
+// every re-render, which does not happen in production.
+const mockRouter = { push: mockPush };
 let searchParams = new URLSearchParams();
 vi.mock('next/navigation', () => ({
-  useRouter: () => ({ push: mockPush }),
+  useRouter: () => mockRouter,
   useSearchParams: () => searchParams,
 }));
 
@@ -26,6 +30,11 @@ vi.mock('@/lib/utils/safe-stripe-redirect', () => ({
   safeStripeRedirect: (url: string) => mockSafeStripeRedirect(url),
 }));
 
+const mockLoggerError = vi.fn();
+vi.mock('@revealui/utils/logger', () => ({
+  logger: { error: (...args: unknown[]) => mockLoggerError(...args) },
+}));
+
 import BillingPage from '../page';
 
 afterEach(() => {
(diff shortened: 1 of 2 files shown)


RevealUIStudio:

The audit honesty is the best part: the premise said no-retry-no-fallback, the code said otherwise (both shipped in #1787), and the PR scoped itself to the two genuine residuals — backoff between attempts and logging the persistent failure. The bounded loop is clean (3 attempts, linear 150ms backoff, break on success), the failure trail keeps driving the pre-existing visible fallback CTA, and the shared logger is used instead of console. Declining @revealui/resilience was right: it imports node:crypto and is not an admin dependency; extending the file's existing loop is the smaller, browser-safe change. The stable-router mock fix is a correct test-hygiene catch (a fresh object per render re-fires effects that list router in deps — a mock artifact, not production behavior). Red-first proof recorded (expected 3, received 2 against base). Security checker: clear.

PR #1820, opened 2026-07-11
fix(ai): resolve built-in agent model from configured provider

RevealUIStudio:
## Problem

The first-run agent surface presumed a provider the configuration UI cannot supply. The built-in platform agents hardcoded proprietary Claude model ids, while the API Keys page (and the whole platform) offers only open-model providers. The create-agent flow never surfaced the provider-key prerequisite, so a user could scaffold an agent that then failed at first dispatch.

## Audit (file:line, both sides)

Config side (providers the key page actually offers):
- `apps/admin/src/app/(backend)/settings/api-keys/page.tsx:11-45` — `PROVIDERS` = `inference-snaps`, `ollama`, `groq`, `huggingface`; initial provider `inference-snaps`.
- `apps/admin/src/app/api/user/api-keys/route.ts:12` — POST schema enum restricts to the same four.
- `packages/contracts/src/providers.ts:8` — canonical `LLM_PROVIDERS` (open models only; the file comment states no proprietary providers).

Template side (hardcoded Claude):
- `packages/ai/src/a2a/card.ts:26` — `THE_CREATOR_DEF.model = 'claude-opus-4-6'`.
- `packages/ai/src/a2a/card.ts:86` — `TICKET_AGENT_DEF.model = 'claude-sonnet-4-6'`.

Already-aligned surfaces (for contrast, unchanged):
- `apps/admin/src/app/(backend)/agents/new/page.tsx:32,51-91` — create-agent templates already default to `AUTO_MODEL` ('').
- `apps/admin/src/lib/components/Agent/index.tsx:53-62` — chat model picker uses `ollama` / `huggingface`.

Why the hardcoded ids were dead as well as wrong:
- `packages/contracts/src/a2a/index.ts:346-370` — `agentDefinitionToCard` never emits `model`, so the A2A card output never carried the Claude id.
- `packages/ai/src/a2a/handler.ts:136-143` — dispatch calls the provider-resolved `llmClient`, never `def.model`.
- `packages/ai/src/llm/client.ts:597-662` — `createLLMClientFromEnv` resolves the provider from the open-model set (default `inference-snaps`).

## What changed

1. `packages/ai/src/a2a/card.ts` — both built-in agents now default to an exported `PROVIDER_RESOLVED_MODEL` sentinel (empty) so the model resolves from the account's configured open-model provider. Claude is no longer the hardcoded default; Claude and other ids remain valid for anyone who wires a compatible provider (the model field, pricing tables, and collab passthrough still accept them).
2. `apps/admin/src/app/(backend)/agents/new/page.tsx` — the create flow fetches the account key metadata on mount. When no provider key is configured it renders a `Callout` (reused `@revealui/presentation` warning primitive, no new local UI) linking to the API Keys page and blocks submit, instead of failing later. A flaky/failed fetch leaves the prerequisite unresolved and never blocks.
3. `apps/admin/src/lib/agents/inference-prerequisite.ts` — pure `resolveInferencePrerequisite` helper plus the shared warning copy.

## User-facing copy added

- Title: "Connect an inference provider first"
- Body: "Your agents run on the inference provider you configure. Add a provider key on the API Keys page and every agent you create will use it automatically."
- Link: "Open API Keys"

## Tests

- `packages/ai/src/a2a/__tests__/a2a.test.ts` — new guard: both built-in agents use `PROVIDER_RESOLVED_MODEL` and their serialized definitions name no proprietary vendor. Proven red without the fix (`expected 'claude-opus-4-6' to be ''`), green with it. Full suite: 14 passed.
- `apps/admin/src/lib/agents/__tests__/inference-prerequisite.test.ts` — new: prerequisite resolution for null vs each offered provider, plus a copy check (no em dash, no proprietary vendor). 4 passed.
- `@revealui/ai` build: clean. Admin `tsc --noEmit`: clean.
- `pnpm exec turbo build --filter=admin` fails at `next build --turbopack`, but the failure is pre-existing on the base branch and unrelated to this change: the error is an Edge Runtime incompatibility in `apps/admin/src/instrumentation.ts` (`process.exit` / `process.stderr`) reached via `packages/mcp/dist/oauth.js` → `remote-servers/route.ts`. None of those files are touched here (zero diff vs base), and the same build fails identically at base HEAD (verified by stashing this change and rebuilding: exit 1 at the same stage).

### DIFF
--- a/apps/admin/src/app/(backend)/agents/new/page.tsx
+++ b/apps/admin/src/app/(backend)/agents/new/page.tsx
@@ -9,10 +9,15 @@ import {
   LinkButton,
   Textarea,
 } from '@revealui/presentation';
-import { Field, Label } from '@revealui/presentation/client';
+import { Callout, Field, Label } from '@revealui/presentation/client';
 import Link from 'next/link';
 import { useRouter } from 'next/navigation';
-import { type ChangeEvent, useReducer } from 'react';
+import { type ChangeEvent, useEffect, useReducer, useState } from 'react';
+import {
+  INFERENCE_PREREQUISITE_COPY,
+  type InferenceKeyMetadata,
+  resolveInferencePrerequisite,
+} from '@/lib/agents/inference-prerequisite';
 import { LicenseGate } from '@/lib/components/LicenseGate';
 import { apiFetch } from '@/lib/utils/csrf';
 
@@ -152,6 +157,27 @@ export default function NewAgentPage() {
 
   const [state, dispatch] = useReducer(formReducer, initialState);
   const { selectedTemplate, name, description, systemPrompt, submitting, error } = state;
+
+  // Resolve the inference-provider prerequisite up front. `null` means the check
+  // has not resolved yet (still loading, or the request failed) — we only block
+  // when we positively know no provider key is configured.
+  const [providerConfigured, setProviderConfigured] = useState<boolean | null>(null);
+
+  useEffect(() => {
+    let active = true;
+    fetch('/api/user/api-keys')
+      .then((r) => (r.ok ? r.json() : null))
+      .then((data: InferenceKeyMetadata | null) => {
+        if (active) setProviderConfigured(resolveInferencePrerequisite(data).configured);
+      })
+      .catch(() => {
+        // Leave the prerequisite unresolved on a fetch error so a flaky request
+        // never blocks agent creation.
+      });
+    return () => {
+      active = false;
+    };
+  }, []);
 
   function applyTemplate(key: TemplateKey) {
     const tpl = TEMPLATES.find((t) => t.key === key);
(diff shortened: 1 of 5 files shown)


RevealUIStudio:

The audit's key finding reframes the fix correctly: the hardcoded Claude ids in the built-in agent definitions were dead values (the card mapper drops model and dispatch resolves the provider client), so the right change was exactly what shipped — a provider-resolved sentinel matching the existing AUTO_MODEL convention, with Claude remaining a valid configurable option rather than the default. That restores the agnosticism canon without a provider-framework refactor.

The prerequisite surfacing is well judged: it blocks creation only on positive knowledge that no key is configured, and a failed metadata fetch leaves the gate open rather than punishing the user for a flaky request. Copy is full-clause, dash-free, and reads aloud naturally ("Your agents run on the inference provider you configure."). Red-first proof recorded on the sentinel guard test.

The flagged pre-existing admin turbopack build failure on base is being tracked separately; it is verified as zero-diff-related here.

PR #1724, opened 2026-07-03
feat(admin): pages read+write through the canonical site-scoped model

RevealUIStudio:
## What (PR 2/3 of the Pages-model unification)

Builds on the engine write-dispatch seam (#1721): the admin CMS `pages` collection now reads AND writes the canonical site-scoped `pages` table through the typed Drizzle bridge, and the collection/render surfaces are refit to the canonical-direct shape. Follows the internal audit of 2026-07-02; collection shape and bridge placement are owner-approved decisions.

### Typed bridge (`apps/admin/src/lib/db/typedCollectionStorage.ts`)
- Handler types + the returned adapter gain `create`/`update`/`delete` (same `undefined` = not-handled contract as the read seam).
- New `pages` handler backed by `@revealui/db/queries/pages`:
  - **Reads**: `findByID`/`find` map canonical rows to collection docs; `_status` mirrors the `status` column so `authenticatedOrPublished`, the drafts UI, and `revalidatePage` keep their contract. The where-parser flattens the access-merged `{and:[...]}` shape (core merges the published-only filter *before* dispatch) and translates `_status` → `status`; shapes it can't express (`or`, non-equals) signal not-handled. Soft-deleted rows are always excluded.
  - **Writes**: `siteId` defaults server-side to the fleet-marketing site; `path` derives from `slug` (and stays in lockstep on slug changes unless explicitly pinned); `blocks`/`seo` are stored directly in their jsonb columns; `blockCount` maintained; delete is the canonical soft-delete. Handled-but-not-found and validation failures throw (never a silent fall-through to dynamic SQL, which cannot satisfy `site_id NOT NULL`).

### Canonical-direct collection + render refit
- `Pages` collection: the hero group and `layout`/`meta` tabs are replaced by a direct `blocks` field — with the new **Hero block** (field-for-field the legacy hero group) available at the top of the array — and a direct `seo` tab. Dead hero/layout write path deleted.
- Generated `Page` type updated accordingly (`blocks`/`seo` + `HeroBlock`); all `Page['layout']`/`Page['hero']` consumers renamed.
- `RenderBlocks` renders `hero` blocks via the existing `RenderHero`; the schema-adapter's generic component-block fallback validates them with no adapter change.
- `(frontend)/[slug]/page.tsx` renders from `blocks` only; `generateMeta` reads `seo` with a legacy `meta` fallback (posts still use `meta`).
- vitest config resolves the `cssVariables` tsconfig alias (newly reachable through RenderBlocks → RenderHero → Media).

[...]

PR #1821, opened 2026-07-11
feat(admin): route posts read+write through the typed Drizzle bridge

RevealUIStudio:
## What

Extend the typed-storage collection bridge to **posts**, the twin of the pages read+write bridge (#1724). Posts writes previously rode the engine's dynamic-SQL path with the camelCase-column trap; they now go through a typed Drizzle handler against the `posts` schema. Follows the internal audit of 2026-07-02 and the merged pages-bridge pattern.

### The old dynamic-SQL posts path (the trap)

The engine builds SQL column identifiers straight from the camelCase collection field keys:

- create: `packages/core/src/collections/operations/create.ts:162` — `const columns = Object.keys(data).filter(...)`
- update: `packages/core/src/collections/operations/update.ts:204` — `const keys = Object.keys(data).filter(...)`

Those keys (`publishedAt`, `featuredImageId`, the hasMany `authors`) never match the snake_case `posts` columns (`published_at`, `featured_image_id`, single `author_id`), and the `posts` table has no `_json` overflow column, so dashboard post writes silently missed real columns. Posts had **no** bridge handler (`apps/admin/src/lib/db/typedCollectionStorage.ts` shipped `pages`/`tenants`/`users` only), so every write fell to that path.

### The fix (`apps/admin/src/lib/db/typedCollectionStorage.ts`)

New `posts` handler with `findByID`/`find`/`create`/`update`/`delete`, using the same `undefined` = not-handled contract as the pages handler:

- **Reads** map the canonical row to the collection doc; `_status` mirrors the `status` column so `authenticatedOrPublished`, the drafts UI, and `revalidatePost` keep their contract; the single `author_id` is reflected back as a one-element `authors` array for `populateAuthors`. The where-parser flattens the access-merged `{and:[...]}` shape and translates `_status` to the `status` column; shapes it can't express (`or`, non-equals) signal not-handled and the engine falls through.
- **Writes** map each collection field to its Drizzle column: `authors[]` to `author_id` (first author; the remainder cannot be held by this table, a pre-existing data-model limitation documented in the handler), `categories` normalized to `string[]`, `_status`/`status` to the `status` column plus the `published` boolean, `publishedAt` coerced to a `Date`. Handled-but-not-found and missing title/slug throw (never a silent fall-through to the trap).

### Parity with the pages bridge

Same file, same handler shape, same not-handled/throw contract, same test shape. The `posts` handler mirrors `pages` field-for-field on the seam, so the future edit-session layer can treat pages and posts uniformly. No third bridge shape introduced.

## Verification

- `pnpm install --frozen-lockfile --prefer-offline` — clean.
- `turbo build --filter=@revealui/core --filter=@revealui/db` — green; `admin` app **compiled successfully** (3.7min). The build's integrated TypeScript phase was killed by the sandbox (OOM); standalone `tsc --noEmit` for `admin` passes **clean** (verified separately).
- Typed-storage suite **23/23** (8 new posts tests: findByID author/_status mapping, access-merged and-where, not-handled `or`, camelCase create round-trip, missing-slug throw, update `_status` to status/published + authors to author_id, not-found throw, soft-delete). Two prior tests that used `posts` as an "unsupported" slug were repointed to `media` (posts is now handled).
- Affected `db` + `collections` + `blocks` suites **250/250**. Biome clean.

### Prove-red

Stashing the handler and running the camelCase create round-trip goes **red**: `create` on `posts` returns `undefined` (unhandled, so the dynamic-SQL trap), so `calls.values[0]` is `undefined` and the assertion fails. With the handler, the Drizzle insert receives `authorId: 'user_7'` (from `authors: ['user_7','user_8']`), `featuredImageId`, normalized `categories`, `status`/`published`, and a `Date` `publishedAt`.

### DIFF
--- a/apps/admin/src/lib/db/__tests__/typedCollectionStorage.test.ts
+++ b/apps/admin/src/lib/db/__tests__/typedCollectionStorage.test.ts
@@ -92,7 +92,7 @@ describe('typedCollectionStorage', () => {
     const storage = createTypedCollectionStorage();
     const result = await storage?.find?.(
       {
-        slug: 'posts',
+        slug: 'media',
         fields: [],
       },
       {},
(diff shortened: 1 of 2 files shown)


RevealUIStudio:

The audit proves the trap was live, not theoretical: posts had no bridge handler, so dashboard post writes fell through to the dynamic-SQL path whose column identifiers come straight from camelCase field keys — publishedAt/featuredImageId/authors never matched published_at/featured_image_id/author_id, and with no _json overflow column on the posts table those writes silently dropped real data. The fix extends the #1724 seam field-for-field rather than inventing a third bridge shape, which is exactly the parity the edit-session layer needs.

Judgment calls both look right: authors[] to the single author_id FK is a pre-existing schema limitation, correctly documented in the handler rather than silently papered over (a multi-author schema change is separate work); and the where-flattening keeps the not-handled signaling contract so unsupported shapes fall back loudly instead of mistranslating. Prove-red recorded (unhandled create returns undefined against base). The OOM-killed integrated TS phase is an environment limit; standalone tsc is clean and CI is authoritative.

PR #1825, opened 2026-07-11
feat(core): honor draft param, write _status, add snapshot ops

RevealUIStudio:
## Summary

Makes drafts work end-to-end through the `@revealui/core` collection operations layer, per foundation item F4 of the internal visual-edit-sessions spec (the edit-path keystone). Extends the existing operations — no parallel path. References the internal CMS content-surfaces audit (finding B5: REST accepts `draft` but `find()` hard-codes `draft:false`; nothing writes `_status`; snapshots unimplemented).

Three changes:
1. `find()`/`findByID()` honor the REST `draft` query param.
2. `update()` persists `_status` transitions (they were silently dropped).
3. New `createSnapshot()`/`restoreSnapshot()` operations modeled on the revisions pattern.

## ⚠️ Awaiting security verdict — DO NOT MERGE

This branch touches `collections/operations`, a security-sensitive surface (the draft-leak boundary). Per the review-before-merge convention it requires a **recorded coordinator verdict from a non-author** before merge. This PR is proposal-shaped only; the author does not self-clear. Do not merge and do not apply any `sec-review:*` label without an independent review.

## Access-control statement for drafts (the invariant)

`draft=true` **never** widens what a caller may read. For every read:

- `access.read` is evaluated exactly as before and AND-merged into the query (unchanged path).
- Independently, a drafts-enabled collection (`versions.drafts`) gets a published-only filter `{_status: {equals: 'published'}}` **by default**. `draft=true` removes **only** this default filter.
- Because the two filters are independent and AND-merged, a caller scoped to published content by `access.read` (an unauthenticated / non-admin reader, via `authenticatedOrPublished`) stays published-only **even when it passes `draft=true`**. Draft content is returned only to a caller whose `access.read` returns `true` (or a clause that admits the row) **and** who asked for `draft=true`.

## File:line map

**Ignored before / fixed:**
- `packages/core/src/api/rest.ts:49,93-97` — parses `draft` (already correct).
- `packages/core/src/collections/operations/find.ts:56` — `draft` was never destructured; `draft:false` hard-coded into `afterRead` (was lines 113,250). Now destructured, published-only filter layered into the AND-merge, real `draft` passed to `afterRead`.
- `packages/core/src/collections/operations/findById.ts` — routed through `find()`'s AND-merge whenever an access clause **or** the published-only filter applies (previously only for a row-level access clause), so an admin without draft mode and a public caller with `draft=true` both get published-only; real `draft` passed to `afterRead`.
- `packages/core/src/collections/operations/update.ts:209-218` — added `_status` to the dynamic-SQL column allowlist (it was dropped, so `_status` never persisted). Gated by the pre-existing `access.update` enforcement.

**New:**
- `packages/core/src/collections/operations/drafts.ts` — `collectionHasDrafts()` + `publishedOnlyReadFilter()` (shared by find/findById; the one place the published-only rule lives).
- `packages/core/src/collections/operations/snapshot.ts` — `createSnapshot()`/`restoreSnapshot()`, gated by `access.update`, delegating to a new optional `collectionStorage` seam. Persistence is owned by the typed store; this PR adds no dynamic revisions table (the typed adapter lands with the typed-storage write bridge, a separate foundation item).
- `packages/core/src/types/runtime.ts` — optional `snapshot`/`restore` on `CollectionStorageAdapter`; `draft` threaded onto the `findByID` option types.
- `packages/core/src/collections/CollectionOperations.ts`, `instance/RevealUIInstance.ts`, `instance/methods/findById.ts` — thread `draft` through the findByID chain; expose `snapshot`/`restore` on the collection surface.

## Tests

New suite `packages/core/src/collections/operations/__tests__/drafts.test.ts` (15 tests) covering draft read semantics on both `find()` and `findByID()`, `_status` writes + access-gated denial, and snapshot/restore delegation + access gating. Two tests are the explicit draft-leak regression guards (an unauthenticated `draft=true` read returns published content only, on both operations).

- `@revealui/core` full suite: **2787 passed** (111 files).
- Operations suite incl. new tests: **156 passed**.
- Biome clean on all changed files.
- `turbo build --filter=@revealui/core --filter=server`: success.

## Prove-red

Reverted the three behavioral files (`find.ts`, `findById.ts`, `update.ts`) to `origin/test`, kept the new tests, and re-ran — **3 tests went red**, matching the three new behaviors:

- `find() … no draft param + admin → published only` (old code returned drafts too)
- `findByID() … draft=false + admin → draft doc returns null` (old code returned the draft)
- `update() persists an _status transition` (old code dropped `_status`)

The two draft-leak invariant guards stayed **green** with the fix reverted — as expected: the boundary is enforced by the `access.read` AND-merge, which `draft=true` cannot bypass. The regression tests lock that a future change cannot make `draft=true` bypass access.

## Checker verdict

```
$ node security-pr-review-check.js --diff origin/test
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test).
   When you open the PR, it needs a recorded coordinator verdict before merge.
```

HOLD is expected. This PR waits for a non-author security review.

### DIFF
--- a/packages/core/src/collections/CollectionOperations.ts
+++ b/packages/core/src/collections/CollectionOperations.ts
@@ -28,6 +28,12 @@ import { deleteDocument } from './operations/delete.js';
 import { deleteMany } from './operations/deleteMany.js';
 import { find } from './operations/find.js';
 import { findByID } from './operations/findById.js';
+import {
+  createSnapshot,
+  type RestoreOptions,
+  restoreSnapshot,
+  type SnapshotOptions,
+} from './operations/snapshot.js';
 import { update } from './operations/update.js';
 import { updateMany } from './operations/updateMany.js';
 
@@ -55,6 +61,8 @@ export class RevealUICollection {
     depth?: number;
     req?: RevealRequest;
     populate?: PopulateType;
+    overrideAccess?: boolean;
+    draft?: boolean;
   }): Promise<RevealDocument | null> {
     return findByID(this.config, this.db, options);
   }
@@ -81,5 +89,13 @@ export class RevealUICollection {
 
   async deleteMany(options: BatchDeleteOptions): Promise<BatchResult<RevealDocument>> {
     return deleteMany(this.config, this.db, options);
+  }
+
+  async snapshot(options: SnapshotOptions): Promise<RevealDocument> {
+    return createSnapshot(this.config, this.db, options);
+  }
+
+  async restore(options: RestoreOptions): Promise<RevealDocument> {
+    return restoreSnapshot(this.config, this.db, options);
   }
 }
--- a/packages/core/src/collections/operations/__tests__/drafts.test.ts
+++ b/packages/core/src/collections/operations/__tests__/drafts.test.ts
@@ -1,0 +1,345 @@
+/**
+ * Draft read/write + snapshot enforcement tests
+ *
+ * Proves the draft boundary that keeps unpublished content from leaking:
+ *   - find()/findByID() honor the `draft` param, but ONLY relax the
+ *     published-only default — never the access.read clause.
+ *   - THE CRITICAL INVARIANT: an unauthenticated read with `draft=true` still
+ *     returns published content only.
+ *   - update() persists `_status` transitions, gated by access.update.
+ *   - createSnapshot()/restoreSnapshot() are gated by access.update and
+ *     delegate to the typed storage seam.
+ */
+
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import type {
+  DatabaseResult,
+  RevealCollectionConfig,
+  RevealRequest,
+} from '../../../types/index.js';
+import { find } from '../find.js';
+import { findByID } from '../findById.js';
+import { createSnapshot, restoreSnapshot } from '../snapshot.js';
+import { update } from '../update.js';
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function mockRequest(overrides: Partial<RevealRequest> = {}): RevealRequest {
+  return {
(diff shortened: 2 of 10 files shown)


RevealUIStudio:
**Security review verdict (Fable, non-author, adversarial read): APPROVE — owner records the gate clearance and merges.**

I re-derived the draft-leak boundary from the diff rather than the PR description:

- **The invariant holds structurally.** find() AND-merges three independent restrictions (caller where, access.read clause, published-only filter), and draft=true omits only the third. There is no code path where draft=true removes or weakens the access clause, so an unauthenticated or authenticatedOrPublished-scoped reader stays published-only regardless of what they request. The two invariant tests pin exactly this, and the prove-red result is the strongest evidence: with the behavioral fix reverted the three feature tests go red while the invariant guards STAY GREEN, meaning the boundary is enforced by the access merge itself, not by the new code — which is precisely the design intent.
- **findById now routes through find()'s proven merge** whenever either a row-level access clause or the status filter applies, closing the by-id side channel that previously only considered access clauses.
- **collectionHasDrafts correctly distinguishes versions-without-drafts** (no _status lifecycle → no filtering), keeping non-draft collections backward compatible.
- **Two nuances I verified and accept, recorded for the future:** (1) overrideAccess:true skips the published-only filter along with access control — unchanged trust model for internal callers, and the proxy already strips the param from external requests; (2) _status writes are gated by access.update, which is the existing permission granularity (no separate publish permission exists in the model) — if a publish-role split ever lands, this is the line to revisit.
- The snapshot seam is access.update-gated, delegated to storage, and deliberately does not build core versions machinery the session layer will supersede — the right scope call.

2787 core tests green, checker HOLD is the expected state pending the owner's recorded clearance. This comment is my verdict; I apply no label.

PR #1843, opened 2026-07-11
feat(auth): record successful and failed logins in the audit trail

RevealUIStudio:
## Why

A successful sign-in emitted **no audit event** anywhere in the fleet. For a receipts-first product, human authentication leaving no receipt is a coherence gap: the audit trail could not answer "who signed in when" from its own records.

## Call-site audit

- `auditLoginSuccess` (`packages/auth/src/server/audit-bridge.ts`) had **zero call sites** — only re-exports (`packages/auth/src/server/index.ts`, `packages/engines/src/users/index.ts`). Dead fleet-wide.
- `auditLoginFailure` — same: **zero call sites**. Dead too.
- `signIn` (`packages/auth/src/server/auth.ts`) → `rotateSession` minted a session on success but recorded nothing to the audit trail.
- The email-verification route (`apps/admin/src/app/api/auth/verify-email/route.ts`) minted via `createSession` and likewise recorded nothing.

## Both success emits wired (same PR, so they never diverge)

- **Sign-in:** `signIn()` emits `auditLoginSuccess(userId, ip, userAgent)` exactly once, right after a fresh session is minted. MFA-pending returns earlier (no session yet), so no premature receipt is written for a half-completed login.
- **Verify-email first login:** the route emits `auditLoginSuccess(userId, ip, userAgent)` exactly once after minting its session.

Both reuse the existing `auditLoginSuccess` primitive and its event shape (`type: auth.login`, `actor.id`, timestamp stamped by `AuditSystem.log`). No new audit pipeline.

## Failure-side finding

**Failures were also unrecorded in the audit trail.** `recordFailedAttempt` (`packages/auth/src/server/brute-force.ts`) only bumps the brute-force **lockout counter** (`brute_force:<email>` storage key); it never writes to the `AuditSystem`. So I wired `auditLoginFailure` on the credential-failure branch (consolidated into one helper alongside the existing `recordFailedAttempt` call). A failed sign-in now leaves a receipt too.

> Out of scope, noted for a follow-up: account-lockout (`auditAccountLocked`) and MFA-completion login success are still unwired — separate branches from the two paths this PR covers.

## Parity ruling: adopt `rotateSession` on verify-email

**Decision: switched verify-email from `createSession` to `rotateSession`** (revoke-then-issue), matching sign-in.

**Trace (why nothing depends on the signup-time session surviving):** for a user who goes through email verification, the signup-time session row is created at sign-up but its token is only ever handed to the client as a cookie **once the email is verified** — the sign-up route sets the `revealui-session` cookie behind an `isVerified` guard. So for an unverified-then-verifying user, that row is an **undelivered orphan** no device holds. Rotating on verification reclaims it and leaves the user with exactly one live session. (First users are auto-verified at signup and never traverse this route, so they are unaffected.)

## Non-blocking

The audit bridge now **swallows and logs** a write failure, so an audit-write error can never fail the operation it records. This is the single seam that makes every emit best-effort.

## Tests + prove-red

New regression tests:
- (a) successful sign-in lands exactly one success event with the actor id
- (b) successful verify-login lands exactly one success event with the actor id
- (c) invalid-credentials sign-in lands exactly one failure event (finding: failures were unrecorded, now wired)
- (d) an audit-write failure does not throw (bridge best-effort), so the login is unaffected
- (e) verify-email mints via `rotateSession` (revoke-then-issue), pinning the parity decision

**Prove-red:** stashing the wiring turns the new tests red (auth: 5 new tests fail; verify-email: the parity + emit tests fail); restoring the wiring returns them green.

Results:
- `pnpm --filter @revealui/auth exec vitest run` -> **528 passed, 19 skipped** (full suite green)
- admin auth route tests (`verify-email-route`, `sign-in-route`) -> **29 passed**
- `pnpm exec turbo build --filter=admin --filter=server` -> **21 tasks successful**

## Security checker verdict

`security-pr-review-check.js --diff origin/test` -> **SECURITY-SENSITIVE (HOLD)**, touching `packages/auth/` + `api/auth/`. This is expected and correct for an auth-surface change.

## Awaiting a non-author security verdict before merge

This PR touches the authentication surface. Per the review-before-merge convention it **must carry a recorded verdict from a non-author reviewer before it merges**. Do not self-merge; do not apply any `*approved` label as the author.

### DIFF
--- a/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
+++ b/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
@@ -14,13 +14,15 @@ import { createHash } from 'node:crypto';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mockCheckRateLimit = vi.fn();
-const mockCreateSession = vi.fn();
+const mockRotateSession = vi.fn();
+const mockAuditLoginSuccess = vi.fn();
 const mockGetUserByVerificationToken = vi.fn();
 const mockUpdateUser = vi.fn();
 
 vi.mock('@revealui/auth/server', () => ({
   checkRateLimit: (...args: unknown[]) => mockCheckRateLimit(...args),
-  createSession: (...args: unknown[]) => mockCreateSession(...args),
+  rotateSession: (...args: unknown[]) => mockRotateSession(...args),
+  auditLoginSuccess: (...args: unknown[]) => mockAuditLoginSuccess(...args),
 }));
 
 vi.mock('@revealui/utils/logger', () => ({
@@ -113,6 +115,7 @@ describe('GET /api/auth/verify-email', () => {
       remaining: 9,
       resetAt: Date.now() + 60000,
     });
+    mockAuditLoginSuccess.mockResolvedValue(undefined);
   });
 
   async function loadRoute() {
@@ -124,7 +127,7 @@ describe('GET /api/auth/verify-email', () => {
     const GET = await loadRoute();
     const res = await GET(makeRequest());
     expect((res as MockRes).url).toContain('error=missing_token');
-    expect(mockCreateSession).not.toHaveBeenCalled();
+    expect(mockRotateSession).not.toHaveBeenCalled();
   });
 
   it('redirects with too_many_attempts when rate limited', async () => {
@@ -137,7 +140,7 @@ describe('GET /api/auth/verify-email', () => {
     const GET = await loadRoute();
     const res = await GET(makeRequest('some-token'));
     expect((res as MockRes).url).toContain('error=too_many_attempts');
-    expect(mockCreateSession).not.toHaveBeenCalled();
+    expect(mockRotateSession).not.toHaveBeenCalled();
   });
 
   it('fails closed  -  redirects with error when rate limit check throws', async () => {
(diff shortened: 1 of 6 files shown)


RevealUIStudio:
**Security review verdict (Fable, non-author, adversarial read): APPROVE — owner records the gate clearance and merges.**

Re-derived from the diff:

- **Both success paths wired in one PR, as the tracker required:** sign-in emits after rotateSession, verify-email emits as the first-login receipt after its (now-rotating) session issuance — the two paths cannot drift because they landed together, each pinned by an exactly-one test.
- **The failure-side finding upgrades the fix:** auditLoginFailure was equally dead, and the consolidation of all four invalid-credential branches (unknown user, OAuth-only, bcrypt error, wrong password) through one helper is the right shape — uniform reason, email/ip/ua as the attempted-actor tuple, and the password never touches the event. Failed logins finally leave receipts.
- **The parity ruling is proven, not asserted:** the signup-time session token only reaches a client cookie after verification (sign-up route gates on isVerified), so the row rotateSession reclaims on verify was an undelivered orphan no device holds. Switching to revoke-then-issue gives both login paths identical session hygiene with zero user-visible change; auto-verified first users never traverse the path.
- **The non-blocking seam is correct where it matters:** one try/catch in the bridge makes every emit best-effort, logging the write failure with the event type — an audit outage can never lock users out, and the test pins both the swallow and the log.
- Prove-red recorded (5 tests red without the wiring); auth suite 528 green; builds clean; the checker HOLD is the expected pre-clearance state.

**Follow-up scope, correctly excluded:** auditAccountLocked and the MFA-completion success path remain unwired — flag them on the tracking record so they ride a small successor rather than silently widening this PR.

For a receipts-first product, this closes the loudest coherence gap in the story: human authentication now leaves the same evidence trail agent actions do. This comment is my verdict; I apply no label.

PR #1830, opened 2026-07-11
feat(server): disarmed tier-aware onboarding lifecycle emails

RevealUIStudio:
## What this is

The tier-aware onboarding lifecycle email sequence: a day-0 welcome, a day-1 check-in, and a day-7 outcome email, plus scheduling and a send ledger. **Sending is disarmed by default** and this PR does not turn it on. It is primarily a copy deliverable riding minimal infrastructure that already existed.

## Infra reuse (extend, never rebuild)

Every moving part is an existing primitive I extended, not a new one:

- **Transport:** sends through the existing `sendEmail` in `apps/server/src/lib/email.ts` (the Gmail REST path via `@revealui/services`). No new mailer.
- **Templates:** I extracted the shared HTML helpers (`emailShell`, `ctaButton`, `supportFooter`, `tierLabel`, `adminUrl`, `billingUrl`) out of `apps/server/src/lib/webhook-emails.ts` into a new `apps/server/src/lib/email-templates.ts`, and pointed the existing Stripe webhook emails at it. Both surfaces now render from one implementation instead of two copies. New senders live in `apps/server/src/lib/lifecycle-emails.ts`.
- **Cron:** one new job registered in the existing daily dispatcher (`apps/server/src/routes/cron/dispatch.ts`, `POST /api/cron/dispatch`, 06:00 UTC). The job file `apps/server/src/routes/cron/lifecycle-emails.ts` copies the structure and X-Cron-Secret validation of `sweep-grace-periods.ts`.
- **Idempotency:** the send ledger reuses the `usage_meters.idempotency_key` pattern (unique text key + `insert().onConflictDoNothing()`).
- **Signals:** tier and agent-action detection reuse the correlated-subquery shape from the activation analytics endpoint (`routes/analytics.ts`) — account tier from `accountEntitlements.tier`, first/weekly agent action from `usage_meters WHERE source='agent'`.

## Disarm mechanism

The transport is called only when `LIFECYCLE_EMAILS_ENABLED` is exactly `'true'`. While disarmed, the job records would-send decisions (dry-run ledger rows plus log lines) and never touches the mailer. The dry-run claim and the real-send claim use **different idempotency keys** (`lifecycle:dry_run:...` vs `lifecycle:sent:...`), so a disarmed run never consumes a real-send slot: once armed, each user still receives their email exactly once. The flag is declared in the config schema (`packages/config/src/schema.ts`) and documented in `.env.template` + `.env.production.example`.

[...]

PR #1853, opened 2026-07-11
fix(admin): guard mock.calls index in verify-email test (TS2532)

RevealUIStudio:
## Problem

`test` HEAD carries a strict-index type error in `apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts:251`:

```
error TS2532: Object is possibly 'undefined'.
```

`mockAuditLoginSuccess.mock.calls[0][0]` indexes `calls[0]` which is possibly undefined under `noUncheckedIndexedAccess`. The PR-only "Typecheck affected packages (PRs — hard fail)" job typechecks the PR-merge result, so **every** open PR merging into test inherits this failure (surfaced on #1830). The push-triggered gate on test does not run the same hard-fail affected-typecheck, which is how it slipped onto test.

## Fix

Guard the first-call access with optional chaining: `mock.calls[0]?.[0]`. The preceding `toHaveBeenCalledTimes(1)` assertion guarantees the call exists at runtime, so behavior is unchanged.

### DIFF
--- a/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
+++ b/apps/admin/src/app/api/auth/__tests__/verify-email-route.test.ts
@@ -248,7 +248,7 @@ describe('GET /api/auth/verify-email', () => {
     await GET(makeRequest('valid-token'));
 
     expect(mockAuditLoginSuccess).toHaveBeenCalledTimes(1);
-    expect(mockAuditLoginSuccess.mock.calls[0][0]).toBe('u1');
+    expect(mockAuditLoginSuccess.mock.calls[0]?.[0]).toBe('u1');
   });
 
   it('does not emit a login-success audit event when no session is created', async () => {


RevealUIStudio:
**Merge-train blocker — owner sec-review needed.**

This PR is a one-line, test-only strict-index guard (`mock.calls[0][0]` → `mock.calls[0]?.[0]`) that fixes a real TS2532 error already sitting on `test`. The PR-only hard-fail Typecheck job typechecks the PR-merge result, so **every** PR merging into test currently inherits this failure (it surfaced on #1830).

All checks are green **except the Security review gate**, which holds only because the touched file lives under `apps/admin/src/app/api/auth/__tests__/` (path-based sensitivity). There is no behavioral/security change — the preceding `toHaveBeenCalledTimes(1)` assertion guarantees the guarded call exists at runtime.

Blocking #1830 behind it. Needs an approving review or the `sec-review:approved` label to merge (author cannot self-clear).

PR #1858, opened 2026-07-11
feat(knowledge-graph): fleet knowledge graph core (GAP-340 P1)

RevealUIStudio:
**Part of GAP-340** (fleet knowledge graph). Later phases (P2 MCP/skill, P3 LLM extraction, P4 sync/explorer, P5 daemon replica) remain — this is P1 only.

Design spec: `.jv/docs/lanes/fleet-knowledge-graph/design-spec.md` (decided; not redesigned here).

## Summary (keyed to spec sections)

- **§4 data model** — `packages/db/src/schema/knowledge-graph.ts`: `kg_episodes`, `kg_nodes`, `kg_edges`, `kg_edge_episodes`, `kg_node_aliases`, `kg_outbox`. Content-addressed ids, bi-temporal columns (`valid_at`/`invalid_at`, `created_at`/`expired_at`), denormalized `repo` partition column, `site_id`. Migration `0021_knowledge_graph.sql` (hand-written per the `0002` precedent; `_custom.json` entry added): HNSW `vector_cosine_ops`, GIN over `tsvector` generated columns, partial current-edge indexes (`WHERE invalid_at IS NULL AND expired_at IS NULL`). **Ships unapplied** — validated only against PGlite. The duplicated `vector` customType from `rag.ts`/`agents.ts` is extracted to shared `_vector-type.ts` and consumed by all three sites.
- **§5 ontology** — `@revealui/knowledge-graph/ontology`: Zod schemas for the prescribed node kinds + edge relations; learned entities fall back to `concept`/`relates-to`; JSON-Schema conversion for future structured-output extraction.
- **§4/§8 ids + convergent writes** — deterministic SHA-256 UUIDs (`deriveNodeId`/`deriveEdgeId`/`deriveEpisodeId`, extending the `node-ids` pattern). The write API applies the class-1/2 merge rules: `ON CONFLICT DO NOTHING` for G-Set rows, `LEAST`/`GREATEST` for monotonic columns. Ops are recorded to `kg_outbox`.
- **§6 ingestion** — `applyScan` (deterministic-scan path with re-scan diff invalidation: absent edges get `invalid_at = scan.referenceTime`, never deleted; presence keyed on the logical `(source, target, relation)` triple so unchanged re-scans are no-ops) and additive `ingestEpisode`. Entity resolution ladder steps 1-2 (natural key + alias). Every edge is forced to trace to ≥1 episode at the ingest API. Embeddings are **injected** and best-effort — they degrade to NULL + deferred backfill when Ollama is down.
- **§6 Tier-1 extractors** — workspace (pnpm-workspace + package.json), ts-project (TypeScript compiler API), db-schema (compiler API over `pgTable`), git (git plumbing), docs-frontmatter (YAML), routes (compiler API). **No regex over source.**
- **§7 retrieval** — `kgSearch`: pgvector cosine (skipped gracefully without a query embedding / pgvector) + `websearch_to_tsquery`/`ts_rank` + recursive-CTE BFS, RRF-fused (k=60), reranked by node-distance + episode-mentions, point-in-time predicates. Plus `kgNeighbors`, `kgPath`, `kgAtTime`.
- **Surface** — `revkg` CLI: `scan`/`search`/`node`/`neighbors`/`at`.

## Constraints honored

- No `@anthropic-ai/sdk` / `openai` in `packages/**` (boundary validator Check 5 green). Embeddings only via `@revealui/ai` `generateEmbedding` (nomic-embed-text/768), wired at the CLI.
- No regex over source (compiler API / YAML / git plumbing only).
- Edges never DELETEd; supersession sets `invalid_at`/`expired_at`.
- secret-path nodes carry the path only, never a value.
- TypeScript strict, ESM, Biome, Vitest. Boundary + migration-journal validators pass.

## P1 checklist

- [x] Extract shared `vector` customType from rag.ts/agents.ts + new schema
- [x] `knowledge-graph.ts` schema + migration (HNSW, GIN tsvector, partial current-edge, repo partition, siteId)
- [x] Content-addressed ids + convergent write API (G-Set / min-wins merge)
- [x] `@revealui/knowledge-graph` package (MIT, 0.1.0): ontology, ids, ingest, search
- [x] Tier-1 extractors (workspace, ts-project, db-schema, git, docs-frontmatter, routes)
- [x] Re-scan diff invalidation
- [x] Hybrid search (vector + FTS + BFS, RRF, rerankers, point-in-time)
- [x] `revkg` CLI (scan/search/node/neighbors/at)
- [x] Vitest incl. PGlite temporal-invalidation + convergence tests
- [ ] First full-fleet scan — deferred: `revkg scan --fleet` is implemented and runs against Neon, but running it here would write to prod (out of scope for this PR, which must not touch prod). Run it after merge with `POSTGRES_URL` set.

## Tests

30 tests pass (7 files). PGlite-backed suites exercise the real SQL (FTS, recursive-CTE BFS, temporal predicates, class-1/2 merge) via a portable DDL variant (`real[]` embeddings, no HNSW) generated from the same `ddl.ts` source that renders the migration; a drift guard asserts the shipped migration matches it.

Prove-red performed (all-new code, so by breaking the load-bearing logic rather than checking out base):
- Disabling re-scan invalidation → 3 temporal tests go red.
- Replacing `first_seen_at = LEAST(...)` with last-writer → the cross-batch convergence test goes red (the interleaving reorders monotonic upserts across scan batches, so it only converges under a genuine min-wins register).

## Escalations for Fable

1. **MIT ↔ FSL licensing boundary.** `@revealui/knowledge-graph` is declared MIT, but embeddings come from `@revealui/ai` (FSL-1.1-MIT). Conservative resolution: the core library imports no LLM SDK and takes an injected `Embedder`; `@revealui/ai` is an `optionalDependency` loaded via a **non-literal dynamic import in the CLI only**, so the OSS→FSL static-import boundary check passes. Confirm this is the intended posture, or whether the package should be FSL.
2. **Scope keying for re-scan invalidation.** Diff invalidation is scoped by `(repo, attributes.extractor)` rather than by relation set, because `workspace` and `ts-project` both emit `contains` and relation-scoping would let one extractor invalidate the other's edges. Minor deviation from the spec's implied relation scoping; flagging for review.
3. **`attributes` LWW is partial in P1.** Attribute merge is jsonb right-wins concat and `attributes_clock` is stored but not yet enforced as `(timestamp, siteId)` LWW. Safe because Tier-1 extractors are idempotent; full LWW deferred to P3. Confirm acceptable.
4. **git extractor is additive.** Commit `implements` edges (distinct historical `valid_at`) route through `ingestEpisode`, not `applyScan`, so reconcile does not churn them.


RevealUIStudio:
## Fable adversarial review (GAP-340 P1, spec conformance pass)

Verdict: REQUEST CHANGES. Two findings must land before merge; both are cheap now and expensive after first data exists. Everything else conforms to the design spec, and the convergence design (op-level replication on deterministic ids) is correctly implemented.

### Must fix before merge

1. Literal NUL bytes in source (`ids.ts` line 15, `engine.ts` scan-key template literals). Git classifies both files as binary, which permanently breaks diff, blame, and review for them. Worse, the NUL is the join separator inside `deriveEdgeId` and `deriveEpisodeId`, so any tool that strips or mangles that byte silently re-keys every edge and episode in the graph. Replace with the escape sequence \u001F (unit separator, written as an escape in source, never as a raw byte) in all four sites. This changes derived ids, which is exactly why it must happen now: the migration is unapplied and zero rows exist. After merge it becomes a full re-key.

2. `applyScan` and `applyOps` run without a transaction. The spec (section 6) requires per-episode transactional application. A crash mid-apply leaves a partially applied scan (episode present, edges half-written, stale edges not invalidated) and the current view stays wrong until a re-run. Wrap the op batch in BEGIN/COMMIT via the executor (add a transactional helper to `KgExecutor` or document and implement caller-side wrapping in the CLI and any future API route).

### Accepted with notes

3. Fact-text drift: a changed `fact` string with an unchanged (source, target, relation) triple is never rewritten because the seen-key skip preserves the existing row. Acceptable for templated Tier-1 facts; add a code comment stating the invariant.
4. Node `deleted_at` is never set in P1 (tombstoning unimplemented). When implemented, derive liveness at read time as `deleted_at > last_confirmed_at` instead of treating the min-wins tombstone as authoritative, otherwise a deleted-then-recreated file stays dead forever. Spec amendment will follow.

### Escalation rulings

- E1 (MIT core + optional dynamic import of FSL `@revealui/ai`): APPROVED. Matches the paywall/core split precedent and the boundary validator passes.
- E2 (re-scan invalidation scoped by repo + extractor): APPROVED, strictly better than relation-set scoping. Spec will be amended to match.
- E3 (partial attribute LWW in P1): ACCEPTED for P1 only. Note that jsonb `||` merge is order-dependent for conflicting keys; Tier-1 determinism makes it safe today. Full per-key LWW with the clock column is a hard P3 requirement.
- E4 (git extractor is additive, never reconciled): APPROVED, matches the episodic intent of the spec.

Review basis: fleet security checker reports no security-sensitive files (exit 0, no coordinator gate). Merge remains an owner action.

RevealUIStudio:
## Fable review follow-up: findings resolved

Both must-fix findings from the review above are resolved in ab00599b9 and independently verified:

1. NUL bytes: both files now report as UTF-8 text with zero NUL bytes in the committed blobs; the separator is an exported SEP constant whose value is the unit separator written as a 6-character backslash escape in source. Future diffs against these files are line diffs, not binary markers.
2. Transactions: KgExecutor gained a transaction() method (BEGIN/COMMIT/ROLLBACK on a single connection; makePoolExecutor checks out a dedicated pooled connection), and both applyScan and ingestEpisode wrap their op batch in it, with embedding backfill outside. The new atomicity test was proven red without the fix.

33/33 package tests pass, typecheck and Biome clean, full pre-push gate green. The design-spec amendments from the escalation rulings (scan scope keyed on repo + extractor, tombstone liveness as a read-time predicate) are committed to the coordination repo.

Verdict: no remaining blockers from this review. Merge remains an owner action (agent-authored PR); CodeQL was still pending at review time and should be green before merge.

PR #1861, opened 2026-07-12
feat(mcp): agent surfaces for the fleet knowledge graph

RevealUIStudio:
## Summary

GAP-349 P2 deliverable A: MCP tools in `@revealui/mcp` exposing the fleet
knowledge graph (`@revealui/knowledge-graph`, P1 shipped in #1858) to agents.

New factory `packages/mcp/src/servers/factories/knowledge-graph.ts`
(`createKnowledgeGraphServer`, exported at `@revealui/mcp` root and via a new
`./kg-server` subpath), following the exact `createXServer` factory pattern
established by `revealui-content.ts` / `docs.ts` — Tool[] + Zod arg schemas
+ `CallToolRequestSchema` switch handler.

Seven tools, keyed to design spec sections:
- `kg_search` — hybrid retrieval (vector + FTS + BFS, RRF-fused) — spec §7
- `kg_get_node`, `kg_neighbors`, `kg_path`, `kg_at_time` — read/traversal wrappers over the P1 search API
- `kg_add_episode` — **the only write tool**. Always calls the additive
  `ingestEpisode` (never `applyScan`, the deterministic-rescan path reserved
  for `revkg scan`). Zod-validates `episodeType` and every node kind / edge
  relation against the ontology enums; `.strict()` schemas throughout so no
  raw SQL or table names can be smuggled through any tool input.
- `kg_context` — spec §8.4 budgeted context assembly: BFS from an anchor,
  rerank by node-distance + episode-mentions, pack node summaries + edge
  facts (with provenance episode ids) into a text block capped at
  `charBudget` characters (default 16000; chars not tokens, per task scope).

**DB connection pattern (extend, not duplicate):** production resolves a
pooled `KgExecutor` lazily via dynamic `import('@revealui/db/pool')` +
`makePoolExecutor` — the same connection pattern `revkg`'s own CLI already
uses (`packages/knowledge-graph/src/cli/revkg.ts`). Tests inject a
PGlite-backed executor via `CreateKnowledgeGraphServerOptions.executor`.

**Graceful degradation:** read tools work with no pgvector and no
embeddings present — `kgSearch`'s vector channel is skipped whenever no
query embedding is supplied. Query-embedding generation is lazy + optional
(`@revealui/ai/embeddings` via a non-literal dynamic import specifier, same
technique `revkg` uses to avoid a hard type/import edge to the optional Pro
package) and wrapped in try/catch per-call, so an Ollama-down
`generateEmbedding()` call falls back to FTS + BFS only, never a hard tool
failure.

## Escalations for Fable

- **MCP SDK `CallToolResult` union assignability.** The SDK's
  `CallToolRequestSchema` handler return type is a union that includes a
  `task`-shaped member (async/background tool results). Returning a value
  through a named helper function (rather than an inline fresh object
  literal) loses TypeScript's literal-freshness exemption for the
  index-signature/task-union check, so `textResult`/`errorResult`/
  `parsed.error` (from the shared `validate-tool-args.ts` `McpToolError`
  type) needed `as CallToolResult` / `as unknown as CallToolResult` casts to
  typecheck. Verified safe at runtime (all 27 new tests pass, including
  explicit error-path assertions), but it's a structural-typing wart worth a
  second look — a cleaner fix might be adding an index signature to the
  shared `McpToolError` interface itself so every factory using
  `validateToolArgs` benefits, rather than casting at each of my 7 call
  sites. Left as-is (additive-only local change) rather than touching a
  shared file used by every other MCP factory in this PR.
- **`EpisodeType` duplication.** `@revealui/knowledge-graph`'s `EpisodeType`
  union (`types.ts`) isn't exported as a runtime array by P1, so the Zod
  enum for `kg_add_episode`'s `episodeType` field duplicates the 7-value
  literal list locally with a comment flagging the lockstep requirement.
  Low-risk (P3 would need to touch this factory anyway on any ontology
  change) but flagging since it's the one place this PR carries knowledge
  that could drift from its source of truth.

## Deferred: bridge exposure (deliverable C)

**Not built.** Audited whether `@revdev/bridge` (separate `revdev` repo/
pnpm workspace) can consume `@revealui/knowledge-graph` today:

- `@revealui/knowledge-graph` is at `0.1.0` in the `revealui` workspace and
  is **not published to npm** (`npm view @revealui/knowledge-graph versions`
  → 404).
- `@revdev/bridge`'s existing `@revealui/*` consumption precedent
  (`packages/daemon/package.json`) is via published semver ranges
  (`"@revealui/contracts": "^1.4.0"`, etc.) — never a path or git dependency
  across the two repos' separate pnpm workspaces.
- Per the task's hard constraint (no path deps across repos, no git deps),
  building C now would require inventing a new, unprecedented cross-repo
  dependency mechanism — out of scope for a P2 agent-surfaces PR.

**Unblock:** publish `@revealui/knowledge-graph` to npm (it already carries
`publishConfig: { access: "public", registry: "https://registry.npmjs.org" }`
and ships via the existing changesets pipeline — `pnpm changeset` →
`pnpm changeset:version` → `release.yml` `workflow_dispatch`, or lands
automatically via `release-canary.yml` on the next push to `test` if a
changeset is present). Once published, `@revdev/bridge`'s change is
mechanical: add `"@revealui/knowledge-graph": "^0.1.0"` to
`packages/bridge/package.json` dependencies and mirror the same seven
tool definitions through its existing MCP tool registration pattern
(`packages/bridge/src/index.ts`), reusing this PR's Zod schemas as the
reference shape.

## Test plan

- [x] `pnpm --filter @revealui/mcp test` — 423 passed, 5 skipped (23 files, incl. 27 new tests)
- [x] `pnpm --filter @revealui/mcp typecheck` — clean
- [x] `pnpm exec biome check` on all changed files — clean
- [x] `pnpm --filter "@revealui/mcp..." build` — clean (full dependency chain)
- [x] `pnpm --filter server typecheck` — clean (consumer of `@revealui/mcp`)
- [x] `pnpm --filter admin typecheck` — clean (consumer of `@revealui/mcp`)
- [x] New tests never touch a real database — PGlite only, built from
      `@revealui/knowledge-graph`'s public export surface
      (`kgDdlStatements` + `makeExecutor`)

### DIFF
--- a/packages/mcp/__tests__/knowledge-graph-factory.test.ts
+++ b/packages/mcp/__tests__/knowledge-graph-factory.test.ts
@@ -1,0 +1,537 @@
+/**
+ * Vitest coverage for the `knowledge-graph` MCP server factory (GAP-349 P2).
+ *
+ * Exercises: tool-schema validation (Zod), kg_add_episode ontology
+ * validation, the read tools (kg_search/kg_get_node/kg_neighbors/kg_path/
+ * kg_at_time) against a PGlite-backed KgExecutor seeded via the P1 additive
+ * ingest path, and kg_context budget enforcement (spec §8.4).
+ *
+ * The PGlite harness mirrors `@revealui/knowledge-graph`'s own
+ * `src/__tests__/test-db.ts`, built only from that package's PUBLIC export
+ * surface (`kgDdlStatements` + `makeExecutor`) since the internal test
+ * helper isn't exported. NEVER touches a real database.
+ */
+
+import { createServer as createHttpServer, type Server as NodeHttpServer } from 'node:http';
+import type { AddressInfo } from 'node:net';
+import { PGlite } from '@electric-sql/pglite';
+import {
+  type EdgeInput,
+  type EpisodeInput,
+  ingestEpisode,
+  type KgExecutor,
+  kgDdlStatements,
+  makeExecutor,
+  type NodeInput,
+} from '@revealui/knowledge-graph';
+import { afterEach, describe, expect, it } from 'vitest';
+import { McpClient } from '../src/client.js';
+import {
+  createKnowledgeGraphServer,
+  KgAddEpisodeArgsSchema,
+  KgAtTimeArgsSchema,
+  KgContextArgsSchema,
+  KgGetNodeArgsSchema,
+  KgNeighborsArgsSchema,
+  KgPathArgsSchema,
+  KgSearchArgsSchema,
+} from '../src/servers/factories/knowledge-graph.js';
+import { createNodeStreamableHttpHandler } from '../src/streamable-http.js';
+
+// ---------------------------------------------------------------------------
+// PGlite test db
+// ---------------------------------------------------------------------------
+
+interface TestDb {
+  exec: KgExecutor;
+  close: () => Promise<void>;
+}
+
+async function createTestDb(): Promise<TestDb> {
+  const pglite = new PGlite();
+  for (const statement of kgDdlStatements({ variant: 'portable' })) {
+    await pglite.exec(statement);
+  }
+  return { exec: makeExecutor(pglite), close: () => pglite.close() };
+}
+
+const REPO = 'revealui';
+const FILE_KEY = 'revealui/packages/ai/src/llm/client.ts';
+const SYMBOL_KEY = `${FILE_KEY}#getClient`;
+const T = new Date('2026-07-01T00:00:00Z');
+
+async function seed(exec: KgExecutor): Promise<void> {
+  const episode: EpisodeInput = {
+    episodeType: 'code-scan',
+    source: 'test',
+    siteId: 'test-site',
+    referenceTime: T,
+    contentRef: { repo: REPO },
+  };
+  const nodes: NodeInput[] = [
+    {
+      kind: 'file',
+      name: 'client.ts',
+      naturalKey: FILE_KEY,
+      repo: REPO,
+      summary: 'the LLM client factory',
+    },
+    {
+      kind: 'symbol',
+      name: 'getClient',
+      naturalKey: SYMBOL_KEY,
+      repo: REPO,
+      summary: 'creates an LLM client',
+    },
+    { kind: 'dependency', name: 'zod', naturalKey: 'npm:zod' },
+  ];
+  const edges: EdgeInput[] = [
+    {
+      source: { kind: 'file', naturalKey: FILE_KEY },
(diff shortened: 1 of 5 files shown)


RevealUIStudio:
## Fable adversarial review (GAP-349 P2, security pass on the MCP surface)

Verdict: APPROVE. No blocking findings. The owner applies `sec-review:approved` to clear the gate (agent-authored PR; the author cannot self-clear).

What was examined: the full factory (819 lines), tool schemas, the write path, and the test suite's coverage claims.

Security posture, confirmed:
- Every SQL statement is parameterized; no string interpolation of user input anywhere.
- All tool inputs are Zod `.strict()` schemas: unknown fields rejected (tested), kinds/relations/episode types enum-bound to the ontology, depths capped (6 BFS, 12 path), limits capped (100), charBudget capped (200k).
- `kg_add_episode` is the only write tool, confined to the additive `ingestEpisode` path (never the rescan reconcile), and every edge it creates carries episode provenance by construction.
- The `@revealui/ai` Pro-package boundary is respected via best-effort dynamic import with graceful FTS+BFS-only degradation; boundary validator passes.

Escalation rulings:
- E1 (CallToolResult casts): ACCEPTED, consistent with every existing factory in this package. Follow-up filed below to fix the shared `validate-tool-args.ts` shape once, for all factories.
- E2 (EPISODE_TYPES literal duplication with lockstep comment): ACCEPTED for P2. P3 must export the runtime array from `@revealui/knowledge-graph` and delete the duplicate.

Accepted with notes (non-blocking, tracked as P3 riders in the lane plan):
1. `kg_search`'s `anchor` takes a raw node id while every other tool takes a natural key; agents will rarely hold raw ids. Resolve natural keys there too.
2. `recordOutbox: true` on hub-side writes accumulates inert `kg_outbox` rows (never pushed, `pushedAt` null) until P5 defines replay. Harmless audit trail today; P5 must define retention.
3. No explicit size cap on `content`/`attributes` payloads; the MCP transport bounds message size in practice, but an explicit cap (for example 1MB content) would make the limit intentional.
4. Error results forward driver error messages to the MCP client; fine for internal tooling, revisit if these tools are ever exposed beyond the fleet.
5. Carried from the P1 incident triage: revkg's header comment states the reverse of the pool's real env precedence, and connection errors should name the target host (never the URL). Both fold into P3.

Companion PR revskills#81 has a separate finding (em dashes in new skill prose, against the 2026-07-09 style rule); fix requested from the implementer before that one merges. It does not block this PR.

Merge remains an owner action:
- gh pr edit 1861 --add-label "sec-review:approved"
- gh pr merge 1861 --squash

PR #1864, opened 2026-07-12
fix(knowledge-graph,db): GAP-349 scan-hardening batch

RevealUIStudio:
Part of GAP-349.

Three defects discovered during the first fleet-wide `revkg scan --fleet` run, fixed in three logically separate commits.

## Fix 1: single-package repo extractor coverage

The `ts-project`/`workspace` extractors only discovered scan targets via pnpm workspace membership, so non-pnpm-monorepo fleet repos produced almost nothing (observed: "agency: 1 nodes, 0 edges" for a whole Vite SPA).

When a repo has no `pnpm-workspace.yaml`, the repo root is now treated as a single implicit package (name from `package.json`, or the directory name), scanned the same way a workspace package's `src` tree is scanned. Repos with no TS still yield their repo node and don't error. Both extractors share `hasPnpmWorkspace`/`implicitPackageName` from `shared.ts` so their derived `packageKey` stays identical — no natural-key drift between the two Tier-1 extractors for the same repo.

New fixture tests: single-package repo with a real `src` tree, repo with `package.json` and no TS, and a bare directory with neither. **Prove-red performed**: reverted the three extractor files to `origin/test` via `git checkout origin/test -- <files>`, ran the new test file — 3 of 5 tests failed against the old discovery logic — then restored from the committed fix via `git checkout HEAD -- <files>` and confirmed green again.

## Fix 2: revkg CLI connection settings

The CLI borrowed `@revealui/db`'s shared pool, whose server-tuned defaults (5s connect timeout, 10s query/statement timeout) killed real fleet scans against a cold Neon compute — a connect timeout on the first run, then a mid-scan "Query read timeout" on the largest repo.

`revkg` now opens its own pool sized for a long-running ingest (30s connect timeout, 5-minute query/statement timeout, `max: 5`). This required adding `createPool` to `@revealui/db/pool` rather than importing `pg` directly in the CLI — the repo's `raw-SQL` `direct-import` gate (`scripts/validate/raw-sql.ts`, enforced in the pre-push gate) only allows a direct `pg` import inside `packages/db/src/`. `createPool` reuses `getConnectionIdentity` (exported since #1859) for the URL-vs-discrete resolution, and deliberately skips the shared pool's `onPoolConnect` setup handler — that handler's `SET statement_timeout` is pinned to the shared pool's 10s budget and would silently override a longer override on every connection. `statement_timeout`/`query_timeout` passed to `createPool` are applied by the `pg` driver itself as connection startup parameters (`Client#getStartupConf`), so no manual `SET` is needed for a non-shared pool.

Also fixed: the header comment's env-var precedence (`DATABASE_URL` wins over `POSTGRES_URL`, not the reverse — it said the opposite), and connection errors now print the target hostname (parsed via `new URL`, never the full connection string or credentials) so the next person can tell localhost from Neon at a glance.

## Fix 3: shared pool SET bug (pre-existing, `packages/db/src/pool.ts`)

`SET statement_timeout TO $1` failed on **every** pool connection with "syntax error at or near $1" — Postgres' `SET` command does not accept bind parameters. The intended 10s statement timeout has never actually applied anywhere, and every connect logged the error.

Extracted `formatValidatedStatementTimeoutMs`, which validates the value is a finite positive integer (it is always the already `parseInt`'d `poolConfig.statement_timeout`, never user input) before interpolating it as a literal — no injection surface. The three setup statements (timezone, statement_timeout, track_io_timing) now run as one multi-statement query on a single `await`, instead of three separate sequential `client.query()` calls on the same client. Unit tests cover the validator's edge cases (non-finite, non-integer, zero/negative, non-numeric string) plus a test that drives the pool's `'connect'` handler through a mock client and asserts the emitted SQL never contains `$1`.

## Verification actually run

- `pnpm --filter @revealui/knowledge-graph test` — 9 files / 38 tests passed
- `pnpm --filter @revealui/db test` — 54 files / 1182 passed, 5 skipped
- `pnpm --filter @revealui/knowledge-graph typecheck` — clean
- `pnpm --filter @revealui/db typecheck` — clean
- `pnpm --filter "@revealui/knowledge-graph..." build` — full dependency chain builds clean
- `pnpm --filter "server..." build` + `pnpm --filter server typecheck` — clean
- `pnpm --filter admin typecheck` (after `apps/admin` Next.js build) — clean
- `npx biome check` (scoped to changed files) — clean
- `npx tsx scripts/validate/raw-sql.ts` — PASS (0 violations)
- Internal fleet security-review checker, run against this diff — no security-sensitive files touched, no coordinator gate required
- Full pre-push gate (24 checks including Raw-SQL, Boundary validation, Security audit) — PASS

## Escalations for Fable

None required for merge. Two notes for awareness, out of scope for this batch:

1. **Implicit-package node kind is always `'package'`**, never `'app'`, since there's no reliable non-speculative signal (no `apps/`-prefix convention exists for a single-package repo). If ontology-precision on this matters for downstream graph consumers, a follow-up could add an app-detection heuristic (e.g. presence of `index.html` / `vite.config.*` at repo root) — deliberately not attempted here to avoid inventing untested heuristics beyond the stated fix scope.
2. Unrelated pre-existing drift noticed in passing: `packages/knowledge-graph/package.json`'s `author` field still carries a retired email address from a prior identity migration. This is repo-wide (likely every `package.json`), not scoped to this PR, and left untouched.

## Note on a stale worktree

A stale local worktree from the now-merged #1859 (`getConnectionIdentity`/pool connectionString fix, referenced above) was found still checked out. It appears not to have been cleaned up after merge. Flagging for owner cleanup since worktree removal is a disposition action outside agent scope.

### DIFF
--- a/packages/db/src/__tests__/pool-statement-timeout.test.ts
+++ b/packages/db/src/__tests__/pool-statement-timeout.test.ts
@@ -1,0 +1,138 @@
+/**
+ * Regression tests for the pool `onPoolConnect` SET-statement bug (GAP-349
+ * fleet scan incident): `SET statement_timeout TO $1` fails on every pool
+ * connection with "syntax error at or near $1" because Postgres' `SET`
+ * command does not accept bind parameters. The fix validates the timeout
+ * value and interpolates it as a literal instead.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mockPoolInstance = {
+  on: vi.fn(),
+  connect: vi.fn(),
+  end: vi.fn(),
+  query: vi.fn(),
+  totalCount: 10,
+  idleCount: 5,
+  waitingCount: 0,
+};
+
+vi.mock('pg', () => {
+  class MockPool {
+    on = mockPoolInstance.on;
+    connect = mockPoolInstance.connect;
+    end = mockPoolInstance.end;
+    query = mockPoolInstance.query;
+    totalCount = mockPoolInstance.totalCount;
+    idleCount = mockPoolInstance.idleCount;
+    waitingCount = mockPoolInstance.waitingCount;
+  }
+  return { Pool: MockPool };
+});
+
+vi.mock('@revealui/utils/database', () => ({
+  getSSLConfig: vi.fn(() => false),
+}));
+
+vi.mock('@revealui/utils/logger', () => ({
+  logger: {
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    debug: vi.fn(),
+  },
+}));
+
+let originalEnv: NodeJS.ProcessEnv;
+
+describe('formatValidatedStatementTimeoutMs', () => {
+  beforeEach(() => {
+    originalEnv = { ...process.env };
+    process.env.NODE_ENV = 'test';
+    vi.clearAllMocks();
+  });
+
+  afterEach(() => {
+    process.env = { ...originalEnv };
+    vi.resetModules();
+  });
+
+  it('formats a finite positive integer as a bare numeric string', async () => {
+    const { formatValidatedStatementTimeoutMs } = await import('../pool.js');
+    expect(formatValidatedStatementTimeoutMs(10000)).toBe('10000');
+    expect(formatValidatedStatementTimeoutMs(1)).toBe('1');
+  });
+
+  it('accepts a numeric string and formats it as a bare numeric string', async () => {
+    const { formatValidatedStatementTimeoutMs } = await import('../pool.js');
+    expect(formatValidatedStatementTimeoutMs('300000')).toBe('300000');
+  });
+
+  it('rejects non-finite values', async () => {
+    const { formatValidatedStatementTimeoutMs } = await import('../pool.js');
+    expect(() => formatValidatedStatementTimeoutMs(Number.NaN)).toThrow(
+      'invalid statement_timeout value',
+    );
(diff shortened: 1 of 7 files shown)


RevealUIStudio:
## Fable review (GAP-349 P2.5 scan-hardening)

Verdict: APPROVE. No blocking findings; no security gate applies (fleet checker: no sensitive files).

Verified in the diff:
- Fix 1: the implicit single-package fallback derives its name from one shared helper used by both extractors, so natural keys and deterministic node ids stay convergent between the workspace and ts-project passes. Fixture tests cover the three repo shapes and were proven red against the old discovery logic.
- Fix 2: createPool in @revealui/db is the right seam (respects the raw-SQL direct-import gate), and correctly skips the shared onPoolConnect handler so the CLI's longer statement budget is not overridden per-connection; timeouts ride pg's own config path instead. Host-only error reporting, never the URL.
- Fix 3: the SET interpolation is digits-only behind an integer validator with a clear why-comment, and the three setup statements now run as one sequential round trip, which also removes the client.query deprecation race.

Escalation rulings: (1) implicit-package kind staying 'package' rather than guessing 'app' is ACCEPTED, a heuristic would misclassify more than it helps; (2) the retired author email in package.json is repo-wide drift, queued for the stale-metadata sweep class, not this PR; (3) the stale #1859 worktree was already removed by the coordinator before this report.

Merge remains an owner action. After merge: git pull, rebuild the knowledge-graph chain, and re-run the fleet scan; the raised timeouts are now defaults so no env overrides are needed beyond DATABASE_URL.

PR #1867, opened 2026-07-12
fix(security): audit stage 0 — make failures loud, prove bugs red

RevealUIStudio:
## Summary

Stage 0 of the approved audit-remediation design. Fixes NO behavior — makes audit write failures impossible to miss, and writes red-first tests that prove today's bugs, so Stage 1's fixes are independently verifiable.

**Updated twice after independent adversarial review** ([first verdict](https://github.com/RevealUIStudio/revealui/pull/1867#issuecomment-4949994129), [my response](https://github.com/RevealUIStudio/revealui/pull/1867#issuecomment-4950054305)). See "Response to review" below for the full history.

**Part 1 — Red-first tests, all four now CHARACTERIZATION tests (real migrated PGlite schema, not the existing `vi.fn()` mock):**

Every one of the four tests below is a characterization test: it pins an exact, falsifiable fact about TODAY's actual (broken) behavior — a Postgres error code + a named constraint, an exact row count, a byte-level canonical-form mismatch, or a specific chained-signature identity — and **all four currently PASS, with zero `it.fails()` anywhere.** `pnpm vitest run packages/db` and `pnpm --filter server test` (the real CI commands) are both fully green.

- `packages/db/src/__tests__/audit-store.integration.test.ts` — a harness-sanity test (a valid insert must persist), then an explicit assertion that the `audit_log` CHECK constraint rejects `@revealui/security`'s severity vocabulary (`low`/`medium`/`high`) with Postgres `code: '23514'` and `constraint: 'audit_log_severity_check'`.
- `apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts` — a harness-sanity test, then: (a) each of the three middleware severities fails with that same specific cause and 0 rows land; (b) the jsonb payload/signature round-trip breaks specifically because Postgres reorders jsonb object keys (asserted by comparing canonical `JSON.stringify` forms of the round-tripped vs. original payload, after confirming every other signed field round-trips identically); (c) **DEFECT: the chain head advances on a rejected insert** — after a valid→rejected→valid sequence, the second valid row's `previousSignature` is asserted to NOT equal the first valid row's signature, and IS asserted (via `verifyAuditSignature`, since the rejected write's signature was never persisted anywhere to read back) to be exactly the signature that would have been computed for the rejected, never-persisted event.

Why characterization, not "assert the fix and let it fail red": `pnpm --filter server test` is the actual CI `Unit Tests` job every other PR in this repo shares. A test that asserts the desired-but-not-yet-true behavior would leave that job permanently red until Stage 1 lands, which is not an acceptance criterion, it's a broken shared build. Stage 1 fixing each underlying bug makes its pinned assertion FALSE — at that point the specific test fails, forcing Stage 1's author to consciously rewrite it to the new correct behavior. That failure-on-fix is the acceptance-criteria mechanism; a red test on a shared branch today is not.

**Part 2 — Make failure loud:**
- `AuditSystem.log()` now returns the full event and throws `AuditWriteError` (event + cause) on a failed write.
- New `packages/security/src/audit-write-failures.ts`: classifies failures (`constraint_violation` / `missing_secret` / `schema_mismatch` / `db_error`) and tracks a failure ratio over a rolling window of the last 500 attempts, escalating to `logger.error` past 1%. Counts SILENTLY — no per-attempt or per-failure log line; every call site already owns its own event-level logging at its own cadence (e.g. `middleware/audit.ts`'s existing 1-in-10 throttle).
- Fixed five silent catches: `middleware/audit.ts`, `api-keys.ts`, `a2a.ts`, admin's setup route, `scripts/admin/bootstrap.ts`, and a fifth found via grep on the same path family, `cron/uptime-check.ts`.
- Removed `PostgresAuditStorage`'s `onConflictDoNothing()` — event ids are fresh `crypto.randomUUID()`s, never idempotency keys, so a duplicate can only mean a bug.
- `validateStartup()` now validates the effective audit HMAC secret at boot (both hosted and forge production modes), duplicating (not replacing) `getAuditSecret()`'s existing per-write defense-in-depth throw.

**Part 3 — the `as never` validator (a speed bump, not a class killer — see below):**
- New CI check `scripts/validate/as-never-values.ts` (TypeScript compiler API, zero authored regex): flags `as never` casting a drizzle `.values()` argument. Wired into `pnpm gate` phase 1 (covered by `gate:quick`).
- Hardened against four of six bypasses an adversarial review demonstrated: parenthesized casts, angle-bracket `<never>` casts, ternary branches, and a single hop through a hoisted local variable (the bypass most likely to occur naturally). Two remain uncaught by design, documented in the function's own doc comment: a spread of a value assembled/cast elsewhere (needs real data-flow analysis) and `as any` (a different, broader pattern this check does not police).
- Repo-wide grep found exactly two genuine instances, both pre-existing, neither fixed here (Stage 1 territory): `apps/admin/src/app/api/setup/route.ts:104` and `scripts/admin/bootstrap.ts:205` — both allowlisted with a reason.

## Response to review (full history)

**Round 1** — independent non-author review ([comment](https://github.com/RevealUIStudio/revealui/pull/1867#issuecomment-4949994129)) returned REQUEST-CHANGES with two blockers:
- **B1** — the four red tests proved nothing specific (`it.fails()` passes on any throw, including a broken harness — demonstrated with a reproducible false-green in an unbuilt workspace). Fixed by dropping `it.fails()` for plain `it()` + specific `expect(...).rejects.toMatchObject({ cause: { code, constraint } })` assertions, harness-sanity tests, and `toBeDefined()` guards.
- **B2** — the new counter logged every failure, defeating `middleware/audit.ts`'s existing 1-in-10 throttle, shipping an ~11x log-volume increase into a path that fails 100% of the time in production today. Fixed: the counter is silent; only the rolling-ratio escalation logs.
- Also fixed: **S1** (ratio was a lifetime cumulative count and could only ever escalate once — now a rolling window that re-arms) and **S3** (the `as never` validator over-claimed "kills the class" — hardened against 4/6 demonstrated bypasses, downgraded the claim to match reality for the other 2).

**Round 2** — after Round 1, the chain-head test asserted the DESIRED correct behavior and genuinely failed today (by design, mirroring what B1's fix pattern implied for the OTHER three tests but not this one). That's wrong for a test inside the actual shared CI `Unit Tests` job: a knowingly-red test there blocks every other PR's test run, not just this one, until Stage 1 lands. Fixed by inverting it to a characterization test like the other three — see Part 1 above. `pnpm vitest run packages/db` and `pnpm --filter server test` are both now fully green with zero `it.fails()`.

Two review suggestions (S2 — the `missing_secret` classification couples to a string literal authored in a different package; S5 — a failed audit write's error message can carry request PII, pre-existing behavior this PR's edited line doesn't change) were not in the fix-now set across either round and are left for the owner to decide whether to fold into this PR or track separately. S4 (the new startup secret-length check) was owner-checked against revvault and cleared.

## Deviations from the literal task spec (and why)

- The task asked for all 4 red-first tests in one file at `packages/db/src/__tests__/audit-store.integration.test.ts`. Split instead: the severity/CHECK-constraint bug (schema-level, owned by packages/db) stayed there; the bugs specific to `PostgresAuditStorage`'s private HMAC-chain state moved to `apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts`, because testing them from packages/db would require packages/db to import apps/server source — an illegal reverse dependency in this monorepo.
- "Prod uses in-memory storage" (from the original task context) is true for the Vercel serverless surface (`apps/server/src/index.ts`'s production branch never calls `audit.setStorage(new PostgresAuditStorage())`), but the Fly worker (`apps/server/src/worker.ts`, the actual long-running production entry) DOES wire Postgres storage unconditionally. Noted for accuracy; doesn't change Stage 0 scope.
- Classifier bug found and fixed during red-proving: drizzle-orm wraps every driver error in a `DrizzleQueryError` whose own `.code` is `undefined` — the real Postgres error (and its `.code`) lives one level down at `.cause`. `classifyAuditWriteFailure` checks both.

## Test results (real numbers, current head)

- `pnpm vitest run packages/db`: 55 files, **1184 passed, 5 skipped, 0 failed**.
- `pnpm --filter server test` (apps/server's own configured command — the actual CI `Unit Tests` job): 156 files, **2893 passed, 0 failed, 1 skipped**.
- **GitHub Actions on this PR head:** `Unit Tests` — pass. `Typecheck` — pass. `Build`, `Integration Tests`, `Integration Tests (extended)`, `Quality`, `Security Gate`, `CodeQL`, `Drizzle Migrations`, `Secret Scanning`, `Dependency Review`, `Submodule Audit` — all pass. `Security review gate` — **fail, expected**: gated on the owner applying `sec-review:approved`, not a code problem.
- `pnpm --filter @revealui/security typecheck`, `pnpm --filter admin typecheck`, `pnpm --filter @revealui/db typecheck`: all clean.
- `pnpm gate:quick`: **PASS** (all phase-1 checks including the `as-never on drizzle .values()` check).
- Full-repo `tsc --noEmit -p tsconfig.json` OOMs/times out in this sandbox regardless of heap size — infeasible to run directly; verified per-package instead (matches how CI's `pnpm -r typecheck` actually gates).

## Not merging

Security-sensitive surface — needs a recorded non-author verdict before merge per the fleet's review-gate convention. Do not merge without `sec-review:approved`. All four Stage-0 tests are green today; Stage 1 should expect specific, individual test failures as it lands each fix, and should treat that failure as its acceptance signal to rewrite the pinned assertion to the new correct behavior.

### DIFF
--- a/apps/admin/src/app/api/setup/route.ts
+++ b/apps/admin/src/app/api/setup/route.ts
@@ -1,3 +1,4 @@
+import crypto from 'node:crypto';
 import { type BootstrapResult, bootstrap, type RevealUILike } from '@revealui/setup/bootstrap';
 import { logger } from '@revealui/utils/logger';
 import { NextResponse } from 'next/server';
@@ -95,19 +96,51 @@ export async function POST(request: Request): Promise<NextResponse<BootstrapResu
       try {
         const { hostname } = await import('node:os');
         const { auditLog } = await import('@revealui/db/schema');
-        await db.insert(auditLog).values({
-          event: 'admin.bootstrap.completed',
-          actor: 'web',
-          severity: 'info',
-          meta: {
+        const { classifyAuditWriteFailure, recordAuditWriteResult } = await import(
+          '@revealui/core/security'
+        );
+        const eventId = crypto.randomUUID();
+        try {
+          await db.insert(auditLog).values({
+            event: 'admin.bootstrap.completed',
+            actor: 'web',
+            severity: 'info',
+            meta: {
+              email: parsed.data.email,
+              source: 'web',
+              hostname: hostname(),
+              seeded: parsed.data.seed ?? true,
+            },
+          } as never);
+          recordAuditWriteResult({ ok: true, eventId, eventType: 'admin.bootstrap.completed' });
+        } catch (auditError) {
+          const reason = classifyAuditWriteFailure(auditError);
+          recordAuditWriteResult({
+            ok: false,
+            reason,
+            eventId,
+            eventType: 'admin.bootstrap.completed',
+          });
+          // Non-fatal — the admin account was already created above. Loud,
+          // classified log instead of a silent catch: this writer targets
+          // columns (`event`/`actor`/`meta`) that do not exist on audit_log
+          // (real columns are `event_type`/`agent_id`/`payload`) and has
+          // never successfully persisted a row in any environment. Stage 1
+          // fixes the column mapping; Stage 0 only makes the failure visible.
+          logger.error('Bootstrap audit log write failed', {
             email: parsed.data.email,
-            source: 'web',
-            hostname: hostname(),
-            seeded: parsed.data.seed ?? true,
-          },
-        } as never);
-      } catch {
-        // Non-fatal — audit log may not be available in all environments
+            eventId,
+            reason,
+            error: auditError instanceof Error ? auditError.message : String(auditError),
+          });
+        }
+      } catch (auditSetupError) {
+        // Non-fatal — audit log subsystem may not be available in all environments
+        logger.error('Bootstrap audit log setup failed', {
+          email: parsed.data.email,
+          error:
+            auditSetupError instanceof Error ? auditSetupError.message : String(auditSetupError),
+        });
       }
     } catch (dbError) {
       // DB client unavailable — the admin was still created by bootstrap(), but
--- a/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts
+++ b/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts
@@ -7,17 +7,34 @@
  *
  * `packages/db/src/__tests__/audit-store.integration.test.ts` covers the
  * severity-vocabulary CHECK-constraint bug at the schema level (owned by
- * packages/db). The three bugs here are specific to THIS class's private
- * module state (the HMAC signing chain, `lastSignature`) and live in
- * apps/server, so they're tested here rather than from packages/db — testing
- * them from packages/db would require packages/db to import apps/server
- * source, an illegal reverse dependency in this monorepo.
+ * packages/db). The bugs here are specific to THIS class's private module
+ * state (the HMAC signing chain, `lastSignature`) and live in apps/server, so
+ * they're tested here rather than from packages/db — testing them from
+ * packages/db would require packages/db to import apps/server source, an
+ * illegal reverse dependency in this monorepo.
  *
- * RevealUI audit-remediation Stage 0: these are red-first tests documenting
- * bugs Stage 1 fixes. Wrapped in `it.fails()` (Vitest's native
- * "expected to fail" API) so CI stays green until Stage 1 removes the
- * wrapper — see the sibling packages/db test for the same convention
- * decision.
+ * RevealUI audit-remediation Stage 0. Each test below pins an exact,
+ * falsifiable fact about TODAY's behavior — a Postgres error code + a named
+ * constraint, an exact row count, a byte-level canonical-form mismatch, or a
+ * signature identity — not merely "something threw". `it.fails()` was tried
+ * first and rejected: it passes on ANY throw, including a broken harness
+ * (unbuilt package, missing schema import, migrations that never applied),
+ * so it proves nothing. Plain `it()` with a specific `expect(...)` cannot be
+ * satisfied by an unrelated failure, and a harness that cannot reach the
+ * assertion fails loudly instead of reporting green.
+ *
+ * The severity-rejection and severity-persistence tests below currently
+ * PASS: they pin the real, current (broken) behavior with enough specificity
+ * that nothing except the actual documented defect can satisfy them. Stage 1
+ * fixing the severity vocabulary will make these assertions false — at that
+ * point they FAIL, forcing Stage 1's author to consciously rewrite them to
+ * the new correct behavior. That failure-on-fix is the acceptance-criteria
+ * mechanism; it is not a bug in these tests.
+ *
+ * The chain-head test asserts a standing invariant (previous_signature must
+ * point at the last row that actually persisted) that should hold both
+ * before and after Stage 1. It FAILS today for real — no wrapper, no
+ * inversion — and should keep passing unmodified once Stage 1 lands.
  */
 import { randomUUID } from 'node:crypto';
 import type { AuditEvent } from '@revealui/security';
@@ -34,6 +51,11 @@ vi.mock('@revealui/db', async () => {
   const actual = await vi.importActual<typeof import('@revealui/db')>('@revealui/db');
   return { ...actual, getClient: () => testDb.drizzle };
 });
+
+/** SQLSTATE + constraint name asserted throughout — see packages/db/src/schema/audit-log.ts:57. */
+const SEVERITY_CHECK_VIOLATION = {
+  cause: { code: '23514', constraint: 'audit_log_severity_check' },
+} as const;
 
 function makeEvent(overrides: Partial<AuditEvent> = {}): AuditEvent {
   return {
@@ -66,73 +88,155 @@ describe('PostgresAuditStorage — real PGlite schema, not a vi.fn() mock', () =
     delete process.env.REVEALUI_AUDIT_HMAC_SECRET;
   });
 
-  it.fails('persists every severity apps/server/src/middleware/audit.ts emits (Stage 1)', async () => {
+  // ── Harness sanity ─────────────────────────────────────────────────────
+  // Must run (and pass) before any of the bug-specific tests below are
+  // trusted. If this fails, the harness itself is broken (unbuilt package,
+  // missing @revealui/db/schema export, migrations that didn't apply, the
+  // getClient() mock not wired) and every other result in this file is
+  // meaningless until this is green again.
+
+  it('harness sanity: a valid severity persists and signs', async () => {
+    const store = new PostgresAuditStorage();
+    const { auditLog } = await import('@revealui/db/schema');
+
+    const event = makeEvent({ severity: 'critical' });
+    await store.write(event);
+
+    const rows = await testDb.drizzle.select().from(auditLog).where(eq(auditLog.id, event.id));
+    expect(rows).toHaveLength(1);
+    expect(rows[0]?.signature).not.toBeNull();
+  });
+
+  // ── Severity vocabulary vs the CHECK constraint ─────────────────────────
+
+  it('every severity apps/server/src/middleware/audit.ts emits is rejected by the CHECK constraint — 0 of 3 rows land (Stage 1 must change this)', async () => {
     const store = new PostgresAuditStorage();
     const { auditLog } = await import('@revealui/db/schema');
 
     // The request middleware emits exactly these three severities
-    // (apps/server/src/middleware/audit.ts:59). Mirror the middleware's own
-    // catch-and-swallow (apps/server/src/middleware/audit.ts:80) so this test
-    // observes what production actually persists, not an unhandled rejection.
+    // (apps/server/src/middleware/audit.ts:59). Capture each outcome
+    // individually — do not swallow-and-count, or an unrelated failure
+    // (wrong error, different cause) would be indistinguishable from the
+    // real defect.
+    const outcomes: Array<{ severity: 'low' | 'medium' | 'high'; error: unknown }> = [];
     for (const severity of ['low', 'medium', 'high'] as const) {
-      await store.write(makeEvent({ severity })).catch(() => {});
+      try {
+        await store.write(makeEvent({ severity }));
+        outcomes.push({ severity, error: undefined });
+      } catch (error) {
+        outcomes.push({ severity, error });
+      }
+    }
+
+    for (const { severity, error } of outcomes) {
+      expect(error, `severity '${severity}' should have thrown a check_violation`).toMatchObject(
+        SEVERITY_CHECK_VIOLATION,
+      );
     }
 
     const rows = await testDb.drizzle.select().from(auditLog);
-    expect(rows).toHaveLength(3);
+    expect(rows).toHaveLength(0);
   });
 
-  it.fails('round-trips a payload through jsonb without breaking the signature (Stage 1)', async () => {
+  // ── jsonb round-trip vs the HMAC signature ──────────────────────────────
+
+  it('the recomputed signature does not match because jsonb reorders payload.metadata keys, not because any value changed (Stage 1 must fix this)', async () => {
     const store = new PostgresAuditStorage();
     const { auditLog } = await import('@revealui/db/schema');
 
     // 'z' then 'a' (same length) — Postgres jsonb normalizes object key
-    // order on storage, so the keys come back in a different order than
-    // they were written in. Severity 'critical' is valid in both the
-    // security package's vocabulary AND the DB CHECK constraint, so this
-    // write succeeds and isolates the jsonb round-trip bug specifically.
+    // order on storage (same-length keys sort lexically), so the keys come
+    // back in a different order than they were written in. Severity
+    // 'critical' is valid in both the security package's vocabulary AND the
+    // DB CHECK constraint, so this write succeeds and isolates the jsonb
+    // round-trip bug specifically — nothing else about the write can be the
+    // cause of what follows.
     const event = makeEvent({ metadata: { z: 1, a: 2 } });
     await store.write(event);
 
     const [row] = await testDb.drizzle.select().from(auditLog).where(eq(auditLog.id, event.id));
     expect(row).toBeDefined();
-    expect(row?.signature).not.toBeNull();
+    if (!row) throw new Error('unreachable: guarded by the assertion above');
+    expect(row.signature).not.toBeNull();
+
+    // Every OTHER signed field round-trips identically — isolates the
+    // mismatch below to payload.metadata specifically, not some other drift.
+    expect(row.timestamp.toISOString()).toBe(event.timestamp);
+    expect(row.eventType).toBe(event.type);
+    expect(row.severity).toBe(event.severity);
+    expect(row.agentId).toBe(event.actor.id);
+
+    const roundTrippedMetadata = (row.payload as { metadata?: unknown } | null)?.metadata;
+    // The VALUE is unchanged (same keys, same values) ...
+    expect(roundTrippedMetadata).toEqual(event.metadata);
+    // ... but the canonical (JSON.stringify) representation is NOT
+    // byte-identical, because jsonb reordered the keys. This is exactly why
+    // verifyAuditSignature — which recomputes the HMAC over
+    // JSON.stringify(payload) — cannot reproduce the original signature at
+    // read time: the bytes it hashes are provably different even though the
+    // data is semantically identical.
+    expect(JSON.stringify(roundTrippedMetadata)).not.toBe(JSON.stringify(event.metadata));
 
     const verified = verifyAuditSignature(
       {
-        timestamp: row?.timestamp.toISOString() ?? '',
-        eventType: row?.eventType ?? '',
-        severity: row?.severity ?? '',
-        agentId: row?.agentId ?? '',
-        payload: row?.payload,
+        timestamp: row.timestamp.toISOString(),
+        eventType: row.eventType,
+        severity: row.severity,
+        agentId: row.agentId,
+        payload: row.payload,
       },
-      row?.signature ?? '',
-      row?.previousSignature ?? null,
+      row.signature ?? '',
+      row.previousSignature,
     );
-    expect(verified).toBe(true);
+    expect(verified).toBe(false);
   });
 
-  it.fails('does not advance the chain head on a rejected insert (Stage 1)', async () => {
+  // ── Hash-chain head vs a rejected insert ─────────────────────────────────
+
+  it('does not advance the chain head on a rejected insert — previous_signature must point at the last row that actually PERSISTED (Stage 1 must fix this)', async () => {
     const store = new PostgresAuditStorage();
     const { auditLog } = await import('@revealui/db/schema');
 
-    // Rejected insert: 'low' is not in the audit_log CHECK vocabulary.
-    await store.write(makeEvent({ severity: 'low' })).catch(() => {});
-
-    // Immediately follow with a valid insert.
-    const validEvent = makeEvent({ severity: 'critical' });
-    await store.write(validEvent);
-
-    const [row] = await testDb.drizzle
+    // Establish a known-good baseline: one row that genuinely persists.
+    const firstValid = makeEvent({ severity: 'critical' });
+    await store.write(firstValid);
+    const [firstRow] = await testDb.drizzle
       .select()
       .from(auditLog)
-      .where(eq(auditLog.id, validEvent.id));
+      .where(eq(auditLog.id, firstValid.id));
+    expect(firstRow).toBeDefined();
+    if (!firstRow) throw new Error('unreachable: guarded by the assertion above');
+    expect(firstRow.signature).not.toBeNull();
 
-    // Nothing was ever actually persisted before this row (the rejected
-    // write never landed), so its previous_signature must be null. Today it
-    // isn't: postgres-audit-storage.ts assigns `lastSignature` from the
-    // rejected write's computed signature BEFORE the insert that throws, so
-    // the module-level chain head advances on a write that never landed.
-    expect(row?.previousSignature).toBeNull();
+    // Rejected insert: 'low' is not in the audit_log CHECK vocabulary.
+    // Confirm both the specific rejection reason AND that it truly never
+    // landed — the invariant below is meaningless if this write silently
+    // persisted after all.
+    const rejected = makeEvent({ severity: 'low' });
+    await expect(store.write(rejected)).rejects.toMatchObject(SEVERITY_CHECK_VIOLATION);
+    const rejectedRows = await testDb.drizzle
+      .select()
+      .from(auditLog)
+      .where(eq(auditLog.id, rejected.id));
+    expect(rejectedRows).toHaveLength(0);
+
+    // Immediately follow with a second valid insert.
+    const secondValid = makeEvent({ severity: 'critical' });
+    await store.write(secondValid);
+    const [secondRow] = await testDb.drizzle
+      .select()
+      .from(auditLog)
+      .where(eq(auditLog.id, secondValid.id));
+    expect(secondRow).toBeDefined();
+    if (!secondRow) throw new Error('unreachable: guarded by the assertion above');
+
+    // The chain head must point at the last PERSISTED row (firstRow), never
+    // at the rejected write that never landed. Today it doesn't:
+    // postgres-audit-storage.ts assigns the module-level `lastSignature` from
+    // the rejected write's COMPUTED signature BEFORE the insert that throws,
+    // so the chain head advances on a write that never made it into the
+    // table. This assertion fails today and should keep passing, unmodified,
+    // once Stage 1 fixes the ordering.
+    expect(secondRow.previousSignature).toBe(firstRow.signature);
   });
 });
--- a/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts
+++ b/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts
@@ -13,28 +13,35 @@
  * packages/db would require packages/db to import apps/server source, an
  * illegal reverse dependency in this monorepo.
  *
- * RevealUI audit-remediation Stage 0. Each test below pins an exact,
- * falsifiable fact about TODAY's behavior — a Postgres error code + a named
- * constraint, an exact row count, a byte-level canonical-form mismatch, or a
- * signature identity — not merely "something threw". `it.fails()` was tried
- * first and rejected: it passes on ANY throw, including a broken harness
- * (unbuilt package, missing schema import, migrations that never applied),
- * so it proves nothing. Plain `it()` with a specific `expect(...)` cannot be
- * satisfied by an unrelated failure, and a harness that cannot reach the
- * assertion fails loudly instead of reporting green.
+ * RevealUI audit-remediation Stage 0. Every test below is a CHARACTERIZATION
+ * test: it pins an exact, falsifiable fact about TODAY's actual (broken)
+ * behavior — a Postgres error code + a named constraint, an exact row
+ * count, a byte-level canonical-form mismatch, or a specific chained
+ * signature identity — not merely "something threw". All four PASS today,
+ * on purpose, with no wrapper and no inversion: `pnpm --filter server test`
+ * must be fully green.
  *
- * The severity-rejection and severity-persistence tests below currently
- * PASS: they pin the real, current (broken) behavior with enough specificity
- * that nothing except the actual documented defect can satisfy them. Stage 1
- * fixing the severity vocabulary will make these assertions false — at that
- * point they FAIL, forcing Stage 1's author to consciously rewrite them to
- * the new correct behavior. That failure-on-fix is the acceptance-criteria
- * mechanism; it is not a bug in these tests.
+ * `it.fails()` was tried first here and rejected: it passes on ANY throw,
+ * including a broken harness (unbuilt package, missing schema import,
+ * migrations that never applied), so it proves nothing — and an earlier cut
+ * of the chain-head test asserted the DESIRED correct behavior instead of
+ * the defect, so it genuinely failed today, which would have put the CI
+ * `test` job (a shared-branch build every other PR depends on) permanently
+ * red until Stage 1 lands. Neither is acceptable. A plain `it()` with a
+ * specific `expect(...)` that asserts the CURRENT (buggy) behavior cannot be
+ * satisfied by an unrelated failure, a broken harness fails it loudly
+ * instead of reporting green, and it stays green in CI because it is true
+ * today.
  *
- * The chain-head test asserts a standing invariant (previous_signature must
- * point at the last row that actually persisted) that should hold both
- * before and after Stage 1. It FAILS today for real — no wrapper, no
- * inversion — and should keep passing unmodified once Stage 1 lands.
+ * Every test here pins real, current, verified behavior with enough
+ * specificity that nothing except the actual documented defect can satisfy
+ * it. Stage 1 fixing the underlying bug makes the pinned assertion FALSE —
+ * at that point the test fails, forcing Stage 1's author to consciously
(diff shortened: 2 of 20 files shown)


RevealUIStudio:
## Security review verdict (guardrail 2): REQUEST-CHANGES

Independent non-author review at PR head `1740801a`, re-derived from the code in a
fresh detached worktree (not the author's). I ran the tests, probed the validator's
AST matcher directly, and captured the real Postgres error myself rather than
taking the PR description on trust.

**No production-outage risk found.** I attacked that axis hardest and cleared it
(details in "What I cleared" below). The two blockers are a test suite that does not
prove what it claims, and a log-volume regression on a path that is failing 100% of
the time in production today.

---

## Blockers

### B1 — The four red tests are satisfied by *any* throw, so they prove nothing specific. Demonstrated, not theorized.

`it.fails()` passes whenever the test body throws, **for any reason**. None of the
four tests assert *why* the write failed.

[`packages/db/src/__tests__/audit-store.integration.test.ts:29-49`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/db/src/__tests__/audit-store.integration.test.ts#L29-L49)
contains no assertion at all — just a bare insert. The PR body claims this is
"confirmed via real Postgres `23514`", but `23514` appears nowhere in the test.

I proved the false-green is reachable. In a workspace where `packages/db` is not yet
built (`@revealui/db/schema` resolves to `./dist/schema/index.js`, which does not
exist — a fresh clone running `npx vitest run packages/db`), this test reports:

```
Test Files  1 passed (1)
     Tests  1 expected fail (1)
```

…while actually dying inside `createTestDb()` with
`Cannot find package '@revealui/db/schema'`. It never executes the insert. It never
touches the CHECK constraint. It is green, and it proves nothing.

For the record: the underlying bug **is** real. I built the package and captured the
error with my own probe:

```
CAUSE CODE: 23514
CAUSE MSG:  new row for relation "audit_log" violates check constraint "audit_log_severity_check"
```

So the finding is not "the bug is fake" — it is "these tests are not evidence of it,"
which defeats the entire stated purpose of Stage 0 (make Stage 1 independently
verifiable).

Same class of hole in
[`postgres-audit-storage.pglite.test.ts:115-137`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L115-L137):
the chain-head test has no `expect(row).toBeDefined()` guard (test 2 at
[:98](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L98)
correctly has one). If the valid `'critical'` write throws for any unrelated reason,
`row` is `undefined`, the assertion fails, and `it.fails()` still goes green.
[`:69-83`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L69-L83)
has the same weakness: it swallows every write error with `.catch(() => {})` and then
only asserts `rows.length === 3`, so any failure cause yields the same green.

**Fix:** pin the cause and drop `it.fails()` entirely. A test that asserts today's
broken behavior explicitly is a real proof *and* fails loudly when Stage 1 changes it:

```ts
await expect(
  db.drizzle.insert(auditLog).values({ ...withSeverity('low') }),
).rejects.toMatchObject({ cause: { code: '23514' } });
```

That cannot pass for a missing import, an unbuilt package, or a typo.

### B2 — `recordAuditWriteResult` emits an unthrottled `logger.warn` per failed write, defeating the throttle directly above it — on a path that fails 100% of the time in prod today.

[`packages/security/src/audit-write-failures.ts:135-139`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L135-L139)
logs `audit_write_failures_total` on **every** failure, with no rate limiting.
It is called from
[`postgres-audit-storage.ts:131-137`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/postgres-audit-storage.ts#L131-L137),
i.e. once per audit write.

That sits one layer *below*
[`middleware/audit.ts:88`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/middleware/audit.ts#L88)'s
deliberate 1-in-10 throttle (`FAILURE_LOG_INTERVAL = 10`). The new emit bypasses it.

Why this bites immediately, re-derived from the code:
- [`worker.ts:58`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/worker.ts#L58) wires `PostgresAuditStorage` unconditionally on the long-running Fly worker.
- [`middleware/audit.ts:60`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/middleware/audit.ts#L60) emits `severity: 'high' | 'medium' | 'low'`.
- [`audit-log.ts:57`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/db/src/schema/audit-log.ts#L57) allows only `('info', 'warn', 'critical')`.

So **every** request-level audit write on the worker fails with `23514` right now.
Audit-failure log lines go from ~0.1/request to ~1.1/request — roughly an 11x
increase, shipped into the exact condition that exists in production today.

The module's own header says it counts the ratio "not the count" precisely to avoid
per-event noise, and then emits per-event noise. Either throttle this warn, or drop
it and rely on the ratio alarm plus the middleware's existing throttled line.

---

## Suggestions

- **S1 — the "rolling" ratio is not rolling.** [`audit-write-failures.ts:124`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L124) documents a "rolling failure ratio", but the counters at [`:106`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L106)/[`:128`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L128) are cumulative and never decay. Once `ratioAlarmFired` latches at [`:146`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L146), re-arming requires the *lifetime* ratio to fall back under 1% — after N failures that needs ~100N subsequent successes. In practice the alarm fires at most once per process and cannot alert on a second, later failure spike. Use a windowed counter, or reset the window when the alarm fires.

- **S2 — `missing_secret` classification is a cross-package string coupling.** [`audit-write-failures.ts:71`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit-write-failures.ts#L71) matches `error.message.startsWith('Audit HMAC signing requires')` against a message authored in a *different* package ([`postgres-audit-storage.ts:32`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/postgres-audit-storage.ts#L32)). Reword that message and classification silently degrades to `db_error` — and the test at [`audit-write-failures.test.ts:16`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/__tests__/audit-write-failures.test.ts#L16) hardcodes the same literal, so it would still pass. Export a typed `MissingAuditSecretError` and classify with `instanceof`.

- **S3 — the `as never` validator is a speed bump, not a class killer.** No authored regex (confirmed: [`as-never-values.ts:177-208`](https://github.com/RevealUIStudio/revealui/blob/1740801a/scripts/validate/as-never-values.ts#L177-L208) uses `ts.createSourceFile` + `forEachChild`, constructs no `RegExp`) — that part is clean, and `as unknown as never` *is* caught. But I ran its exported matcher against variants and bypassed it 6 ways:

  | variant | result |
  |---|---|
  | `.values(x as never)` | CAUGHT |
  | `.values([x as never])` | CAUGHT |
  | `.values(x as unknown as never)` | CAUGHT |
  | `const p = {...} as never; .values(p)` | **MISSED** |
  | `.values(<never>{...})` | **MISSED** |
  | `.values(({...} as never))` | **MISSED** |
  | `.values(c ? (x as never) : y)` | **MISSED** |
  | `.values([...rowsCastElsewhere])` | **MISSED** |
  | `.values({...} as any)` | **MISSED** |

  The hoist-to-a-variable bypass is the one that matters — it is the most natural way anyone would write that code. `isAsNever` at [`:181-183`](https://github.com/RevealUIStudio/revealui/blob/1740801a/scripts/validate/as-never-values.ts#L181-L183) only inspects the direct argument node. It genuinely prevents *exact* recurrence of the two known instances, which has value; the PR body's "kill the `as never` class" overstates what it does.

- **S4 — one genuinely new boot-fail case; check the live value before deploying.** [`validate-startup.ts:281`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/validate-startup.ts#L281) now hard-fails boot when `REVEALUI_AUDIT_HMAC_SECRET` is *set but under 32 chars*. `getAuditSecret()` ([`postgres-audit-storage.ts:29`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/postgres-audit-storage.ts#L29)) enforces no length, so such a deploy signs fine today (HMAC-SHA256 accepts any key length) and would refuse to boot after this merge. It is already out-of-contract (`secretSchema = z.string().min(32)`, [`packages/config/src/schema.ts:14`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/config/src/schema.ts#L14)/[`:88`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/config/src/schema.ts#L88)), so the check is defensible — but confirm the live value's length in revvault before this reaches prod. Related: [`packages/config/src/modules/reveal.ts:29`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/config/src/modules/reveal.ts#L29) resolves the same var pair with `||` while the other two sites use `??` — a third resolution site with different empty-string semantics.

- **S5 — PII in failure logs (pre-existing, but this PR re-authors the line).** Drizzle's error message embeds the query params. I captured it verbatim: `params: 5701099d-…,2026-07-12T…,data.read,low,agent-1,{},[]`. [`middleware/audit.ts:94`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/middleware/audit.ts#L94) logs `cause.message`, so a failed request-audit write serializes the event payload into logs — including `actor.ip` and `userAgent` ([`middleware/audit.ts:64-67`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/middleware/audit.ts#L64-L67)). The old line logged `err.message` with identical content, so this is **not** a regression — but the PR is already editing this exact line and could log a redacted field set instead. Same shape at [`api-keys.ts`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/routes/api-keys.ts) (payload there is `credentialId`/`provider`/`label` — no key material, verified).

---

## Nits

- [`as-never-values-allowlist.json`](https://github.com/RevealUIStudio/revealui/blob/1740801a/scripts/validate/as-never-values-allowlist.json) pins exact line numbers (104, 205). Any unrelated edit above those lines shifts them and fails CI with a confusing message. Correct at head (validator passes clean), and arguably intentional as a re-review trigger — worth a comment saying so.

---

## What I independently re-derived and *cleared* (attacked, did not find)

These were the axes I was most suspicious of. All came back clean:

- **No new refuse-to-boot in a correctly-configured deploy.** The new check sits *after* the `NODE_ENV !== 'production'` early return ([`validate-startup.ts:218`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/validate-startup.ts#L218)), so it is production-only. `REVEALUI_SECRET` is required in **both** modes (`required-env.ts:29` hosted; `REQUIRED_IN_PRODUCTION_FORGE` at [`validate-startup.ts:137`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/lib/validate-startup.ts#L137) — so **forge/self-hosted is covered**, not a gap). I specifically hunted for a `??`-vs-`||` mismatch against `getAuditSecret()` and **did not find one** — both use `??`, so the fallback genuinely mirrors the write path. The only new failure mode is S4.
- **Removing `onConflictDoNothing()` cannot 500 a live request.** `event.id` is a fresh `crypto.randomUUID()` minted per `AuditSystem.log()` call ([`audit.ts`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/security/src/audit.ts)), never caller-supplied, so a duplicate-key throw is unreachable. Verified.
- **No new unhandled rejection.** There are exactly two `audit.log()` call sites: [`middleware/audit.ts:57`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/middleware/audit.ts#L57) (floated, but `.catch()`-terminated) and [`uptime-check.ts:43`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/routes/cron/uptime-check.ts#L43) (inside `try/catch`). Both handle the throw. All five rewritten catches still swallow — **none rethrow**. `PostgresAuditStorage.write()` rethrows, but its only consumer is `AuditSystem.log()`, caught at both sites. No lambda-crash path.
- **No memory leak, no hot loop.** The counters are a fixed 4-key record plus two ints; nothing accumulates per event.
- **No secret, signature, or HMAC value is logged anywhere** in the new code. The `missing_secret` error text names the env var, never its value. Axis clean.
- **Scope respected — this really is Stage 0.** `audit.setStorage(...)` is still inside the dev-only guard ([`index.ts:1313-1322`](https://github.com/RevealUIStudio/revealui/blob/1740801a/apps/server/src/index.ts#L1313-L1322)) and the severity vocabulary is still unmapped ([`audit-log.ts:57`](https://github.com/RevealUIStudio/revealui/blob/1740801a/packages/db/src/schema/audit-log.ts#L57)). No Stage 1 behavior change smuggled in.
- Validator runs clean at head (0 violations, 3160 files); allowlist reasons are non-empty and both entries resolve.

---

Blockers are B1 and B2. B1 in particular: Stage 0's whole contract is "prove the bugs
red so Stage 1 is verifiable," and as encoded, the proofs would survive a typo, a
missing import, or an unbuilt package. Pin the causes and this is a good change.

Not applying `sec-review:approved` and not merging — the owner applies the label on a
recorded verdict.

*Reviewed by an independent non-author agent (Opus 4.8). Findings re-derived from the
code at PR head; tests, validator, and the Postgres error captured by execution, not
by reading the PR description.*

RevealUIStudio:
Addressed both blockers and the two requested follow-ups in 9be9c4565 (on top of the reviewed 1740801a4).

**B1** — dropped `it.fails()` everywhere. All four tests are now plain `it()` with an explicit `expect(...).rejects.toMatchObject({ cause: { code: '23514', constraint: 'audit_log_severity_check' } })` (or the equivalent try/catch capture for the multi-attempt test), a `toBeDefined()` guard on every row lookup, and a harness-sanity test at the top of each file that must pass first. Reproduced your false-green concern was structurally eliminated by removing the inversion — an uncaught exception in a plain `it()` always fails it, there's no wrapper left to hide behind. The chain-head test now runs a genuine valid→rejected→valid sequence and asserts `previousSignature` equals the *first* valid row's signature, not merely `null`.

Ran each rewritten assertion for real: the severity-rejection and severity-persistence tests currently PASS (they pin today's actual, verified behavior — Stage 1 changing the vocabulary will make them false and fail, which is the point); the chain-head invariant test currently FAILS for real, no wrapper (confirmed via `pnpm --filter server test`: 2892 passed, 1 failed — that one failure, by design). Full numbers in the updated PR body.

**B2** — `recordAuditWriteResult()` no longer logs anything per attempt or per failure. Confirmed by re-running the pglite suite: the `audit_write_failures_total` stderr line that appeared in every failing-write test run before is gone. The module counts silently now; only the rolling-ratio escalation still logs.

**S1** — ratio is now computed over a fixed-size rolling window (last 500 attempts) instead of a lifetime cumulative count. Added a test that fires the alarm, drives 500 successes through to age the failures out of the window, confirms `failed === 0`, then drives a second independent spike and confirms `logger.error` is called a second time — proving the alarm re-arms rather than latching for the process lifetime.

**S3** — ran your six bypass cases against the matcher directly. Hardened it against four (parens, angle-bracket `<never>`, ternary branches, and the one you flagged as mattering most — a single hop through a hoisted local variable). The other two (a spread of a value cast/assembled elsewhere, and `as any`) are left uncaught by design — documented in the function's own doc comment and in the test file, rather than claimed as covered. Downgraded the PR body's "kill the class" framing to "speed bump" to match.

S2 (missing_secret string coupling) and S5 (PII in the failure log line) weren't in the coordinator's fix-now set — noted in the PR body for the owner to decide. S4 was already owner-checked against revvault and cleared, no change needed.

Not merging — still needs a recorded verdict, and this PR now intentionally ships one genuinely failing test (the chain-head invariant) as Stage 1's acceptance criterion.

RevealUIStudio:
**Re-verdict claim (guardrail 2):** a fresh Fable session is running the independent re-derivation of the post-review commits (`9be9c4565` + `713d8059e`) against the REQUEST-CHANGES verdict above — fresh detached worktree, tests re-run, blocker fixes verified against the recorded failure modes (not the response comment). Verdict comment to follow. If the original reviewer session is already mid-re-review, comment here and I stand down.

RevealUIStudio:
Fixed. The chain-head test (last commit, 713d8059e) is now a characterization test like the other three: it asserts the DEFECT as it exists today rather than the desired fix, renamed to lead with "DEFECT:" so the mechanism is unmissable, with a comment explaining that Stage 1 landing the fix should make this specific test fail, forcing a conscious rewrite to the correct invariant.

Confirmed real CI status on this PR head (not just local runs):

- **Unit Tests: pass**
- **Typecheck: pass**
- Build, Integration Tests, Integration Tests (extended), Quality, Security Gate, CodeQL, Drizzle Migrations, Secret Scanning, Dependency Review, Submodule Audit: all pass
- Security review gate: fail — expected, gated on the owner applying `sec-review:approved`

Local numbers backing that: `pnpm vitest run packages/db` — 1184 passed, 0 failed. `pnpm --filter server test` (the actual CI command) — 2893 passed, 0 failed, 1 skipped. Zero `it.fails()` anywhere in either file.

PR body updated to describe the characterization-test mechanism uniformly across all four tests, plus the full two-round review history. Not merging.

RevealUIStudio:
## Security review verdict (guardrail 2), re-review at `713d8059e`: APPROVE

Both blockers from my first review ([earlier comment](https://github.com/RevealUIStudio/revealui/pull/1867#issuecomment-4949994129)) are **closed**. I re-derived this from the code at the new head in a fresh detached worktree, re-ran the exercises that broke the previous cut, and ran one new one the author did not: I simulated Stage 1's fix to check the tests actually invert.

I did not take the fix summary as evidence. Every claim below I re-executed.

---

## B1 — CLOSED (verified by reproducing the exact failure I used to break the old cut)

My original finding was that `it.fails()` reported a **green** `1 expected fail` while dying inside `createTestDb()` on an unresolved `@revealui/db/schema`, never reaching the insert.

I reproduced that setup verbatim against the new head — same fresh worktree, `pnpm install`, **no build**, `packages/db/dist` confirmed absent:

```
Test Files  1 failed (1)
      Tests  2 failed (2)
```

Loudly red where it was previously green. The false-green is gone.

The defense holds for the right reason, too. `createTestDb()` now runs in `beforeEach`, so a broken harness fails **every** test in the file rather than being swallowed by an inverted wrapper. I confirmed the same on the apps/server file: with `@revealui/core` unbuilt, the harness-sanity test at [`postgres-audit-storage.pglite.test.ts:105`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L105) fails loudly with `Cannot find package '@revealui/core/security'` — exactly the "dead harness cannot report green" behavior it is there to provide. I tried to defeat it and could not: there is no path where the subject goes unexercised and the suite still reports green.

**Do they pass for the right reason?** Yes — each pins the specific defect, not "something threw":
- [`audit-store.integration.test.ts:84-88`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/db/src/__tests__/audit-store.integration.test.ts#L84-L88) — `rejects.toMatchObject({ cause: { code: '23514', constraint: 'audit_log_severity_check' } })`. Pins SQLSTATE **and** the constraint name. A missing import, an unbuilt package, or a typo cannot satisfy this.
- [`:119-146`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L119-L146) — captures each of the three severities' outcomes **individually** and asserts the same code+constraint on each, then asserts 0 rows. No more swallow-and-count.
- [`:150-199`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L150-L199) — isolates jsonb key reordering *specifically*: asserts every other signed field round-trips identically, asserts the metadata **value** is `toEqual` (unchanged) while its `JSON.stringify` is `not.toBe` (bytes differ), *then* asserts `verified === false`. It cannot pass on a generic signature mismatch.
- [`:203-272`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/__tests__/postgres-audit-storage.pglite.test.ts#L203-L272) — the strongest of the four. It does not merely assert "previousSignature is wrong"; it recomputes the HMAC and proves `secondRow.previousSignature` **is exactly the rejected write's signature**, chained from `firstRow.signature`. That identity can only hold if the chain head advanced on a write that never landed.

All four pass in a built workspace (I built `@revealui/db` + `@revealui/core` and ran them: `Tests 4 passed`).

### The inversion contract — I tested it rather than reasoning about it

The whole mechanism rests on "Stage 1's fix will falsify these." A characterization test that survives the fix is decoration, so I applied Stage 1's chain-head fix myself — moved the `lastSignature` assignment in [`postgres-audit-storage.ts:104-108`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/postgres-audit-storage.ts#L104-L108) to *after* the successful insert — and re-ran:

```
× DEFECT: the chain head advances on a rejected insert ...
AssertionError: expected '16799e909009f6b69d…' not to be '16799e909009f6b69d…'
      Tests  1 failed
```

It flips red with precisely the designed assertion (`previousSignature` now *does* equal `firstRow.signature` — the correct invariant). The inversion contract is real, not asserted. Reverted cleanly afterward.

Reasoning through the other three: fixing the severity mapping makes writes succeed, so the `toMatchObject` on the captured error and the `toHaveLength(0)` both flip. Fixing the canonical-form/jsonb bug makes `verified` true, flipping `expect(verified).toBe(false)`. Harness-sanity is a harness invariant and correctly stays green across Stage 1 — that one is not an acceptance criterion and does not need to be.

---

## B2 — CLOSED

[`audit-write-failures.ts`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts) now contains exactly **one** `logger` call ([`:164`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L164)), the latched ratio escalation. Grep confirms zero per-failure emits remain in the counter layer, and `audit_write_failures_total` is gone from the codebase entirely.

**Cannot hot-loop under the 100%-failure condition on the Fly worker.** With every write failing, the ratio pins at 1.0, stays above threshold, and `ratioAlarmFired` ([`:162-163`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L162-L163)) keeps it silent after the first line. The author's own tests encode this and they pass: *"never logs on an individual failure, no matter how many are recorded"* and *"escalates via logger.error exactly once while the ratio stays above threshold"*.

The reason-keyed signal I did not want to lose is still there: [`middleware/audit.ts:88-93`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/middleware/audit.ts#L88-L93) still logs `reason` on its throttled 1-in-10 line, and the escalation carries a `byReason` breakdown. Loudness preserved, amplification removed.

## S1 — CLOSED

Rolling window of the last 500 attempts ([`:98`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L98), [`:149-152`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L149-L152)), not lifetime cumulative. The alarm re-arms — the test *"re-arms and re-fires after the ratio recovers below threshold and spikes again"* passes.

## S3 — Substantially addressed, honestly scoped

I re-ran my 6 bypasses against the hardened matcher. **4 of 6 now caught** (hoisted variable, angle-bracket `<never>`, extra parens, ternary). The 2 remaining (spread of an externally-cast array, `as any`) are not just "documented" — they are **encoded as explicit passing tests** at [`as-never-values.test.ts:105`](https://github.com/RevealUIStudio/revealui/blob/713d8059/scripts/validate/__tests__/as-never-values.test.ts#L105) and [`:113`](https://github.com/RevealUIStudio/revealui/blob/713d8059/scripts/validate/__tests__/as-never-values.test.ts#L113), so the gaps are pinned rather than hidden. The "kill the class" claim is downgraded to "speed bump" in the source header. That is the right call and the right framing.

I probed 3 further variants and found 2 still uncaught: a `let` **reassigned** after declaration (`let p; p = {...} as never`), and a cast returned from a function (`const mk = () => ({...}) as never; .values(mk())`). Both are the same no-data-flow-analysis class the header already discloses. Not blocking — noting them so the gap set is complete.

---

## Did the fixes break anything? (the axis I previously cleared — re-checked)

- **Rolling window memory: bounded, not a leak.** `window` is capped at 500 entries by the `shift()` at [`:150-152`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L150-L152). Worst case ~500 small objects. No unbounded growth.
- **Harness-sanity → CI flake risk: acceptable, but thin margin.** `createTestDb()` now runs per-test in `beforeEach`. I measured 12–13s per bootstrap; the repo's own note cites 9–19s observed, against `hookTimeout: 30000` in both [`packages/db/vitest.config.ts:10`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/db/vitest.config.ts#L10) and [`apps/server/vitest.config.ts:56`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/vitest.config.ts#L56). That is only ~1.6x headroom at the worst observed bootstrap on a loaded runner. **Not a regression** — the apps/server file already used `beforeEach` + `createTestDb`, and moving the packages/db setup out of the test body into a hook actually *fixed* a latent problem (in the body it was bound by the 5000ms default `testTimeout`, which a 12s bootstrap would blow). Flagging the margin as a durable CI-hygiene item, not a blocker for this PR.
- Outage axis unchanged and still clear: no new refuse-to-boot path, no new throw on a live request, no new unhandled rejection. Nothing in the fix commits touches those surfaces.
- Validator runs clean on the repo (0 violations); fleet checker correctly still reports HOLD pending the label.

---

## Still open (all non-blocking, unchanged from my first review)

- **S2** — `missing_secret` is still classified by `error.message.startsWith('Audit HMAC signing requires')` ([`:86`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L86)), a cross-package string coupling to a message authored in apps/server. A typed error class would be sturdier. Fine to defer to Stage 1.
- **S4** — [`validate-startup.ts:281`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/lib/validate-startup.ts#L281) still hard-fails boot on a *set-but-under-32-char* `REVEALUI_AUDIT_HMAC_SECRET`, which signs fine today. Already out-of-contract per `secretSchema`, so the check is right — but **confirm the live value's length in revvault before this reaches prod.** This is the one item I would not merge-and-forget.
- **S5** — drizzle error messages embed query params, so [`middleware/audit.ts:94`](https://github.com/RevealUIStudio/revealui/blob/713d8059/apps/server/src/middleware/audit.ts#L94) logging `cause.message` serializes the event payload (incl. `actor.ip`, `userAgent`) into logs. Pre-existing, not a regression.

## Nits

- The PR description's Part 2 still advertises an `audit_write_failures_total` structured-log counter. That event name no longer exists anywhere in the tree (correctly — it was the thing I asked to be removed). Worth editing the body so it doesn't mislead a future reader.
- `computeCounters()` ([`:126-136`](https://github.com/RevealUIStudio/revealui/blob/713d8059/packages/security/src/audit-write-failures.ts#L126-L136)) rescans up to 500 entries on every audit write, and `window.shift()` is O(n). Trivial in absolute terms; an incremental counter pair would be O(1). Cosmetic.
- The validator's variable lookup is name-based and not scope-aware (self-disclosed in the header). Worst case is a false positive, which fails closed — safe, just occasionally annoying.

---

## Verdict

**APPROVE.** B1 and B2 are closed, and I confirmed it by re-running the exact exercise that broke the previous cut plus a Stage 1 fix-simulation the author did not run. The four tests now pin specific, falsifiable facts about today's behavior, they go red when the harness is broken, and they provably invert when the bug is fixed. That is what Stage 0 was supposed to deliver, and it now does.

Not applying `sec-review:approved` and not merging — the owner applies the label on this recorded verdict. Please action **S4** before this reaches production.

*Re-reviewed by an independent non-author agent (Opus 4.8) at `713d8059e`. Findings re-derived from code; false-green reproduction, characterization-test runs, Stage 1 inversion simulation, and validator bypass probes all executed, not inferred.*

RevealUIStudio:
## Security re-verdict (guardrail 2): APPROVE

Independent non-author re-derivation of the post-review commits (`9be9c4565` + `713d8059e`) against the REQUEST-CHANGES verdict above, by a different session than both the author and the first reviewer. Fresh detached worktree at `713d8059e`; I re-ran the suites and re-derived each finding's fix from the code, not from the response comment.

### Blockers — both verified fixed, empirically

- **B1 (tests proved nothing specific): FIXED, and I reproduced the original false-green probe.** In the fresh worktree with `packages/db` deliberately unbuilt — the first reviewer's exact repro — the rewritten integration test now fails **loudly** (`Test Files 1 failed, Tests 2 failed` on the unresolved import) where the reviewed head reported deceptive `1 expected fail` green. After building, it passes for real: the `23514` + `audit_log_severity_check` pin fires against actual PGlite. All four rewritten tests are plain `it()` with cause-specific `toMatchObject` pins, `toBeDefined()` guards, and a harness-sanity test first. Suites at head, run by me: db integration **2/2**, server pglite **4/4**, security **18/18**, validator **14/14**.
- **B2 (unthrottled per-failure warn): FIXED.** `packages/security/src/audit-write-failures.ts` now contains exactly one log site — the ratio escalation (`logger.error` at `:164`). No per-attempt or per-failure emit remains; the module doc records why.

### Follow-ups

- **S1 (latching alarm): fixed as claimed.** Rolling 500-attempt window; the re-arm test drives spike → recovery → second spike and asserts `logger.error` called **twice** (`audit-write-failures.test.ts:166-185`). Verified green.
- **S3 (matcher bypasses): honestly resolved.** Parens, `<never>` assertion syntax, ternary branches, and the one-hop hoisted-variable bypass are now caught (I read the AST changes — still zero authored regex; the hop bound prevents cyclic-declaration loops). The two remaining bypasses (multi-hop, spread, `as any`) are documented in the function doc + tests instead of claimed, and the PR body's framing is downgraded to "speed bump". Correct posture.
- **The intentionally-red chain-head test is gone** — `713d8059e` converted it to a passing characterization of the defect (asserts `previous_signature` equals the never-persisted rejected write's signature), so nothing red ships to `test`. This resolves what would otherwise have been my blocker: the fleet just paid twice for inherited-red-on-test (the TS2532 and the doc-currency incidents).
- **S2 / S5** remain owner-routed, recorded in the PR body — consistent with the first verdict's non-blocking classification.

### One non-blocking observability nit (no re-review needed)

`lastFailure` is retained independently of the rolling window, so a ratio escalation can name a `lastFailureEventId` whose failure has already aged out of the window that fired the alarm. Cosmetic; fine to leave or fold into Stage 1.

### Disposition

Verdict recorded; the `Security review gate` red is the gate awaiting the label, which I do not apply (agent-authored PR — the owner applies `sec-review:approved` and merges).

PR #1866, opened 2026-07-12
feat(knowledge-graph): GAP-349 P4 electric sync + explorer + drift

RevealUIStudio:
Part of GAP-349.

## A. Electric shape routes + hooks

- `apps/admin/src/app/api/shapes/{kg-nodes,kg-edges,kg-edge-episodes}/route.ts` — copied the `coordination-sessions` template exactly (getSession auth gate, `checkAIFeatureGate`, `prepareElectricUrl`, `runtime = 'nodejs'`, `dynamic = 'force-dynamic'`).
- `repo` query param partitioning: a new `isRepoIdentifier` char-set validator in `apps/admin/src/lib/utils/identifier-validation.ts` (extends the existing `isSyncIdentifier`/`isUuid` module rather than duplicating it — the repo charset needs `.` for repo names like `.jv`, which the existing sync-identifier charset excludes). No authored regex anywhere in the new validation path.
- `kg-edge-episodes` is **not** repo-partitioned — see the escalation below.
- `useKgNodes`/`useKgEdges`/`useKgEdgeEpisodes` + a `useKnowledgeGraph` composite in `packages/sync/src/hooks/useKnowledgeGraph.ts`, exported from the package root. Deliberately **no mutation functions** — Electric sync stays read-only for this graph (tested explicitly).

## B. Read-only graph explorer

- `apps/admin/src/app/(backend)/knowledge-graph/page.tsx`, gated `feature="ai"` (same tier as the other AI-adjacent admin surfaces).
- Repo picker (backed by a tiny new `GET /api/kg/repos` distinct-query route — deliberately *not* an Electric shape, so the picker never has to sync the whole unfiltered fleet graph just to discover repo names), kind filter, name/natural-key search, node detail pane (summary, attributes, current facts with provenance episode ids), and a date-input point-in-time toggle that filters the already-synced edges client-side (`valid_at`/`invalid_at` predicate).
- No writes to `kg_*` from the UI. No graph canvas (list/detail only, per the P4 scope note).

## C. Yjs curation overlay

- `packages/sync/src/collab/kg-view.ts`: per-view `kg-view-<slug>` documents on the **existing** `yjs_documents`/`yjs_document_patches` tables. Annotations, pins, and layout positions all ride the **existing** `set_key` patch type (no new patch type) — see the file's header comment for the exact patch-shape table.
[...]

PR #1868, opened 2026-07-12
fix(knowledge-graph): GAP-349 retrieval-quality batch (FTS + scan determinism + gate widening)

RevealUIStudio:
Part of GAP-349.

## Item 1 — FTS name normalization (the fix that makes search work)

`revkg search "electric proxy"` returned 0 hits against 14k+ real revealui nodes because Postgres's `english` tsvector parser tokenizes hyphenated/path-like names (`electric-proxy.ts`, `apps/admin/src/lib/api/electric-proxy.ts`) as a single opaque token, so word-level `websearch_to_tsquery` queries never matched a `search` tsvector generated from raw `name` + `summary`.

- New `kg_nodes.search_text` column, populated at ingest time via a character-walk normalizer (`packages/knowledge-graph/src/search/normalize-text.ts`, zero regex) that turns separators (`/ - _ .`) into spaces and inserts a space at camelCase boundaries. The original `name` is also folded in unnormalized, so exact single-token queries still hit.
- The generated `search` tsvector is repointed at `search_text` instead of `name`+`summary`.
- Migration `0022_kg_search_text.sql`: `kg_nodes` already carries real rows on Neon (this migration ships **unapplied**, owner applies), so it backfills existing rows in-migration and rebuilds the generated column + its GIN index. **Backfill scope**: the SQL backfill is separator-only (plain `replace()`, no regex) — camelCase backfill is not expressible in plain SQL and completes incrementally the next time each node is re-upserted by a scan (the ingest engine always writes full normalization on every upsert).
- `packages/knowledge-graph/src/db/ddl.ts` (the PGlite-test DDL builder) now emits the CREATE TABLE statement matching `0021` verbatim, followed by the ALTER sequence matching `0022` verbatim, so the drift test (`ddl-migration-sync.test.ts`) validates against both migration files concatenated.
- Tests: PGlite proof that `electric-proxy.ts` is found by "electric proxy" and `getPoolStats` is found by "pool stats"; an exact-token regression test; and a **prove-red** test that builds the OLD (pre-fix) `kg_nodes` shape directly and shows the same query returning 0 hits.

## Item 2 — scan churn investigation + fix (data quality)

Two consecutive fleet re-scans of unchanged trees each produced +5 new edges (revdev) / +5 new nodes (revealui).

**Root cause**: `listDir` (`extractors/shared.ts`) returned raw `readdirSync` order, which is POSIX-unspecified. `workspace.ts`'s `packageKindByName` map resolves a `depends-on` edge's target kind (`'package'` vs `'app'`) by iterating discovered workspace dirs with **last-write-wins** semantics — two dirs in the same `packages/`/`apps/` topDir that happen to declare the same npm package name (a real, if unusual, artifact of a rename where the old dir wasn't removed) could resolve to a different winner kind on a later scan with zero source changes, flipping the derived node id of every `depends-on` edge targeting that name (`deriveNodeId` includes `kind`) and producing false invalidate+insert churn. `ts-project.ts` and `routes.ts` fold the same unsorted listing into their own discovery.

**Unstable keys**: `depends-on` edge ids whose target is a `package`/`app` node resolved via `packageKindByName` (workspace.ts) — specifically when two directories in the same `packages/*` or `apps/*` glob declare the same `package.json` `name`. `deriveNodeId('package', ...)` vs `deriveNodeId('app', ...)` differ by `kind`, so the SAME logical dependency edge gets two different possible ids depending on iteration order.

**Fix**: sort at the single source (`listDir`) so every consumer (workspace/ts-project/routes) inherits a stable order regardless of filesystem/kernel readdir behavior.

**Repro note**: an in-process AND cross-process double-run of every Tier-1 + git extractor against the real `revdev`/`revealui` checkouts (fixed `now`, PGlite-free, never touched Neon) came back byte-identical in this sandbox environment — expected, since `readdirSync` happens to already return alphabetical order for these directories here. This is a genuine POSIX-unspecified-behavior latent defect, not one reproducible on demand on every filesystem/kernel. Verified the fix is load-bearing via a mocked-`readdirSync` unit test that proves red without `.sort()` and green with it.

**Tests added**: `extractors/__tests__/determinism.test.ts` (every Tier-1 + the additive `git` extractor emit an identical node/edge key set across two runs of a realistic fixture; a targeted regression fixture reproducing the duplicate-package-name scenario) and `extractors/__tests__/shared.test.ts` (`listDir` sort unit tests, including the mocked prove-red).

## Item 3 — sensitive-path list widening

Added `api/shapes/` and `api/sync/` to both sensitive-path lists (lockstep, textually identical entries) so future Electric shape/sync routes trip the security gate before the knowledge-graph's own kg-nodes/kg-edges shape routes (P4, #1866) land unflagged:
- `scripts/validate/security-review-gate.cjs` (this PR)
- `.jv` `scripts/security-pr-review-check.js` — pushed separately to `revealui-jv:test` at `ed22b449b` (see below)

This PR's own diff (items 1-2: `packages/` + `packages/db/migrations/`) does not touch `api/` paths, so it does not self-flag — confirmed by running the checker against this branch (see Verification below).

## Peer PR overlap

PR #1866 (P4, open) touches `packages/sync` + `apps/admin` + `revkg.ts`'s drift subcommand. This PR does not touch `revkg.ts`, `packages/sync`, or `apps/admin` — item 1's ingest/search/DDL changes and item 2's extractor changes are in disjoint files.

## Verification

- `pnpm --filter @revealui/knowledge-graph test` — 52/52 pass (includes the 3 new test files).
- `pnpm --filter @revealui/db test` — 1182/1182 pass (5 pre-existing skips).
- `pnpm turbo run typecheck --filter=@revealui/db --filter=@revealui/knowledge-graph` — clean.
- `pnpm biome check` over all touched files — clean.
- `pnpm validate:migrations` — 23 SQL files, 23 journal entries, all checks passed.
- Full pre-push gate (`pnpm gate --phase=1 --changed`, feature-branch scope) — PASS on push.
- `node scripts/validate/security-review-gate.cjs --diff origin/test` on this branch — "no security-sensitive files touched" (confirms the self-flag concern is a non-issue).

## Escalations for Fable

None. All three items resolved within scope; item 2's root cause is a defensive fix for POSIX-unspecified `readdirSync` ordering rather than a reproduced live incident in this sandbox — flagging for awareness in case a future live-fleet re-scan surfaces a DIFFERENT churn pattern this fix doesn't cover (in which case the next investigation should extend the `determinism.test.ts` fixture rather than re-derive the methodology).


RevealUIStudio:
## Fable review (GAP-349 retrieval-quality batch)

Verdict: APPROVE. No blocking findings.

- Migration 0022 verified: idempotent guards throughout, separator-only SQL backfill honestly scoped (camelCase completes via the always-overwrite search_text upsert on the next scan, correctly documented), and drop/recreate is the right mechanism for repointing a generated column. Applying it on ~16k rows is trivial.
- The churn root cause is convincing: unsorted readdirSync order feeding a last-write-wins kind resolution is exactly the class that produces small stable-count false churn, and fixing it at the single listDir source is right. The honest no-live-repro disclosure plus the mocked-order red/green test is the correct standard of proof here.
- The revealui gate-list widening matches the .jv counterpart textually; lockstep held.

Also recording here the review of .jv commit ed22b449b (the fleet checker widening), since it has no PR of its own: content APPROVED, two entries plus a lockstep comment, no behavioral risk. Process note for the record: .jv scripts/** changes fall under the recorded-verdict convention and should ride a PR next time; the direct push was my briefing gap, not the implementer's improvisation, and this comment serves as the post-hoc verdict.

Merge remains an owner action; migration 0022 ships unapplied.

PR #1872, opened 2026-07-12
fix(server): converge hosted entitlements on trial checkout

RevealUIStudio:
## Converge hosted entitlements on trial checkout

A hosted customer who completed a paid trial checkout received zero entitlements: the license row minted correctly, but the `account_entitlements` row was never written, so every read resolved the buyer to `free`. Reproduced end-to-end on production 2026-07-12.

Four compounding defects in the webhook entitlement write path (`syncHostedSubscriptionState`), fixed together because each one masks the next:

1. **The `account_entitlements` status CHECK constraint omitted `'trialing'`** while the `customer.subscription.created` handler writes the raw subscription status into the entitlement upsert — every trial entitlement INSERT threw a check violation. Migration `0023` widens the constraint; `account_subscriptions` already allowed it, and the read path (`isHealthyStatus`) already branches on it.
2. **The staleness guard compared local wall-clock `updated_at` against Stripe `event.created`**, so within a webhook burst the second event was always judged stale — and a stale subscription UPDATE early-returned the whole function, silently skipping the entitlement upsert on `checkout.session.completed` and on every Stripe retry. Replaced with a per-table event-to-event guard on a new nullable `last_event_at` column: NULL passes, a strictly-older event skips that row only, and each table guards independently. The out-of-order protection this guard exists for (a delayed `.updated(active)` must not resurrect state written by a newer `.deleted`) is preserved and tested.
3. The non-transactional subscription-first ordering left partial state that poisoned every retry; the per-table guard makes retries convergent.
4. **An unresolvable tier defaulted to `'free'`** — a silent downgrade. Now the entitlement write is skipped, an ERROR is logged, and an idempotent `unreconciled_webhooks` row (`entitlement.unresolved-tier`) is recorded for the drainer instead.

### Tests (each proven red against the original code first)

New PGlite integration suite (`entitlement-convergence.pglite.test.ts`) applies the REAL migrations and drives the REAL webhook route:
- trialing `subscription.created` → `checkout.session.completed` converges to a pro/active entitlement (red on base: 500 check-violation + stale-skip)
- both event orders + redeliveries converge to the same terminal state (red on base: 500)
- guard semantics: older event skips per-row, newer event applies even when the sibling row was skipped, NULL always applies (red on base: check violation)

Full server suite 157 files / 2892 passed; `@revealui/db` 55 files / 1184 passed; typecheck clean.

### Notes for review

- Design: `.jv` `docs/gap-specs/GAP-356-hosted-entitlement-resolution-design.md` (merged, jv#481).
- Residual accepted: Stripe `event.created` has 1-second granularity and equal timestamps skip; both burst events carry the same tier, so the terminal entitlement stays healthy either way and converges on the next later-stamped event.
- Residual accepted: after a first unresolvable-tier event stamps the subscription row `planId='free'`, a second unresolvable event would resolve `'free'` from it; never access-granting, and the first skip already recorded the alert row.
- Flagged for follow-up (not in this PR): the stale "callers will default to pro" comment on `resolveOptionalTier`; the generated-type layer not picking up `last_event_at` (nothing consumes it there).

Security-sensitive surface (`routes/webhooks.ts`): requires the recorded guardrail-2 verdict before merge. This PR does not close GAP-356 — closure needs the live re-smoke (design §5 F6) after deploy, plus the PR-2 reconciler.


RevealUIStudio:
## Guardrail-2 verdict (Fable, independent adversarial pass): **APPROVE**

Reviewer is the GAP-356 design author and coordinator; the implementation was produced by a separate Opus implementer session and reviewed here against the merged design (jv#481) and the live code. The label is the owner's to apply; this comment records the verdict.

**What I verified independently (not taken from the implementer's report):**

1. **Diff read in full.** The rework is confined to `syncHostedSubscriptionState` + the migration/schema pair; no read-path or gate code is touched. `requireFeature` / `entitlementMiddleware` semantics unchanged.
2. **The guard cannot resurrect revoked state.** The documented WH-3 resurrection vector (delayed `.updated(active)` after `.deleted`) stays closed: both tables' rows carry the newer event's `last_event_at`, so the stale event's `lt()` fails on both. Proven by test 3's asymmetric-clock construction and permutation B.
3. **INSERT paths are unguarded by design** and both writes report independently — the D2 shared-fate early return is gone (the `return` at the old stale-skip is deleted; I checked the control flow, not just the report).
4. **The F3 alert is genuinely idempotent**: `unreconciled_webhooks.event_id` is the PRIMARY KEY (`webhook-reconciliation.ts:20`), so `onConflictDoNothing` on the synthetic id collides correctly across retries.
5. **Migration is idempotent and validator-clean**: `DROP CONSTRAINT IF EXISTS` + `duplicate_object`-guarded re-add + `ADD COLUMN IF NOT EXISTS`, journal idx 23, `_custom.json` entry per the 0014/0015 precedent; `migration-journal` validator passes (it failed the first push and was fixed in `ffbd5bb85`).
6. **Tests re-run by the reviewer**: 3/3 pass; red-proofs were produced against base code per the implementer transcript (500 check-violation + stale-skip on the original constraint/guard).
7. **Downgrade-cap gating** (`entitlementApplied`) is strictly safer than the original: a stale-skipped downgrade event can no longer cap resources.

**Residuals, accepted with rationale (also in the PR body):**
- Equal-second Stripe timestamps skip (1s `event.created` granularity). Both burst events carry the same tier, so the terminal entitlement is healthy either way; a later-stamped event converges status. Not fixable from `event.created` alone.
- `coerceHostedTier('free')` + the subscription row's retained `'free'` planId fallback means F3's no-free-from-ignorance property holds only until an unresolvable event has stamped the subscription row; a second unresolvable event would write a free entitlement. Never access-granting, alert row already recorded. Consider tightening in PR-2.
- rowCount-shape cast on the drizzle update result matches the pre-existing pattern; tests assert row state, not driver counts.

**Not cleared by this verdict:** GAP-356 closure. That needs the PR-2 reconciler + the owner-run live re-smoke (design §5 F6) after deploy.

PR #1879, opened 2026-07-13
fix(server): harden the entitlement reconciler (GAP-356 PR-2 review)

RevealUIStudio:
Addresses the **REQUEST-CHANGES** verdict on #1873 ([verdict comment](https://github.com/RevealUIStudio/revealui/pull/1873#issuecomment-4953543688)).

#1873 shipped a cron that **writes paid entitlements in production**. An independent adversarial review found four defects that each either over-grant access, disable the alert channel the cron exists to provide, or starve the cron that revokes access. **None of it reached production** — #1873 is on `test` only, and the promote was held. But `test` cannot promote until this lands, because any promote carries that cron to `main`.

I wrote the code under review. Everything below is a fix to my own defect.

## B1 — Over-grant: no freshness bound (the worst one)

The healer never read `currentPeriodEnd`. `HEALTHY_SUBSCRIPTION_STATUSES` was a bare string check on `status`, and `isHealthyStatus` grants full tier for `trialing` with no expiry comparison.

So: a trial ends → Stripe's expiry webhook **fails** (the exact failure class this cron backstops) → the local row stays frozen at `trialing`/`pro` → **every tick, my cron re-granted Pro. Forever. Free.** I created a path that converts a stale local row into granted access with no Stripe event behind it. That capability did not exist before.

A subscription is now an entitlement source only while `status` is healthy **AND** `current_period_end` is null or in the future. Provenance is not currency: a signature-verified webhook wrote the row, but that says nothing about whether it's still true. A stale row alerts and is **never** healed from.

## B3 — The alert channel closed itself (proven)

`drain-unreconciled` selects **every** unresolved row and hands `event_id` to `stripe.events.retrieve()`. A synthetic id 404s → it classifies `event-missing` → **it marks the row resolved.** The drainer runs every 15 minutes (the GAP-142 external scheduler); this cron runs daily (Vercel Hobby allows one cron). So the alert was auto-closed within 15 minutes — and because `recordDrift` keyed idempotency on `event_id` alone, **that account could never raise a drift alert again.**

The original idempotency test passed only because no drainer ran between its two ticks. Production has one.

Fixed on both sides:
- `lib/synthetic-events.ts` — the drainer now skips cron-authored ids. This also protects the **two pre-existing callers** with the same latent bug (`cron-sub-orphan:`, `entitlement-unresolved-tier:`), which were silently being auto-resolved too.
- `recordDrift` scopes its check to **unresolved** rows, and re-opens a resolved one rather than losing the alert to a PK conflict.
- A successful heal now **closes its own alert**, so the row doesn't linger unresolved forever.

## B2 / B4 — The terminal rail was backwards

I added a rail refusing to heal `revoked`/`expired`/`canceled` entitlements, and flagged it for the verdict. The review traced every reachable status and showed it is **unreachable in practice** — `revoked` is never written to entitlements (all 14 sites target `licenses`), and `expired`/`canceled` always accompany a non-healthy subscription, which bails earlier.

The one case it *did* catch: a customer who **resubscribed** after expiry and whose entitlement write failed. My rail denied exactly the person it should have healed. Meanwhile the real resurrection vector — a manually **deleted** row — sailed straight through, because the check required the row to exist.

Removed, and replaced with an explicit, documented authority model:

> **Stripe is the cut-off mechanism, not the database.** Cancel the subscription and the local row stops being healthy, so the cron stops healing. Hand-editing `account_entitlements` (tier→free, or deleting the row) is **not** a supported revocation — it will be healed back within a day, *correctly*, because a fresh healthy subscription says the customer is paying.

With B1's freshness bound in place, that model is safe: a dead trial no longer looks healthy.

## B5 — Cost, and the dangerous ordering

Vercel Hobby allows **one cron/day**, and `dispatch.ts` runs all ~13 jobs sequentially inside a single 30s function.

- I had 4 queries per account under a 500-row cap. Past 500 accounts the remainder was **silently never reconciled** — the cap was a scan limit masquerading as a batch. A keyset cursor doesn't fix that under a daily tick (account #501 waits a day, #1001 two days), so the job is now **set-based**: one scan query for all accounts, with the license probe folded in as a correlated `EXISTS`. Writes happen only for the (small) drift set.
- I registered it **before `sweep-grace-periods`** — the cron that **expires** access for non-payers. If my job ate the shared budget, revocation wouldn't run that day. It now runs **after** the sweep. A healer that grants access must never starve the revoker. (Running after the sweep also means it sees the sweep's own expirations and won't re-grant what was just expired.)

## Also fixed

- **Mode scoping** on every read and write. `account_id` is the entitlement PK, so a heal driven by a **test-mode** subscription would have flipped the account's single row to `mode: 'test'` — making it invisible to the live consumer. Textbook "paying customer gated as free."
- **`last_event_at` is no longer nulled on UPDATE.** NULL means *any* event wins — including a **stale** one replayed by the drainer with its original `created`. That re-opened the WH-3 out-of-order window PR-1 just closed. NULL on INSERT only.
- **`TIER_RANK` de-duplicated** — now reuses the single map in `downgrade-cap.ts` via a new `isUpgrade` beside `isDowngrade`. Shipping a third copy in a PR whose thesis is *"duplicate mappings are the bug"* was self-defeating.
- Ops is paged only for **unhealed** drift (a fully-healed run is a success, not an incident), and the response returns exceptions rather than every account id.

## Tests: 5 → 13, and the two critical ones proven red

Each new rail has a test, including the four the review found **untested**:

| Test | Guards |
|---|---|
| stale subscription is never healed from | **B1** |
| re-alerts after a resolved drift row | **B3** |
| resubscribed customer over a terminal entitlement **is** healed | **B4** |
| test-mode row never touches a live entitlement | mode scoping |
| `last_event_at` preserved on UPDATE | WH-3 window |
| unresolvable tier → alert, no write | F3 rail |
| `RECONCILE_ENTITLEMENTS_HEAL=false` → alert, no heal | kill switch |
| healed row carries the right features/limits | shared builder |

The two blocker fixes were **proven red** by mutation: removing the freshness bound fails the stale-subscription test (the expired trial gets free Pro); removing the unresolved-scope fails the re-alert test.

**433/433 pass** across cron + webhook + lib suites. Typecheck and Biome clean.

---

**Guardrail 2:** touches cron + entitlement write paths. **Requesting an independent adversarial verdict — I am the author of both the original defect and this fix, and I am not clearing my own work.** Not self-merging, not applying gate labels.

### DIFF
--- a/apps/server/src/lib/downgrade-cap.ts
+++ b/apps/server/src/lib/downgrade-cap.ts
@@ -34,6 +34,19 @@ export function isDowngrade(oldTier: LicenseTierId, newTier: LicenseTierId): boo
   return TIER_RANK[newTier] < TIER_RANK[oldTier];
 }
 
+/**
+ * Returns true if newTier has strictly higher rank than oldTier. Equal tiers are
+ * NOT an upgrade.
+ *
+ * The entitlement reconciler's monotonic-upward heal guard uses this. It lives
+ * here, beside `isDowngrade` and sharing the one `TIER_RANK`, rather than in a
+ * second copy of the map — a duplicated tier ordering is exactly the drift class
+ * GAP-356 already is.
+ */
+export function isUpgrade(oldTier: LicenseTierId, newTier: LicenseTierId): boolean {
+  return TIER_RANK[newTier] > TIER_RANK[oldTier];
+}
+
 // =============================================================================
 // Capping Result
 // =============================================================================
--- a/apps/server/src/lib/hosted-entitlement.ts
+++ b/apps/server/src/lib/hosted-entitlement.ts
@@ -36,29 +36,11 @@ const KNOWN_FEATURE_KEYS = new Set<string>([
   'analytics',
 ] satisfies (keyof FeatureFlags)[]);
 
-/**
- * Tier ordering, low to high. Used by the reconciler's monotonic-upward heal
- * guard: a heal may raise a tier but must never lower one, because a downgrade
- * is a real business decision that belongs to the (signature-verified) webhook
- * path, not to a cron inferring state from local rows.
- */
-const TIER_RANK: Record<HostedTier, number> = {
-  free: 0,
-  pro: 1,
-  max: 2,
-  enterprise: 3,
-};
-
 export function coerceHostedTier(value: string | null | undefined): HostedTier | undefined {
   if (value === 'free' || value === 'pro' || value === 'max' || value === 'enterprise') {
     return value;
   }
   return undefined;
-}
-
-/** True when `next` is strictly higher than `current`. Equal tiers are NOT an upgrade. */
-export function isTierUpgrade(current: HostedTier, next: HostedTier): boolean {
-  return TIER_RANK[next] > TIER_RANK[current];
 }
 
 export function toFeatureRecord(features: object | null | undefined): Record<string, boolean> {
@@ -82,10 +64,16 @@ export function toFeatureRecord(features: object | null | undefined): Record<str
 /**
  * The `account_entitlements` column values for a tier + status.
  *
- * `lastEventAt` is the event-to-event staleness cursor (F2). The reconciler
- * passes `null` deliberately: a healed row carries no event provenance, so the
- * very next webhook — whatever its `event.created` — must win over it. A healed
- * row must never be able to make a real event look stale.
+ * `lastEventAt` is the event-to-event staleness cursor (F2).
+ *
+ * The webhook path passes `event.created`. The reconciler passes `null` **only
+ * when INSERTing** a brand-new row, so the next webhook wins over a synthesized
+ * one. It must NOT null the cursor on an UPDATE: the guard is
+ * `last_event_at IS NULL OR last_event_at < event.created`, so NULL lets *any*
+ * event win — including a stale one. `drain-unreconciled` replays old events
+ * with their original `created`, so nulling an existing cursor would re-open the
+ * out-of-order window PR-1 closed. On an update the caller preserves the
+ * existing cursor.
  */
 export function buildHostedEntitlementValues(params: {
   tier: HostedTier;
--- a/apps/server/src/lib/synthetic-events.ts
+++ b/apps/server/src/lib/synthetic-events.ts
@@ -1,0 +1,49 @@
+/**
+ * Synthetic `unreconciled_webhooks.event_id` prefixes.
+ *
+ * Several crons record ops-visible alerts by writing a row into
+ * `unreconciled_webhooks` with a MADE-UP `event_id` (so re-runs collide and the
+ * alert is idempotent). Those ids are not Stripe events and can never be
+ * replayed.
+ *
+ * This matters because `drain-unreconciled` selects EVERY unresolved row and
+ * hands `event_id` to `stripe.events.retrieve()`. Stripe 404s on a synthetic id,
+ * the drainer classifies it `event-missing`, and **marks the row resolved**.
+ *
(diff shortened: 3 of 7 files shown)


RevealUIStudio:
## Guardrail-2 verdict (Fable, independent adversarial pass): **APPROVE**

Reviewer authored neither #1873 nor this fix (this session's GAP-356 work was the design, jv#481, and the PR-1 verdict on #1872). Verdict independently derived from the diff, not the PR body. The four attack questions from the checkpoint, each verified in code:

**(a) B1 — does the freshness bound close the over-grant? YES.** The bound is enforced at BOTH decision points: source selection (`subscriptionUsable = statusHealthy AND periodFresh`) and the heal path (a `subscriptionStale` row takes the `drift-alert-only-stale-subscription` branch before any write is reachable). The frozen-trial scenario now alerts with an explicit stale message and never heals. Tested (`does NOT heal from a STALE subscription`).

**(b) B3 — is the alert channel actually re-opened? YES, on all three sides.** The drainer skips every id matching `SYNTHETIC_EVENT_ID_PREFIXES` before the Stripe retrieve (verified the registry against the real writers: `cron-orphan:` in reconcile-customers, `cron-sub-orphan:` in reconcile-stripe-subscriptions, `entitlement-unresolved-tier:` matching PR-1's F3 writer). `recordDrift` scopes its suppress check to `resolvedAt IS NULL` and re-opens a resolved row via `onConflictDoUpdate` instead of losing the alert to the PK. A successful heal closes its own row, so the suppress-while-open / re-alert-after-resolve cycle is coherent. All three transitions are tested.

**(c) B2/B4 — is dropping the terminal rail for "Stripe is the cut-off" sound? YES, given B1.** The mirror failure (cancellation webhook lost) is covered: while the paid period is unexpired the customer legitimately keeps access; once `current_period_end` passes, the freshness bound stops the heal. The resubscribed-customer case the old rail wrongly denied is now healed and tested. The authority model is documented in the file header where the next editor will see it.

**(d) B5 — budget and ordering? YES.** One set-based scan (mode-scoped joins, license probe as correlated `EXISTS`), writes only for the drift set, an 8s in-loop budget guard that degrades to alert-only (alerts land before the guard, so nothing is silent), and registration AFTER `sweep-grace-periods` in dispatch.ts with the revoke-before-heal rationale in place — and the sweep's own expirations fail the healthy check, so just-expired access cannot be re-granted the same tick.

Also verified: mode scoping on every join and write (the test-mode-flips-live-row hazard is closed and tested); `last_event_at` preserved on UPDATE heals (WH-3 window stays closed, tested); `TIER_RANK` deduplicated via `isUpgrade`.

**Residuals, accepted (none blocking):**
1. `current_period_end IS NULL` counts as fresh. A healthy row that never got a period end would heal indefinitely. Narrow: the webhook writers populate it (checkout falls back to trial end), and treating NULL as stale would break legitimate rows. Acceptable; worth remembering if a NULL-period row ever shows up in a drift alert.
2. Multiple subscription rows per account fan the scan out (the header comment says fan-out is membership-only — slightly inaccurate). Worst case is a redundant idempotent heal from scan-stale data; only healthy AND fresh rows can heal, monotonic upward, so no access impact.
3. The license `EXISTS` probe checks `status='active'` but not `expires_at` — an expired-by-date license still marked active keeps an account in alert-only drift (licenses never heal, so no grant). Alert-noise risk only.

**Conditions:** merge on green CI (checks were still running at review time). The gate label is the owner's to apply. This verdict clears #1879 only — GAP-356 still closes on F6 (a real live purchase working), not on this merge.

PR #1896, opened 2026-07-16
feat(ai): GAP-360 PR-2 — per-account key resolver + dispatch wiring + per-tenant worker gate

RevealUIStudio:
**Part 2/3 of GAP-360** ([design spec](https://github.com/RevealUIStudio/revealui-jv/blob/main/docs/gap-specs/GAP-360-hosted-agent-execution-design.md)). The security-sensitive core: per-account key decryption at dispatch time. **Needs a guardrail-2 verdict + `sec-review:approved` before merge** (not self-applied).

Builds on PR-1 (`createLLMClientForUser`, widened provider surface, `hostedViable`, boot warning) already merged on this branch.

### §5.2 — the single resolver
New `resolveLLMClientForRequest(userId, db, ctx)` in [`packages/ai/src/llm/resolve.ts`](packages/ai/src/llm/resolve.ts). One home, consumed by every dispatch site — no site-local copies (extend-before-create). Order:

1. **Per-user BYOK** — `createLLMClientForUser(userId, db, audit, { hostedViableOnly })` (preferred).
2. **Site inference config** — `workspace_inference_configs` (active; on hosted, hostedViable providers only).
3. **Deployment env** — `createLLMClientFromEnv` — **SELF-HOSTED ONLY**. Forbidden on hosted (the exact silent-localhost defect class).
4. **Hosted + nothing above** → throw typed `LLMNotConfiguredError` → **HTTP 409** with a machine-readable body naming `/settings/api-keys` (not 402, not 500).

Hosted detection reuses the existing deployment-mode signal (`detectDeploymentMode` at [validate-startup.ts:133](apps/server/src/lib/validate-startup.ts#L133), keyed on `REVEALUI_LICENSE_PRIVATE_KEY` presence; admin mirrors it per [instrumentation.ts:49](apps/admin/src/instrumentation.ts#L49)). No `VERCEL` sniffing.

### §5.4 — dispatch-site coverage (all seven, one resolver)

| Site | Wiring |
|---|---|
| a2a `tasks/send` (a2a.ts:1173) | resolver; userId from `c.get('user')`; deleted the `:83-84` "no per-request key" comment; 409 as a JSON-RPC error |
| agent-stream (agent-stream.ts:167) | resolver; userId from session; 409. Free-tier local path untouched |
| ticket dispatch (agent-dispatcher.ts:45) | `buildDispatcher` now resolves via the resolver; sync callers pass session `user.id` |
| durable worker (jobs/agent-dispatch.ts) | resolver + §5.6 per-tenant gate; userId from the **task's owning account** (`ticket.reporterId`), not the payload |
| admin chat (app/api/chat/route.ts:310) | resolver; userId from `authSession.user.id`; 409; resolved client reused for the RAG embedding call |
| RAG embeddings (embeddings/index.ts:68) | `generateEmbedding` gains an optional pre-resolved `client`; the authenticated admin-chat path passes it (see deviations for rag-index) |
| CLI / harness adapter | **env, unchanged** — local/self-hosted context by definition |

`agent-tasks.ts` also now sets `reporterId: user.id` on ticket creation, so the worker has an authenticated owner to derive from (without it, hosted dispatch would fail closed on every ticket).

### §5.6 — per-tenant worker gate
Replaced the singleton `isFeatureEnabled('ai')` in the durable worker with per-account entitlement resolution via the GAP-356 canonical path (`accountEntitlements` scoped by `getConfiguredStripeMode()`, grace-expiry fail-safe mirrored from `entitlementMiddleware`) in new `apps/server/src/lib/account-entitlement.ts`. The "pre-existing gap" comment is retired. On self-hosted (or a flag rollback) the singleton check remains, byte-unchanged.

## §6 security invariants → tests

| # | Invariant | Implementation | Test |
|---|---|---|---|
| 6.1 | userId from authenticated context only, never request body/params | resolver takes userId as an argument; worker derives it from `ticket.reporterId` | `resolve.test.ts` "passes the exact userId argument", "skips … when userId is null"; **`agent-dispatch.test.ts` "derives userId from ticket.reporterId and ignores the payload userId"** |
| 6.2 | plaintext lifetime = request scope; no key cache/log/serialized row | client constructed per request; no cache; nothing logs/returns a key | resolver constructs per call (no module state) |
| 6.3 | decryption server-side only via `decryptApiKey` | site path calls `decryptApiKey`; BYOK path delegates to PR-1's `createLLMClientForUser` | `resolve.test.ts` branch-2 (decrypt invoked) |
| 6.4 | audit every access (`byok:key:accessed`) | `DrizzleAuditStore` wired at server sites; admin in-memory per **GAP-338** (recorded limitation) | audit sink threaded through `ctx.auditStore` (fires inside `createLLMClientForUser`) |
| 6.5 | fail-closed: unknown provider / failed decrypt / non-hostedViable → 409, never env on hosted | try/catch → `LLMNotConfiguredError` on hosted; `hostedViableOnly` filter before decrypt | `resolve.test.ts` "hosted … never calls the env factory", "fails closed when the BYOK factory throws", "skips a non-hostedViable site config" |
| 6.6 | no new secret surface (validate-startup stays AI-keyless) | no platform env keys added | boundary validation + no validate-startup change |

Full suite: **`@revealui/ai` resolver 17/17**, **worker 3/3**, plus updated a2a (47/47), agent-tasks, agent-stream(+success/elicit) — 64/64 touched server tests, 125/125 ai/llm tests, 74/74 a2a+rag-index, admin chat 7/7.

## Red-proof (prove-red first)

Both highest-risk behaviors proven to fail without the fix:

- **(b) userId cannot come from the request body** — `git checkout origin/test -- apps/server/src/jobs/agent-dispatch.ts` (restore pre-change worker), ran the worker test → **2 failed** ("derives userId from ticket.reporterId…", "fails closed when the owning account lacks the ai feature"); restored → **3/3 green**.
- **(a) hosted never falls through to env** — `resolve.ts` is a new file, so prove-red by removing the step-3 hosted guard (`if (!hosted) return env`) → **3 failed** ("branch 4 throws", "hosted … never calls env", "skips non-hostedViable site config"); restored → **17/17 green**. (git-checkout method for the modified worker file; guard-removal for the new resolver file since there is no pre-change version to restore.)

## §5.5 admin run-surface routing — **DEFERRED**

Deferred to its own PR. The `next.config.mjs` rewrite is authorable, but agent-stream is an **SSE** surface and streaming-through-rewrite is flagged as risky in the spec itself; it needs a live hosted SSE test this build environment cannot run. Per the "defer if uncertain / don't balloon the security PR" guidance, it is separated so this PR stays focused on the key-resolution core. `next.config.mjs` currently has no `rewrites()` block; the change is self-contained and independent of this PR.

## Feature flag — `HOSTED_BYOK_DISPATCH` (§7)

One-release rollback lever. Default **ON for hosted, absent/off for self-hosted**. When off, every site returns `createLLMClientFromEnv()` directly — self-hosted env-first behavior is **byte-unchanged**. Explicit `true/1/on/yes` or `false/0/off/no` override; unrecognized values fall back to the deployment default. The worker gate mirrors the same flag (inlined — apps/server may not statically import the Pro `@revealui/ai`; boundary-enforced).

## Spec deviations (with justification)

- **rag-index background indexing** (rag-index.ts:160,320) still uses the env embedding path. The embeddings *module* fix (optional `client`) ships and the user-facing admin-chat RAG path is wired; rag-index is background indexing (not agent execution), and per-account embedding needs an embedding-capable provider (anthropic BYOK cannot embed) — a model-selection concern outside this PR's security core. It can adopt the same `client` param in a follow-up with no further module change.
- **§6.4 admin audit sink** is in-memory until **GAP-338** closes (recorded limitation, not a blocker). Server sites persist via `DrizzleAuditStore`.

## Changeset
`@revealui/ai`: **minor**. `server`/`admin` are in the changeset ignore list (`.changeset/config.json`) — no changeset entries, per config.

## Security checker (verbatim)
```
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: apps/server/src/routes/a2a.ts, apps/server/src/routes/agent-stream.ts, apps/server/src/routes/agent-tasks.ts
```
(exit 1 = HOLD, as expected.)

## Gate tails
- Pre-push gate: **Result: PASS** (migration journal, raw-SQL, empty-catch, as-never, stripe-client, pricing-lockstep, client-bundle-safety, security-audit, coverage — all ✓; Phase 1 Biome/typecheck/build ✓).
- `pnpm --filter @revealui/ai typecheck` → **0 errors**; `server` typecheck → **0**; `admin` typecheck → **0**.
- Boundary validation → **passed** (no static Fair Source imports in apps/scripts).
- No authored regex; TypeScript strict; explicit return types on exports.

**Part 2/3 of GAP-360; needs a guardrail-2 verdict before merge.**

### DIFF
--- a/.changeset/gap-360-pr2-resolver.md
+++ b/.changeset/gap-360-pr2-resolver.md
@@ -9,4 +9,14 @@ On hosted, an unconfigured account fails closed with a typed `LLMNotConfiguredEr
 (HTTP 409 naming `/settings/api-keys`) instead of silently falling through to the
 localhost default. Gated behind `HOSTED_BYOK_DISPATCH` so self-hosted env-first
 behavior is byte-unchanged. `createLLMClientForUser` gains a `hostedViableOnly`
-option; `generateEmbedding` accepts an optional pre-resolved client.
+option; `generateEmbedding` accepts an optional pre-resolved client. Emits a
+one-time operator warning when the hosted BYOK dispatch flag is explicitly
+disabled (every account shares the deployment env client while the lever is
+pulled).
+
+Security fix (guardrail-2): the durable worker now derives dispatch identity
+from the authenticated dispatcher captured server-side at enqueue time, never
+from `ticket.reporterId` — that column is client-writable via the general
+tickets API with no ownership check, so reading it for key resolution would
+let an attacker who owns a board decrypt a victim's BYOK key by planting the
+victim's user id in `reporterId` before dispatching.
--- a/.changeset/gap-360-pr2-resolver.md
+++ b/.changeset/gap-360-pr2-resolver.md
@@ -1,0 +1,12 @@
+---
+"@revealui/ai": minor
+---
+
+Add the per-request LLM client resolver (GAP-360 PR-2). `resolveLLMClientForRequest`
+is the single home for key resolution at every agent-dispatch site: per-user BYOK
+first, then site-level inference config, then deployment env on self-hosted only.
+On hosted, an unconfigured account fails closed with a typed `LLMNotConfiguredError`
+(HTTP 409 naming `/settings/api-keys`) instead of silently falling through to the
+localhost default. Gated behind `HOSTED_BYOK_DISPATCH` so self-hosted env-first
+behavior is byte-unchanged. `createLLMClientForUser` gains a `hostedViableOnly`
+option; `generateEmbedding` accepts an optional pre-resolved client.
--- a/apps/admin/src/__tests__/integration/api/chat.test.ts
+++ b/apps/admin/src/__tests__/integration/api/chat.test.ts
@@ -60,17 +60,29 @@ vi.mock('@revealui/ai/embeddings', () => ({
   }),
 }));
 
-vi.mock('@revealui/ai/llm/server', () => ({
-  createLLMClientFromEnv: vi.fn(() => ({
-    chat: vi.fn().mockResolvedValue({
-      content: 'I can help you with that!',
-      toolCalls: [],
-    }),
-    getResponseCacheStats: vi.fn().mockReturnValue(undefined),
-    getSemanticCacheStats: vi.fn().mockReturnValue(undefined),
-  })),
-  createLLMClientForUser: vi.fn().mockResolvedValue(null),
-}));
+const makeChatClient = () => ({
+  chat: vi.fn().mockResolvedValue({
+    content: 'I can help you with that!',
+    toolCalls: [],
+  }),
+  embed: vi
+    .fn()
+    .mockResolvedValue({ vector: new Array(1536).fill(0), model: 'test', dimension: 1536 }),
+  getResponseCacheStats: vi.fn().mockReturnValue(undefined),
+  getSemanticCacheStats: vi.fn().mockReturnValue(undefined),
+});
+
+vi.mock('@revealui/ai/llm/server', () => {
+  const createLLMClientFromEnv = vi.fn(() => makeChatClient());
+  return {
+    createLLMClientFromEnv,
+    createLLMClientForUser: vi.fn().mockResolvedValue(null),
+    // GAP-360: the admin chat route resolves the per-account client via the
+    // resolver. It delegates to createLLMClientFromEnv so a test that overrides
+    // createLLMClientFromEnv drives both the chat and embedding paths.
+    resolveLLMClientForRequest: vi.fn(async () => createLLMClientFromEnv()),
+  };
+});
 
 vi.mock('@revealui/ai/memory/vector', () => ({
   VectorMemoryService: vi.fn().mockImplementation(() => ({
--- a/apps/admin/src/app/api/chat/route.ts
+++ b/apps/admin/src/app/api/chat/route.ts
@@ -1,6 +1,7 @@
 import { getSession } from '@revealui/auth/server';
 import { ChatRequestContract } from '@revealui/contracts';
 import { apiClient } from '@revealui/core/admin/utils/apiClient';
+import { getClient } from '@revealui/db';
 import { logger } from '@revealui/utils/logger';
 import type { NextRequest } from 'next/server';
 import { checkAIFeatureGate } from '@/lib/middleware/ai-feature-gate';
@@ -56,6 +57,9 @@ async function loadChatAIDeps() {
   ]);
   const [embeddingsMod, llmServerMod, vectorMod, cmsMod, registryMod] = results;
 
+  // GAP-360: the per-request resolver lives on llm/server.
+  const resolveLLMClientForRequest = llmServerMod?.resolveLLMClientForRequest;
+
   // Log which specific modules failed so operators can diagnose missing Pro deps
   const failed = moduleNames.filter((_, i) => results[i] === null);
   if (failed.length > 0) {
@@ -69,6 +73,7 @@ async function loadChatAIDeps() {
   return {
     generateEmbedding: embeddingsMod.generateEmbedding,
     createLLMClientFromEnv: llmServerMod.createLLMClientFromEnv,
+    resolveLLMClientForRequest,
     VectorMemoryService: vectorMod.VectorMemoryService,
     createAdminTools: cmsMod.createAdminTools,
     ToolRegistry: registryMod.ToolRegistry,
@@ -300,21 +305,49 @@ export async function POST(request: NextRequest) {
         | undefined;
     }
 
-    // Create LLM client from environment-configured open models
-    let llmClient: ChatLLMClient;
+    // Resolve the LLM client via the GAP-360 resolver: per-account BYOK on
+    // hosted, env on self-hosted. userId is the authenticated session user
+    // (§6.1). Hosted detection mirrors the deployment signal used elsewhere
+    // (REVEALUI_LICENSE_PRIVATE_KEY presence, matching instrumentation.ts).
+    // NOTE (§6.4 / GAP-338): the admin process has no durable audit sink yet,
+    // so no auditStore is wired — byok:key:accessed does not persist here until
+    // GAP-338 closes. Recorded limitation, not a Phase-1 blocker.
+    const isHosted = !!process.env.REVEALUI_LICENSE_PRIVATE_KEY;
+    let resolvedClient: ChatLLMClient;
     try {
-      logger.info('Creating LLM client from env', {
-        provider: process.env.LLM_PROVIDER,
-        model: process.env.LLM_MODEL,
-      });
-      llmClient = aiDeps.createLLMClientFromEnv();
-    } catch (_err) {
+      if (aiDeps.resolveLLMClientForRequest) {
+        resolvedClient = (await aiDeps.resolveLLMClientForRequest(
+          authSession.user.id,
+          getClient(),
+          {
+            isHosted,
+          },
+        )) as unknown as ChatLLMClient;
+      } else {
+        resolvedClient = aiDeps.createLLMClientFromEnv();
+      }
+    } catch (err) {
+      const code = (err as { code?: unknown } | null)?.code;
+      if (code === 'LLM_NOT_CONFIGURED') {
+        const e = err as { message?: unknown; settingsPath?: unknown };
+        return new Response(
+          JSON.stringify({
+            success: false,
+            error: typeof e.message === 'string' ? e.message : 'No LLM provider is configured.',
+            code: 'LLM_NOT_CONFIGURED',
+            settingsPath:
+              typeof e.settingsPath === 'string' ? e.settingsPath : '/settings/api-keys',
+          }),
+          { status: 409, headers: { 'Content-Type': 'application/json' } },
+        );
+      }
       return createApplicationErrorResponse(
         'LLM provider not configured',
         'LLM_NOT_CONFIGURED',
         503,
       );
     }
+    const llmClient: ChatLLMClient = resolvedClient;
 
     logger.info('LLM client ready', { userId: authSession.user.id });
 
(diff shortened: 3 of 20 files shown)


RevealUIStudio:
## Guardrail-2 verdict: REQUEST-CHANGES (1 blocker)

Independent adversarial review (fresh reviewer, refute-prompted) + author's own firsthand confirmation at file:line. The resolver core is well-built — attacks on fail-closed (§6.5), plaintext lifetime (§6.2), self-hosted-unchanged flag (§6.4), the GAP-356-faithful entitlement gate (§5.6 logic), and the realm-safe error mapper (§6.6) all HOLD. **One blocker on the tenancy identity (§6.1) must be fixed before merge.**

### BLOCKER — cross-account BYOK key decryption via client-writable `reporterId`
The durable worker derives the key-decryption identity from a mutable ticket column:
- `apps/server/src/jobs/agent-dispatch.ts:105` — `const ownerUserId = ticket.reporterId ?? null;` → `accountHasAiFeature(db, ownerUserId)` (:114) and `buildDispatcher(db, tenantId, { userId: ownerUserId })` → `createLLMClientForUser(ownerUserId)` which **decrypts that user's stored BYOK key**.
- But `reporterId` is client-writable on the general tickets API: create schema `reporterId: z.string().optional()` spread verbatim into `createTicket(db, { ...body })`; update schema `reporterId: z.string().nullable().optional()` into `updateTicket({ ...data })`. `assertBoardAccess` checks board/tenant ownership only — never that `reporterId` belongs to the caller.

**Bypass:** attacker owns board B → `POST /boards/B/tickets` with `{ reporterId: <VICTIM_USER_ID> }` (victim ids are returned by `serializeTicket`) → `POST /api/agent-tasks/{id}/dispatch` (only checks `board.tenantId === tenant.id`, true) → worker re-derives `ownerUserId = VICTIM`, passes the victim's entitlement gate, decrypts the **victim's** key, runs the attacker's prompt on it, returns output to the attacker. Cross-tenant key theft + billing fraud on the exact hosted path this PR secures. **This PR introduces it** — pre-PR the worker used `createLLMClientFromEnv()`, so `reporterId` had zero security weight.

The author half-closed this: `agent-tasks.ts:206` forces `reporterId: user.id` at its own creation site — right instinct, but it covers one of three write paths; the general tickets `POST`/`PATCH` remain open and feed the same worker.

**Required fix:** dispatch identity must come from a trusted, non-client-writable source. Preferred: capture the authenticated dispatcher's `user.id` at each enqueue site (agent-tasks dispatch route, a2a durable path — audit ALL enqueue paths) and persist it on the JOB payload; the worker uses that persisted authenticated identity and never `ticket.reporterId` for key selection. (Note: the payload-userId the worker currently distrusts is trustworthy IFF it's set server-side from the session at enqueue — the fix is to set it there, not to fall back to the mutable row.) Add a regression test: a ticket whose `reporterId` is a foreign userId set via the tickets API must NOT cause that user's key to be decrypted. Independently, consider dropping `reporterId` from the tickets create/update request bodies (spoofing vector regardless of this PR).

### Non-blocking
- **Low** — `resolve.ts:130` `HOSTED_BYOK_DISPATCH=off` reverts hosted to `createLLMClientFromEnv()` (intended §7 break-glass) — add an operator warning-log when that path runs on hosted, since every hosted account then shares the deployment key.
- **Low** — `account-entitlement.ts:60` `.limit(1)` on memberships picks an arbitrary account for multi-account users; verify the `'active'` status literal matches `entitlementMiddleware`.
- **Nit** — the worker red-proof asserts only against a poisoned *payload* `userId`, not a poisoned `reporterId`; its 'real-owner-from-row' framing overstates the guarantee. Add the reporterId-poisoning case.

`sec-review:approved` NOT applied — gate stays HELD until the blocker fix lands and re-review passes. Fix sent back to the implementer.

RevealUIStudio:
## Guardrail-2 re-review: APPROVE — blocker closed

The reporterId tenancy blocker is fixed correctly and minimally. Verified firsthand on the fix branch, not from the report:

**Linchpin (exactly one authenticated enqueue site):** the only job-enqueue of the dispatch payload is `runDurableDispatch` (agent-tasks.ts:523), reached solely from the two `requireAgentTaskRole`-gated handlers (agent-tasks.ts:220, :345). The a2a.ts:984 hit is an SSE stream-controller enqueue, unrelated. So the payload's identity cannot originate from client input.

**Identity source corrected:** `runDurableDispatch` sets `userId: user.id` from the authenticated `user` (agent-tasks.ts:528). The worker reads `dispatcherUserId = data.userId` (agent-dispatch.ts:119) and feeds it to BOTH `accountHasAiFeature(db, dispatcherUserId)` and `buildDispatcher(db, data.tenantId, { userId: dispatcherUserId })`. `ticket.reporterId` now appears only in explanatory comments — no code path uses it for key selection. The original bypass (poison a board's ticket with a victim userId → decrypt the victim's key) is dead: a poisoned reporterId is never read.

**Regression test:** the planted-foreign-reporterId test was red-proofed against the vulnerable head (2 red → 4/4 green after fix), so it genuinely guards the fix rather than passing trivially.

**Non-blocking items also addressed:** break-glass operator warning-log when HOSTED_BYOK_DISPATCH is explicitly off on hosted; the `.limit(1)` multi-account ambiguity carries a TODO; the `'active'` membership-status literal confirmed matching entitlementMiddleware.

The resolver core held on the first pass and is unchanged. §5.5 admin agent-stream rewrite remains correctly deferred to its own PR.

**Verdict: APPROVE.** This clears the guardrail-2 gate. `sec-review:approved` is the owner's to apply (this is agent-dispatched work on a security surface — I record the verdict, the owner disposes). It should be applied only now that the recorded verdict is APPROVE.

RevealUIStudio:
## Guardrail-2 security verdict: HOLD (1 blocker, 1 medium)

Reviewed the full diff against the GAP-360 §6 invariants. The resolver design is sound and the userId handling is genuinely careful. But one dispatch site reintroduces the exact defect class §6.1 warns about, on a different parameter.

### BLOCKER (HIGH) — cross-tenant inference-key decrypt via `body.workspaceId` (agent-stream)

`resolveLLMClientForRequest` step 2 decrypts `workspace_inference_configs[workspaceId].encryptedApiKey` (`resolve.ts` `resolveSiteInferenceClient`, `decryptApiKey(config.encryptedApiKey)`) with **no check that the authenticated user is authorized for that site**. The only site that reaches step 2 with a client-controlled value is agent-stream:

```
apps/server/src/routes/agent-stream.ts:174
  workspaceId: body.workspaceId ?? c.get('tenant')?.id,
```

`c.get('tenant')` is membership-validated (`createTenantMembershipValidator`), but `body.workspaceId` takes `??` precedence and is **not** validated against the user. The membership check also validates an *account* id, while `workspaceId` is a *site* id (`workspace_inference_configs.workspaceId → sites.id`), so even a legitimate tenant header would not authorize the site.

**This PR introduces it.** On `test`, agent-stream resolves the client via `createLLMClientFromEnv()` (a single deployment key) and `body.workspaceId` is only a prompt label (`test:agent-stream.ts:154`, `Workspace: ${workspaceId}`). After this PR, that same client-controlled value selects and decrypts a per-site secret.

**Exploit (hosted, `HOSTED_BYOK_DISPATCH` default-on):**
1. Authenticated AI-entitled account (passes `requireAIAccess` + `requireTaskQuota`).
2. Attacker configures **no** personal BYOK key, so step 1 (`createLLMClientForUser`) returns null.
3. Attacker POSTs `/api/agent-stream` with `body.workspaceId = <victim site id>`.
4. Step 2 decrypts the victim site's stored provider key and builds the client.
5. Attacker's agent runs on the victim's paid inference key — cross-tenant key use, billed to the victim, under attacker-controlled prompts.

This is the §6.1 rule ("never a client-writable value; a ticket's `reporterId` has no ownership check") applied verbatim to `body.workspaceId`. The worker path correctly derives userId server-side; the site-config path does not apply the same discipline to workspaceId.

**Fix (either, centralize in the resolver so all future sites inherit it):**
- Scope `resolveSiteInferenceClient` to sites the `userId` is authorized for — join `workspace_inference_configs → sites → accountMemberships` and require an active membership before decrypt; or
- Drop `body.workspaceId` from the resolver call and pass only a server-derived, membership-validated site id.

Add a test mirroring the worker's §6.1 test: "ignores a `body.workspaceId` the user has no membership in → 409 / null, no decrypt."

### MEDIUM — admin-chat BYOK access is unaudited (§6.4)

`app/api/chat/route.ts:319` calls the resolver with `{ isHosted }` and **no `auditStore`**, so `createLLMClientForUser`'s `if (auditStore)` guard is skipped and **no `byok:key:accessed` event fires** for the admin surface. §6.4 states "audit every access"; the PR discloses admin audit is "in-memory per GAP-338", but here it is absent, not in-memory. Either wire an audit store (even in-memory) or, if intentional, adjust the §6.4 claim so the invariant matches the code.

### Verified sound
- userId is argument-scoped everywhere; worker derives it from the server-captured dispatcher, not the payload (§6.1 tests present and proven-red).
- Fail-closed on hosted for failed decrypt / non-hostedViable / unknown provider → `LLMNotConfiguredError` → 409, never env fallthrough (`resolve.ts` steps 3/4).
- No key is logged, cached, serialized, or returned; error messages name the settings path, not the key (§6.2).
- a2a and admin chat pass no workspaceId, so neither reaches the step-2 decrypt (not affected by the blocker).
- Feature-flag default keeps self-hosted byte-unchanged; break-glass warns once.

Not applying `sec-review:approved`. Happy to re-verify once the workspaceId authorization lands.

RevealUIStudio:
Blocker fix is up as #1898 (targets this branch): `resolveSiteInferenceClient` now gates the site-key decrypt on `userCanAccessSite(userId, siteId)` (owner or `site_collaborators`), so a client-supplied `workspaceId` can no longer decrypt another tenant's site key. +3 tests, proven red against the pre-fix resolver; 20/20 resolve suite, typecheck + Biome clean. Merge #1898 into this branch to clear the guardrail-2 blocker. The MEDIUM admin-chat audit-store gap (§6.4) is unaddressed by that fix and left for your call.

PR #1900, opened 2026-07-16
refactor(setup,mcp): extract shared revvault CLI client

RevealUIStudio:
## Summary

The fleet secrets rule anticipates exactly one shared TypeScript client for invoking the `revvault` secrets CLI. Today six call sites each hand-rolled the spawn: the MCP OAuth provider (the one fully-featured implementation, with a `RevvaultError` class and a path-safety guard) plus five setup/admin/probe scripts with divergent, partial reimplementations.

This PR extracts the shared client to `@revealui/setup/revvault` and migrates all six call sites onto it.

- New module `packages/setup/src/revvault/index.ts`:
  - `createRevvaultVault()` / `createMemoryVault()` — the async `Vault` (get/set/delete/list), moved verbatim from the MCP OAuth provider for its long-lived token-storage use case.
  - `readRevvaultSecret()` — tolerant sync lookup, returns `undefined` on any failure.
  - `requireRevvaultSecret()` — fail-fast sync lookup; throws a `RevvaultError` naming the exact vault path and the `revvault set <path>` command to fix it, per the fleet secrets rule ("missing secrets must be loud and name the path to set").
  - `writeRevvaultSecret()` / `revvaultSecretExists()` — the write and existence-check halves used by the guarded generators.
  - `assertSafePath()` / `assertSafePrefix()` — the same traversal/shell-metacharacter guard the MCP provider already had, now applied uniformly (previously only two of the six call sites validated their paths at all).
- `packages/mcp/src/oauth.ts` now imports and re-exports from the shared module; its own public exports (`Vault`, `RevvaultError`, `createRevvaultVault`, `createMemoryVault`, `RevvaultVaultOptions`) are unchanged, so `remote-client.ts` and `index.ts` need no changes.
- Five script call sites migrated, each keeping its original tolerance/strictness:
  - `scripts/setup/credentials.ts` — tolerant multi-key YAML-ish lookup (unchanged parsing, now backed by `readRevvaultSecret`).
  - `scripts/setup/gen-staging-secrets.ts` — its existing injectable `VaultExists`/`VaultWrite` test seams now delegate their default implementations to the shared client.
  - `scripts/setup/stripe-revvault-sync.ts` — its injectable `RevvaultSetter` default now delegates to `writeRevvaultSecret`.
  - `scripts/admin/bootstrap.ts` — required lookups (email/password) now throw via `requireRevvaultSecret` instead of a hand-rolled not-found branch; optional lookups (name, force-rotate) use `readRevvaultSecret`.
  - `scripts/electric-latency-probe/probe.ts` — required lookups keep the probe's own formatted failure message via a thin wrapper around `requireRevvaultSecret`; optional lookups use `readRevvaultSecret`. This also fixes a stale fallback binary path (`~/suite/revvault/...`, a retired location) picked up as a side effect of removing the local spawn wrapper.

## Home chosen: `@revealui/setup`, not `@revealui/utils`

Per extend-before-create, I audited both existing OSS packages before adding a new subpath anywhere:

- `@revealui/utils` (logger, db ssl-config, validation) has zero `child_process` precedent and is consumed broadly by admin API routes; nothing in it currently owns "invoke an external tool."
- `@revealui/setup` already targets `node24` explicitly, its own package description already says "secrets validation," it already spawns a child process (`execSync` in `system-tune/detect.ts`), and one of the five migrated call sites (`scripts/admin/bootstrap.ts`) already depended on it (`@revealui/setup/bootstrap`).

Both packages resolve identically for the monorepo's two consumption paths (workspace `package.json` + built `dist/` exports for real packages; the root `tsconfig.json` `@revealui/setup/*` wildcard for the `scripts/` tsx call sites), so there was no bundling-safety tiebreaker — the tiebreaker was which package already owns this kind of tooling. `@revealui/setup` was the better fit on the merits, not just the fallback.

`@revealui/mcp` (Fair Source) depending on `@revealui/setup` (MIT) is the allowed Pro→OSS direction; the reverse would not have been.

## Test plan

- [x] `pnpm --filter @revealui/setup build` — new `revvault` tsup entry builds clean
- [x] `pnpm --filter @revealui/setup typecheck` — clean
- [x] `pnpm --filter @revealui/setup test` — 98/98 pass (19 new tests for the extracted client, mocking `node:child_process`)
- [x] `pnpm --filter @revealui/mcp typecheck` / `build` — clean after building its dependency graph
- [x] `pnpm --filter @revealui/mcp test` — 423 passed, 5 skipped (unchanged; no existing test targets `oauth.ts` directly, so this confirms no regression elsewhere in the package)
- [x] `pnpm exec vitest run scripts/setup/__tests__/gen-staging-secrets.test.ts scripts/setup/__tests__/stripe-revvault-sync.test.ts` — 20/20 pass (their existing injectable-dependency tests don't exercise the default implementations directly, so this only confirms no import/wiring break)
- [x] Runtime smoke test of `credentials.ts` and `gen-staging-secrets.ts --dry-run` via `tsx` (no live revvault installed in this environment; both degrade the same way they did before — tolerant fallback / dry-run)
- [x] `biome check` clean on all touched files
- [x] Grepped for remaining direct `spawnSync`/`execFileSync`/`spawn(` of the revvault binary outside the shared client — none found
- [x] `pnpm validate:changesets` — clean

## Security note

This PR touches `packages/mcp` (Fair Source), which is a security-review-gated surface. Per the review-before-merge convention, this must not merge without a recorded coordinator verdict (`sec-review:approved` label or equivalent). I have not applied that label.

### DIFF
--- a/.changeset/revvault-shared-client.md
+++ b/.changeset/revvault-shared-client.md
@@ -1,0 +1,14 @@
+---
+"@revealui/setup": minor
+"@revealui/mcp": patch
+---
+
+Add a shared `revvault` CLI client at `@revealui/setup/revvault`, consolidating
+the six hand-rolled `revvault` spawn implementations across the MCP OAuth
+provider and the setup/admin/probe scripts into one module. Adds a fail-fast
+`requireRevvaultSecret` helper that throws naming the exact vault path when a
+secret is missing, alongside tolerant (`readRevvaultSecret`) and write
+(`writeRevvaultSecret`, `revvaultSecretExists`) variants for scripts that
+previously reimplemented their own spawn + error handling. `@revealui/mcp`'s
+OAuth provider now consumes the shared client with no change to its public
+API or behavior.
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "@playwright/test": "catalog:",
     "@revealui/core": "workspace:*",
     "@revealui/services": "workspace:*",
+    "@revealui/setup": "workspace:*",
     "@size-limit/file": "^12.1.0",
     "@size-limit/webpack": "^12.1.0",
     "@stripe/mcp": "^0.3.3",
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -16,6 +16,7 @@
     "@revealui/db": "workspace:*",
     "@revealui/knowledge-graph": "workspace:*",
     "@revealui/security": "workspace:*",
+    "@revealui/setup": "workspace:*",
     "dotenv": "catalog:",
     "jose": "catalog:"
   },
--- a/packages/mcp/src/oauth.ts
+++ b/packages/mcp/src/oauth.ts
@@ -38,9 +38,6 @@
  *      `saveTokens` verbatim.
  */
 
-import { spawn } from 'node:child_process';
-import { homedir } from 'node:os';
-import { join } from 'node:path';
 import type { OAuthClientProvider } from '@modelcontextprotocol/sdk/client/auth.js';
 import type {
   OAuthClientInformation,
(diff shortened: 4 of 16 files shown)


RevealUIStudio:
## Security review (fable, adversarial pass)

Reviewed against the pre-PR packages/mcp/src/oauth.ts as the behavioral baseline.

**Checked:**
- Vault CRUD port: line-by-line equivalent to the original (same error ordering in get(), including throw-before-not-found on nonzero exit; regex-to-includes conversions are semantically equal; argv-array spawn, no shell; identity passed via env; timeouts preserved).
- New sync helpers: argv-array spawnSync, fail-fast errors name the vault path and never include secret values, tolerant/required split is coherent, exists() reads only the exit code.
- Call-site migrations: semantics preserved per site (tolerance kept where a site deliberately tolerated absence).
- Dependency direction: mcp (Pro) consuming setup (MIT) is sound; no cycle.

**Finding (fixed in-branch):** the regex-to-charset conversion of the path guards dropped the anchored pattern's one-or-more requirement, so empty paths and prefixes passed where the original threw, and an empty prefix would have let list('') enumerate the entire vault. Fixed with explicit length checks + a regression test proven red without the fix (commit 6a01876dc).

**Verdict: APPROVE** as of 6a01876dc. Clearing the gate (the sec-review label) is the owner's action.

PR #1902, opened 2026-07-16
feat(mcp): revealui-mcp bin launcher + mcpName + server.json for MCP Registry listing

RevealUIStudio:
## What

Makes `@revealui/mcp` listable on the MCP Registry (registry.modelcontextprotocol.io) and runnable straight from npm:

- `packages/mcp/src/cli.ts`: a `revealui-mcp` bin that launches one first-party server per process (`npx @revealui/mcp <server>`). Static allowlist over the 12 servers already shipped in `dist/servers/`; no user input reaches an import path; no behavior change to any server. `code-validator` stays repo-only (it is excluded from the build at `packages/mcp/tsconfig.json`).
- `packages/mcp/package.json`: adds `bin` and `mcpName: io.github.revealuistudio/revealui` (the registry requires `mcpName` in the published package to match `server.json`'s `name`).
- `packages/mcp/server.json`: registry metadata (schema 2025-12-11), npm package entry with stdio transport and the required positional server argument.
- Changeset: patch bump.

## Why

First zero-gate ecosystem listing from the cert/partner roadmap (`.jv/docs/anthropic-certs/partner-programs.md` has the publish runbook). The package is already public on npm (0.7.1) but had no runnable entrypoint, so a registry listing would have been broken without the bin.

## Audit

Pre-change audit confirmed: no existing bin or CLI in the package to extend (servers run via repo-local `tsx`/`pnpm` scripts only, per `packages/mcp/configs/README.md`); each server module self-starts on import (e.g. `src/servers/stripe.ts` runs `main()`), which the launcher relies on.

## Verification

- Build, typecheck, lint, tests green (423 passed, 5 skipped).
- Live protocol smoke: `node dist/cli.js docs` answered an MCP `initialize` with a valid JSON-RPC response from `revealui-docs`; no-arg and unknown-name paths exit 2 with usage.
- Fleet security checker (`security-pr-review-check.js --diff origin/test`): no security-sensitive files, no coordinator gate.

## Post-merge (owner, from the runbook)

After the changeset release publishes 0.7.2: `mcp-publisher init/login github/publish` per `.jv/docs/anthropic-certs/partner-programs.md`. Note the namespace assumption: `io.github.revealuistudio` requires the GitHub login used for `mcp-publisher login github` to control the RevealUIStudio account; if publishing as joshua-v-dev instead, flip `mcpName` + `server.json` name to `io.github.joshua-v-dev/revealui` (or use the domain-verified `com.revealui/*` path) before publishing.

### DIFF
--- a/.changeset/mcp-registry-listing.md
+++ b/.changeset/mcp-registry-listing.md
@@ -1,0 +1,5 @@
+---
+"@revealui/mcp": patch
+---
+
+Add a `revealui-mcp` bin launcher, the `mcpName` field, and `server.json` so the package is listable on the MCP Registry (registry.modelcontextprotocol.io) and runnable straight from npm: `npx @revealui/mcp <server>`. The launcher is a static allowlist over the servers already shipped in `dist/servers/`; no behavior change to any server.
--- a/.changeset/mcp-registry-listing.md
+++ b/.changeset/mcp-registry-listing.md
@@ -1,5 +1,5 @@
 ---
-"@revealui/mcp": patch
+"@revealui/mcp": minor
 ---
 
 Add a `revealui-mcp` bin launcher, the `mcpName` field, and `server.json` so the package is listable on the MCP Registry (registry.modelcontextprotocol.io) and runnable straight from npm: `npx @revealui/mcp <server>`. The launcher is a static allowlist over the servers already shipped in `dist/servers/`; no behavior change to any server.
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -8,6 +8,10 @@
     "directory": "packages/mcp"
   },
   "license": "FSL-1.1-MIT",
+  "bin": {
+    "revealui-mcp": "./dist/cli.js"
+  },
+  "mcpName": "io.github.revealuistudio/revealui",
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.29.0",
     "@revealui/config": "workspace:*",
--- a/packages/mcp/server.json
+++ b/packages/mcp/server.json
@@ -1,18 +1,18 @@
 {
   "$schema": "https://static.modelcontextprotocol.io/schemas/2025-12-11/server.schema.json",
   "name": "io.github.revealuistudio/revealui",
-  "description": "First-party MCP servers for the RevealUI runtime: content, email, memory, and Stripe servers for RevealUI apps, plus adapters for Stripe, Supabase, Vercel, Neon, Playwright, and Next.js devtools. Launch one server per process: revealui-mcp <server>.",
+  "description": "MCP servers for the RevealUI runtime. First-party: contracts, docs, revealui-content, revealui-email, revealui-memory, revealui-stripe. Adapters (require pnpm on PATH; they launch the upstream MCP package): stripe, supabase, vercel, neon, playwright, next-devtools. Launch one server per process: revealui-mcp <server>.",
   "repository": {
     "url": "https://github.com/RevealUIStudio/revealui",
     "source": "github"
   },
-  "version": "0.7.2",
+  "version": "0.7.1",
   "packages": [
     {
       "registry_type": "npm",
       "registry_base_url": "https://registry.npmjs.org",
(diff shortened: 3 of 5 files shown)


RevealUIStudio:
## Coordinator Security Verdict (independent adversarial review, non-author agent)

**VERDICT: APPROVE** (reviewed at bd9dedb6d; findings addressed in 26c4544c4)

### Rationale

I attempted to refute this change's safety and could not find a security blocker. The launcher's argv handling is safe by construction: `process.argv[2]` is used only as a key into a static, hardcoded `SERVERS` allowlist (src/cli.ts), so no user input ever reaches an `import()` path or a shell; an unknown key falls through to a usage message and `exit(2)`. Critically, the bin does not add any new executable code: all 12 modules already ship inside `dist/` under the existing `"files": ["dist"]`, so `node node_modules/@revealui/mcp/dist/servers/stripe.js` was already reachable from a published install pre-PR. The bin and registry listing add discoverability and a documented entry point, not a new attack surface. Spot-checking stripe/supabase/vercel/neon/playwright/next-devtools: every one spawns a hardcoded third-party MCP package with a static argument array, passes credentials via env rather than argv, reads secrets only from env, and logs them as `[configured]`, never the value, and never to stdout where it would corrupt JSON-RPC. The `revealui-*` servers use `StdioServerTransport` with no network bind. `server.json`/`mcpName` leak nothing, and the namespace matches between the two files as the registry requires. Additive and non-breaking for existing consumers.

### Findings (all non-blocking; disposition in follow-up commit 26c4544c4)

- **[Suggestion] server.json version drift** — was hardcoded 0.7.2 vs package.json 0.7.1. FIXED: server.json now tracks package.json and a unit test (`src/__tests__/server-json.test.ts`) enforces lockstep, so the version-bump PR fails until both agree.
- **[Suggestion] Adapter launchers fetch-and-run unpinned upstream MCP packages at runtime** (pre-existing behavior; this PR makes it publicly discoverable). PARTIALLY ADDRESSED: the registry description now states the adapters require pnpm on PATH and launch the upstream package. Pinning the upstream versions is a recommended follow-up.
- **[Nit] feat commit vs patch changeset** — FIXED: changeset is now minor per the fleet 0.x versioning rule (new capability = minor).
- **[Nit] description imprecision** (Stripe listed twice, contracts/docs omitted) — FIXED: description now lists all 12 servers accurately.
- **[No finding] next-devtools transient port probe** binds-then-closes; pre-existing, not a served listener.

Confirmed clean: `"files": ["dist","LICENSE"]` unchanged (server.json and the changeset do not ship in the tarball); FSL-1.1-MIT posture unaffected; no secrets in any added file.

*Review basis: worktree diff `origin/test...HEAD`, independent agent pass distinct from the PR author session. Label application and merge remain owner actions.*

RevealUIStudio:
## Security review (fable, adversarial pass)

**Checked:**
- The launcher's argument handling: the allowlist is a static object literal and process.argv[2] only ever selects a key, so no user input reaches an import path. Confirmed no dynamic path construction anywhere.
- Publish mechanics: build is plain tsc over src/** with files: ["dist"], so dist/cli.js ships in the tarball with its shebang preserved; the bin entry resolves.
- server.json: metadata only, no secrets or internal endpoints; the lockstep test pinning server.json version to package.json closes the registry version-drift failure mode.
- The supabase adapter mention is the documented customer-facing adapter, not internal-stack drift.

**One finding (non-blocking, fix before or right after merge):** prototype-chain keys defeat the truthiness guard. SERVERS['__proto__'] returns Object.prototype and SERVERS['constructor'] returns Function, both truthy, so `revealui-mcp __proto__` crashes with an unhandled TypeError instead of the usage message. Not exploitable (nothing user-controlled is invoked as code), but a published bin should not crash on adversarial argv. One-line fix in src/cli.ts:

```ts
const launch =
  requested && Object.hasOwn(SERVERS, requested) ? SERVERS[requested] : undefined;
```

**Verdict: APPROVE** with the nit recorded. The failing check is the security gate itself, which clears via the sec-review:approved label (owner action).

PR #1909, opened 2026-07-16
fix(ci): classify security-enforcement machinery as security-sensitive

RevealUIStudio:
## The gap

The `security-review-gate.cjs` `SECURITY_PATHS` list decides whether a PR is security-sensitive (and thus needs a recorded review verdict before merge). It covered auth, license, payment, and content surfaces but **not the security-enforcement machinery itself**. So a PR editing the gate script, the audit-guard decision script, or their workflows classified as NOT security-sensitive and bypassed the review gate. Surfaced 2026-07-16 when the sec-audit-label-guard work went in without a recorded verdict.

## The change

Add self-protection markers to `SECURITY_PATHS` (exact substrings):

- `scripts/validate/security-review-gate`
- `scripts/validate/sec-audit-label-decision`
- `.github/workflows/security-review-gate`
- `.github/workflows/sec-audit-label-guard`
- `.github/workflows/security.yml`

Now any PR touching the machinery that decides what is security-sensitive is itself classified security-sensitive and must carry a recorded verdict.

Also exports `SECURITY_PATHS` + `classifyFiles` and guards `main()` behind `require.main === module` so the new unit test asserts classification without spawning the CLI (matches `sec-audit-label-decision.cjs`). Behavior when run directly is unchanged.

## Test

New `scripts/validate/__tests__/security-review-gate.test.ts` (12 cases): each new marker's file classifies as sensitive, a machinery-only changeset is flagged, and an unrelated file (`apps/marketing/app/components/Hero.tsx`) is NOT flagged. Red-proof: with the markers removed, 11 cases fail while the unrelated-file case still passes — isolating the markers. All 12 pass with the fix; the full pre-push gate passes.

## Self-trigger (intended)

Because the in-repo gate runs the branch's OWN `security-review-gate.cjs`, adding the `scripts/validate/security-review-gate` marker means **this PR classifies itself as security-sensitive** and requires the `sec-review:approved` label before merge. That is correct and intended — the gate self-protects immediately. Confirmed locally: `node scripts/validate/security-review-gate.cjs --diff origin/test` now HOLDs (exit 1) on this branch. The owner applies the label after review.

## Lockstep

The fleet checker carries an identical `SECURITY_PATHS` list and is being widened with the same five markers in lockstep (separate change). Keep the two lists in lockstep on every future widening.

### DIFF
--- a/scripts/validate/__tests__/security-review-gate.test.ts
+++ b/scripts/validate/__tests__/security-review-gate.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, it } from 'vitest';
+
+const { SECURITY_PATHS, classifyFiles } = require('../security-review-gate.cjs');
+
+// The enforcement-machinery markers this gate must self-protect. A PR editing
+// any of these files has to carry a recorded guardrail-2 verdict before merge.
+const ENFORCEMENT_MACHINERY_FILES = [
+  'scripts/validate/security-review-gate.cjs',
+  'scripts/validate/sec-audit-label-decision.cjs',
+  '.github/workflows/security-review-gate.yml',
+  '.github/workflows/sec-audit-label-guard.yml',
+  '.github/workflows/security.yml',
+];
+
+describe('classifyFiles — enforcement-machinery self-protection', () => {
+  it.each(ENFORCEMENT_MACHINERY_FILES)('flags %s as security-sensitive', (file) => {
+    expect(classifyFiles([file])).toContain(file);
+  });
+
+  it('flags a whole changeset when only a machinery file is touched', () => {
+    const changed = ['README.md', 'scripts/validate/sec-audit-label-decision.cjs'];
+    expect(classifyFiles(changed)).toEqual(['scripts/validate/sec-audit-label-decision.cjs']);
+  });
+
+  it('does NOT flag an unrelated file (red-proof for the new markers)', () => {
+    // This benign path contains none of the machinery markers. If the markers
+    // were removed, ENFORCEMENT_MACHINERY_FILES above would stop being flagged;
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
## Guardrail-2 verdict: APPROVE

Reviewed firsthand (dispatcher, not author). This widens the gate to classify the security-enforcement machinery itself as sensitive, closing the gap where a PR editing the gate/guard scripts or their workflows bypassed guardrail-2.

- **Markers correct + lockstep** — the 5 substring markers (security-review-gate + sec-audit-label-decision scripts, security-review-gate + sec-audit-label-guard + security.yml workflows) are added verbatim, and confirmed byte-identical in value to the fleet checker's copy (jv#501). Keeps the two-lists-in-lockstep discipline.
- **CLI behavior preserved** — the added `module.exports` + `if (require.main === module) main()` guard is the standard testability pattern (matches sec-audit-label-decision.cjs); run directly, main() still fires all the same process.exit paths. Verified in the diff.
- **Test is genuine** — new `__tests__/security-review-gate.test.ts`, 12 pass; red-proof is marker-isolated (removing only the marker block fails the 11 machinery-classification/marker cases while the unrelated-file negative case still passes), so the markers are proven load-bearing, not a tautology.
- **Self-trigger is intended + working** — `--diff origin/test` now reports SECURITY-SENSITIVE (exit 1) on this PR's own edit to the gate. The gate self-protects immediately; correct.

Not a crown-jewel logic change (path-list + testability refactor), so a focused review is proportional. `sec-review:approved` is the owner's to apply after review (agent-dispatched work). Note: applying that label now runs through the harness audit gate + this newly-widened gate — dogfooding the enforcement on the enforcement PR.

PR #1911, opened 2026-07-17
feat(harnesses): add OpenCodeAdapter, the first external-CLI harness

RevealUIStudio:
## Summary

Phase 3 of the OpenCode integration: `OpenCodeAdapter`, the first external-CLI `HarnessAdapter` in `@revealui/harnesses`. Wraps the `opencode` CLI over `execFile` (mirroring the CLI-wrapping pattern in `translation-layer.ts`), promotes its capability profile from declared-roadmap to shipped, and adds a config writer for the governed RevealUI MCP endpoint.

- `OpenCodeAdapter` (`src/adapters/opencode-adapter.ts`): detects (`opencode --version`), drives one-shot `opencode run <prompt> --format json` for `headless-prompt`/`generate-code`/`analyze-code`, writes config for `apply-config`, uses existing `config-sync.ts` machinery for `sync-config`/`diff-config`, reports `apply-edit` as explicitly unsupported (opencode applies its own edits during `run`), and discovers running instances via the existing process-detector.
- Promotes the `opencode` capability profile from `ROADMAP_PROFILES` to `TOOL_PROFILES` (`protocol/capabilities.ts`), registers the adapter in `auto-detector.ts`, and updates the pinned contract tests (`protocol-types.test.ts`) accordingly.
- Registers the first entry in the previously-empty content-generator registry: `OpenCodeGenerator` (`content/generators/opencode.ts`) renders commands/agents into `.opencode/commands/*.md` and `.opencode/agents/*.md`. Rules/skills are intentionally not rendered by this generator — OpenCode has no documented native surface for either (rules already reach OpenCode via the separate, existing `protocolConfigToAgentsMd` → `AGENTS.md` path).
- Adds `protocolConfigToOpencodeConfig` (`protocol/config-normalizer.ts`) emitting the governed-MCP `opencode.json` block from the design doc — bearer token is always the `{env:VAR}` substitution syntax, never a literal value, regardless of what a caller places in `ProtocolConfig.environment.variables`.

## Test plan

- [x] `pnpm --filter @revealui/harnesses test` — 21 files / 319 tests, all green (3 new test files + 2 updated pinned files)
- [x] `pnpm --filter @revealui/harnesses typecheck` — clean
- [x] `pnpm --filter @revealui/harnesses lint` — clean (Biome)
- [x] Adapter contract tests: registry lifecycle (register / dup-throw / unregister-disposes), capabilities shape, `execute()` mapping for every command variant including the **graceful no-binary failure path** (exercised against the real absent `opencode` binary in this environment, not a mock — `which opencode` confirms it isn't installed)
- [x] `apply-config` path-traversal guard (refuses writes outside the project root) + nested-directory creation
- [x] Config-writer security test: asserts a literal token planted anywhere in `ProtocolConfig.environment.variables` never appears in the emitted JSON — only the `{env:VAR}` substitution string does
- [x] `opencode run --format json` output-shape parsing: **NOT verified against a real binary** — `opencode` is not installable in this build/sandbox environment. `parseOpenCodeRunOutput` is unit-tested against the documented-but-unverified shape (object with `output`/`text`/`message`/`content`, bare JSON string, non-JSON plain text) with graceful fallback in every case. This is flagged, not silently claimed as verified — real-binary pinning is a deferred follow-up for whichever environment first has `opencode` installed.
- [x] `.changeset/opencode-adapter.md` added (`@revealui/harnesses: minor`)
- [x] Internal security-sensitive-path checker run against this diff: reported no security-sensitive files, no coordinator gate required. `packages/harnesses` is confirmed NOT on the sensitive-path list. Flagged anyway per the build brief: the config writer does emit a token *env-var name* (never a value) into `opencode.json`; the no-literal-token invariant has its own dedicated test.

Left unmerged per instructions.

### DIFF
--- a/.changeset/opencode-adapter.md
+++ b/.changeset/opencode-adapter.md
@@ -1,0 +1,5 @@
+---
+"@revealui/harnesses": minor
+---
+
+add OpenCodeAdapter, the first external-CLI harness adapter: detects and drives the `opencode` CLI (headless run, config apply/sync/diff, running-instance discovery), promotes the `opencode` capability profile from roadmap to shipped, registers an OpenCode content generator (commands + agents), and adds `protocolConfigToOpencodeConfig` for wiring the governed RevealUI MCP endpoint into `opencode.json` via env-var token substitution (never a literal token).
--- a/packages/harnesses/src/__tests__/opencode-adapter.test.ts
+++ b/packages/harnesses/src/__tests__/opencode-adapter.test.ts
@@ -1,0 +1,277 @@
+import { mkdtemp, readFile, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import {
+  describeOpenCodeExecError,
+  OpenCodeAdapter,
+  parseOpenCodeRunOutput,
+} from '../adapters/opencode-adapter.js';
+import { HarnessRegistry } from '../registry/harness-registry.js';
+import type { HarnessCommand, HarnessEvent } from '../types/core.js';
+
+describe('OpenCodeAdapter', () => {
+  let projectRoot: string;
+  let adapter: OpenCodeAdapter;
+
+  beforeEach(async () => {
+    projectRoot = await mkdtemp(join(tmpdir(), 'opencode-adapter-test-'));
+    adapter = new OpenCodeAdapter({ projectRoot });
+  });
+
+  afterEach(async () => {
+    await adapter.dispose();
+    await rm(projectRoot, { recursive: true, force: true });
+  });
+
+  it('has id opencode and a display name', () => {
+    expect(adapter.id).toBe('opencode');
+    expect(adapter.name).toBe('OpenCode');
+  });
+
+  describe('getCapabilities', () => {
+    it('declares generate/analyze/config support but not edit or workboard', () => {
+      const caps = adapter.getCapabilities();
+      expect(caps).toEqual({
+        generateCode: true,
+        analyzeCode: true,
+        applyEdit: false,
+        applyConfig: true,
+        readWorkboard: false,
+        writeWorkboard: false,
+      });
+    });
+  });
+
+  describe('isAvailable / getInfo -- graceful no-binary failure', () => {
+    // `opencode` is not installed in this build environment (verified via
+    // `which opencode` during the audit for this PR). These assertions
+    // exercise the REAL execFile ENOENT path, not a mock.
+    it('isAvailable returns false when the opencode binary is missing', async () => {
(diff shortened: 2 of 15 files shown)


RevealUIStudio:
## Review verdict — Phase 3 OpenCodeAdapter

**Verdict: APPROVE (quality review).** Not security-gated (the fleet checker confirms `packages/harnesses` is not a security-sensitive surface), so this is a correctness/quality pass, not a `sec-review:approved` gate. Builder-adjacent note: I authored the GAP-371 spec and dispatched this build, so a peer read is welcome but not required to merge an unt-gated PR.

### What I verified against source
- **Config writer is structurally leak-proof (the one security-relevant surface).** `protocolConfigToOpencodeConfig` (`config-normalizer.ts`) builds the Authorization header exclusively from the env-var *name* — `Bearer {env:${tokenEnvVar}}` — and deliberately never reads `config.environment.variables` for a token value. There is no code path where a literal bearer token could reach the emitted config, which is stronger than the "test that it doesn't leak" the brief asked for. Permission keys reuse the existing `isSafeMcpServerName` prototype-pollution allowlist.
- **Adapter (`opencode-adapter.ts`)** — clean `execFile` wrap with sane timeouts (`--version` 3s, `run` 120s) and a 10 MB buffer cap; complete `execute()` mapping with honest `success:false` for `apply-edit` and workboard (matching `RevealUIAgentAdapter`); an unprompted **path-traversal guard** in `applyConfig` (rejects writes outside the resolved project root); graceful `ENOENT`/no-binary handling. No authored regex.
- Profile promoted `ROADMAP_PROFILES` → `TOOL_PROFILES`; adapter registered in `auto-detector.ts`; content generator registers (fixing the previously-empty registry); pinned tests updated. 319 harness tests green.

### One tracked follow-up (not a defect)
- **The `opencode run --format json` output shape is not verified against a real binary** — `opencode` isn't installable in the build sandbox, so `parseOpenCodeRunOutput` is a documented, defensive best-guess (tries `output`/`text`/`message`/`content`, falls back to raw JSON, then plain text). This is correctly flagged as deferred in the code and PR body, not claimed as verified. **Phase 4 (e2e against a real `opencode` binary → `/api/mcp`) is the natural home to pin it** — until then, treat the parse keys as provisional.

### Nits (non-blocking)
- `tokenEnvVar` is interpolated into the header without a name-validity check. Not a security hole (no token value is ever present, and a malformed name just yields a non-resolving substitution → auth fails closed at `/api/mcp`), but validating it with the same `isSafeMcpServerName` barrier would be tidy.
- `parseOpenCodeRunOutput`'s key-guessing is a reasonable stopgap; delete the guesses once the real shape is pinned in Phase 4.

Mergeable at owner discretion — no label required. Recommend the deferred real-binary pin be carried on GAP-371 Phase 4.

PR #1912, opened 2026-07-17
fix(mcp): omit resources surface on governed mcp mount (GAP-373)

RevealUIStudio:
## GAP-373 (blocker): governed MCP resources surface bypasses per-request auth + audit

Post-merge finding from the #1910 guardrail-2 review, **Blocker 2**.

### The bug

`createRevealuiContentServer(options)` advertised `capabilities: { tools: {}, resources: {} }` and registered `ListResourcesRequestSchema` + `ReadResourceRequestSchema` handlers that resolve credentials via the process-global `resolveCredentials()` (ambient `REVEALUI_API_KEY`), **not** the per-request `options.credentialsProvider`, and write **no** `auditSink` receipt. The tools path (`CallToolRequestSchema`) is correctly governed via `credentialsProvider` + `auditSink`.

The governed `/api/mcp` mount ([mcp-endpoint.ts:164](apps/server/src/routes/mcp-endpoint.ts)) constructs the factory **with** a `credentialsProvider`. So on that endpoint any authenticated MCP client could `resources/read` `revealui-content://<collection>/<id>` and have the server fetch with the ambient admin-equivalent key instead of the caller's token, **unaudited** — a cross-tenant read channel the moment `REVEALUI_API_KEY` is set in the API process env.

### The fix (design section 4-D scope)

The countersigned design decision (section 4-D) scopes Phase 1's governed endpoint to **only the five read tools, no resources surface**. Rather than re-route the resource handlers through the governed seams, the resources surface is made **structurally absent** when governed:

- The `resources` capability is advertised only when **not** governed (`options?.credentialsProvider === undefined`). Governed construction advertises `{ tools: {} }` only.
- Both `server.setRequestHandler(...Resource...)` registrations are guarded behind `if (!governed)`.
- Stdio/local mode (constructed with no options, single-operator trust) keeps resources unchanged.

The governed mount never advertises or depends on resources (grep of [mcp-endpoint.ts](apps/server/src/routes/mcp-endpoint.ts) for `resource`/`capabilit` is empty; the construction comment already states "Phase 1 exposes read tools only"), so disabling is the minimal correct shape and the smaller blast radius the design prefers over re-routing.

### Files changed

- [packages/mcp/src/servers/factories/revealui-content.ts](packages/mcp/src/servers/factories/revealui-content.ts): `governed` flag, conditional `resources` capability, both resource handlers wrapped in `if (!governed)`. Tools path untouched.
- [packages/mcp/__tests__/revealui-content-governed.test.ts](packages/mcp/__tests__/revealui-content-governed.test.ts): new negative test (the original PR had **no** resources test).

### Prove-red evidence

The new test was run against the **pre-fix** factory first (TDD order) and failed exactly as expected:

```
FAIL  GAP-373 governed path exposes no resources surface > refuses resources/list and resources/read, never reaching the ambient-key backend
  expected client.listResources() to reject with McpCapabilityError
  + Received: [ posts/row-1, pages/row-1, products/row-1, media/row-1 ]   <- fetched via the ambient-key backend
Test Files  1 failed (1)
     Tests  2 failed | 8 skipped (10)
```

Both new tests fail pre-fix (capability advertised; `resources/list`/`resources/read` reach the ambient backend). After the fix both pass:

```
Test Files  1 passed (1)
     Tests  2 passed | 8 skipped (10)
```

### Test results

- `pnpm --filter @revealui/mcp exec vitest run` — **25 files passed, 437 passed, 5 skipped** (green).
- `pnpm --filter @revealui/mcp typecheck` — clean.

### pglite mount status (separate #1910 finding, report-only)

The #1910 review found `apps/server/src/routes/__tests__/mcp-endpoint.pglite.test.ts` **5/7 RED** (broken Hono mount). On this base (`test` @ #1910 merge) it is now **GREEN: 7/7 passed** — the builder's post-verdict commits fixed the mount. Not a second blocker.

Refs GAP-373, #1910 review Blocker 2

### DIFF
--- a/packages/mcp/__tests__/revealui-content-governed.test.ts
+++ b/packages/mcp/__tests__/revealui-content-governed.test.ts
@@ -18,7 +18,7 @@
 import { createServer as createHttpServer, type Server as NodeHttpServer } from 'node:http';
 import type { AddressInfo } from 'node:net';
 import { afterEach, describe, expect, it } from 'vitest';
-import { McpClient } from '../src/client.js';
+import { McpCapabilityError, McpClient } from '../src/client.js';
 import {
   createRevealuiContentServer,
   type McpToolAuditRecord,
@@ -301,5 +301,67 @@ describe('I-5 a receipt per call', () => {
     expect(result.isError).toBeFalsy();
     // The read executed despite the audit failure (degrade, not fail-closed).
     expect(backend.calls.length).toBeGreaterThan(0);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// GAP-373 — the governed path exposes no resources surface (#1910 Blocker 2)
+// ---------------------------------------------------------------------------
+//
+// The resource handlers read through the process-global resolveCredentials()
+// (ambient REVEALUI_API_KEY), never the per-request credentialsProvider, and
+// write no receipt. On the governed mount that is a cross-tenant read channel:
+// any authenticated client could resources/read with the ambient admin key,
+// unaudited. Per the countersigned design §4-D, Phase 1's governed endpoint
+// exposes ONLY the five read tools — the resources capability must be
+// structurally absent when a credentialsProvider is present, not merely
+// credential-gated. (Stdio/local mode keeps resources; see the stdio launcher.)
+
+describe('GAP-373 governed path exposes no resources surface', () => {
+  it('does not advertise the resources capability on the governed mount', async () => {
+    const backend = await startFakeBackend(okBackend);
+    const governed = await startGoverned({
+      backendUrl: backend.url,
+      authInfo: () => ({ token: 'tokenA', extra: { userId: 'A', accountId: 'acct-A' } }),
+    });
+    const client = await connectClient(governed.url);
+    const caps = client.getServerCapabilities();
+    await client.close();
+
+    // Tools stay; resources is structurally absent on the governed path.
+    expect(caps?.tools).toBeDefined();
+    expect(caps?.resources).toBeUndefined();
+  });
+
+  it('refuses resources/list and resources/read, never reaching the ambient-key backend', async () => {
+    const prev = process.env.REVEALUI_API_KEY;
+    const prevUrl = process.env.REVEALUI_API_URL;
+    try {
+      const backend = await startFakeBackend(okBackend);
+      // Ambient admin-equivalent credentials point at the real backend, so a
+      // resource handler that consulted resolveCredentials() (the bug) would be
+      // observable here as a backend call carrying the ambient key.
+      process.env.REVEALUI_API_KEY = 'ambient-admin-key';
+      process.env.REVEALUI_API_URL = backend.url;
+      const governed = await startGoverned({
+        backendUrl: backend.url,
+        authInfo: () => ({ token: 'tokenA', extra: { userId: 'A', accountId: 'acct-A' } }),
+      });
+      const client = await connectClient(governed.url);
+
+      await expect(client.listResources()).rejects.toThrow(McpCapabilityError);
+      await expect(client.readResource('revealui-content://posts/row-1')).rejects.toThrow(
+        McpCapabilityError,
+      );
+      await client.close();
+
+      // The ambient key never reached the backend: no cross-tenant read channel.
+      expect(backend.calls).toHaveLength(0);
+    } finally {
+      if (prev === undefined) delete process.env.REVEALUI_API_KEY;
+      else process.env.REVEALUI_API_KEY = prev;
+      if (prevUrl === undefined) delete process.env.REVEALUI_API_URL;
+      else process.env.REVEALUI_API_URL = prevUrl;
+    }
   });
 });
--- a/packages/mcp/src/servers/factories/revealui-content.ts
+++ b/packages/mcp/src/servers/factories/revealui-content.ts
@@ -548,9 +548,17 @@ async function fetchCollectionsFromAdmin(
  * consume this factory; the factory itself is transport-agnostic.
  */
 export function createRevealuiContentServer(options?: CreateRevealuiContentServerOptions): Server {
+  // A `credentialsProvider` marks the governed/hosted mount (`/api/mcp`). On
+  // that path Phase 1 exposes ONLY the five read tools (design §4-D): the
+  // resource handlers below resolve the ambient service credential rather than
+  // the caller's token and write no receipt, so on a multi-tenant mount they
+  // would be a cross-tenant, unaudited read channel. The capability is therefore
+  // structurally absent when governed — not merely credential-gated. The stdio /
+  // local mount (no options) keeps resources unchanged (single-operator trust).
+  const governed = options?.credentialsProvider !== undefined;
   const server = new Server(
     { name: 'revealui-content', version: '1.0.0' },
-    { capabilities: { tools: {}, resources: {} } },
+    { capabilities: governed ? { tools: {} } : { tools: {}, resources: {} } },
   );
 
   // Per-server memoization: resolve the effective collection set once and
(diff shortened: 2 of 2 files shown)


RevealUIStudio:


Independent non-author verification (I coordinated/specified this fix but did not write the code; see transparency note below). Closes GAP-374 / #1910 Blocker 2.

**The fix is correct, minimal, and scope-aligned (§4-D):**
- `revealui-content.ts`: `governed = options?.credentialsProvider !== undefined`; capability becomes `governed ? { tools: {} } : { tools: {}, resources: {} }`; both `ListResourcesRequestSchema` and `ReadResourceRequestSchema` handlers wrapped in `if (!governed)`. So the governed `/api/mcp` mount advertises no resources capability and registers no resource handlers. The ambient-`REVEALUI_API_KEY` read channel is structurally absent, not merely credential-gated. Tools path untouched; stdio/local mode unchanged.
- Sole governed call site (`mcp-endpoint.ts:164`) passes `credentialsProvider` and is Phase-1 read-tools-only, so resources is never needed there. Confirmed.

**Evidence:**
- New negative test asserts the resources capability is absent on the governed mount and `resources/list`/`resources/read` are refused with `McpCapabilityError` without reaching the backend (with `REVEALUI_API_KEY` set to prove no fallback). **Proven red** against the pre-fix factory (pre-fix: capability advertised, ambient key fetched `posts/pages/products/media`; post-fix: refused).
- `@revealui/mcp` suite green (437 passed); typecheck clean.

**Also confirmed (the #1910 Blocker-1 concern):** the pglite mount suite is now **7/7 green** on `test` — the broken Hono mount was fixed by the post-verdict commits, so it is NOT a second unresolved blocker.

**Not re-addressed here (deferred, tracked on GAP-374):** the High-3 unvalidated `collection` arg (arbitrary authenticated GET `/api/*`) and High-4 (IDOR proven only against a stub). Bounded to the caller's own token authority; worth a follow-up, not a blocker for this fix.

**Transparency:** I dispatched this fix, so if you want a fully arms-length second eye on a security change, that's reasonable; the change itself is small, mechanical, and evidenced. This verdict clears the resources-bypass blocker.

Apply `sec-review:approved` to record the clearance, then merge.

RevealUIStudio:
## Independent review verdict — GAP-374 resources-bypass fix

**Verdict: APPROVE** for the resources bypass (the blocker). Independent reviewer: I did not build this PR. One confirmed residual (out of this PR's scope) is flagged below and must not be lost.

### The fix is correct, complete, and lockstep
`createRevealuiContentServer` now derives `governed = options?.credentialsProvider !== undefined` and, when governed:
- Advertises `{ tools: {} }` only — the `resources` capability is **structurally absent**, not merely credential-gated (matches the design's I-4 posture).
- Registers neither `ListResourcesRequestSchema` nor `ReadResourceRequestSchema` (both wrapped in `if (!governed)`), so there is no handler to reach the ambient `resolveCredentials()`.
- Leaves the tools path (`CallToolRequestSchema`) and the stdio/local mount (no `credentialsProvider` → `governed=false`) unchanged.

The negative test proves the property that matters: on the governed mount, `resources/list` and `resources/read` reject with `McpCapabilityError` **and `backend.calls` is length 0** — the ambient admin key never reaches the backend, so the cross-tenant read channel is gone. This closes GAP-374 Blocker 2 (I-1/I-2/I-4/I-5 on the resources surface). The capability/handler lockstep is called out in a comment, which is the right guard against regression.

### Confirmed residual — NOT fixed here, and it undermines Phase 2 (#1913)
GAP-374 also lists an unvalidated `collection` arg (its "High 3"). **Confirmed real, and I'd rate it higher than High given the Phase-2 interaction:** `ListContentArgsSchema`/`GetContentArgsSchema` declare `collection: z.string().min(1)` (revealui-content.ts:230,239) — a free string interpolated into `apiGet(..., \`/api/${collection}\`)`. So an authenticated caller can pass `collection=users` to `revealui_list_content` and reach `GET /api/users` — **bypassing the exact admin/pro gate #1913 puts on `revealui_list_users`.** Per-tool authz is defeatable as long as one tool can reach another tool's endpoint. It also widens the tool to arbitrary `GET /api/*` under the caller's token (bounded only by REST access control, which is a weaker backstop than the intended tool gate).

**Recommendation:** validate `collection` against the resolved exposed-collections allowlist (deny-by-default) — ideally folded into #1913 **before it merges**, since #1913's authz model is not sound while this escape is open. I'm flagging the same point on #1913 for its reviewer.

### Other GAP-374 residuals
Blocker 1 (pglite mount suite state) and High 4 (IDOR proven only against a stub backend) should be confirmed in the Phase-4 e2e against a real backend, not a fake. Not blockers on THIS PR.

Merge gate: security-sensitive; the in-repo gate holds until `sec-review:approved`. This verdict clears the review requirement for the resources fix; **the owner applies the label** (agents propose, owner disposes). The `collection`-arg escape is a separate follow-up that should gate #1913, not this PR.

RevealUIStudio:
## Fable independent adversarial verdict: APPROVE (concurring)


Per the recorded Opus/Fable review division (workboard note 2026-07-17 ~00:55Z + my ~02:15Z ack), this is the independent Fable gate verdict on the GAP-374 resources bypass. Full independence: I authored none of the GAP-371 spec, dispatches, builds, or prior verdicts on this surface.

### Verified firsthand (not inherited from the prior verdicts)
- **Re-ran the suite on the PR head** (dbbf9564f, isolated worktree): `revealui-content-governed.test.ts` 10/10 green.
- **Re-derived prove-red**: restored the base (`origin/test`) factory under the PR's tests; exactly the 2 new GAP-373 tests fail (capability advertised, ambient-key backend reached), 8 invariant tests still pass. The fix, not test drift, closes the channel.
- **Source-verified the three load-bearing claims**: the governed mount always constructs with a `credentialsProvider` (apps/server/src/routes/mcp-endpoint.ts:164) so `governed` cannot be false on the hosted path; the tools path has no ambient fallback when governed (the provider is the sole source and a throw records a `failed` receipt, revealui-content.ts:704-729); both resource handlers and the capability advert are gated on the same `governed` flag.

### One assurance the prior verdicts did not record
The negative test's `McpCapabilityError` is thrown **client-side** (`McpClient.requireCapability`, packages/mcp/src/client.ts:465/472), so on its own it would not prove the server refuses a raw `resources/read` from a client that skips the capability check, and `backend.calls === 0` is vacuous when the client never sends. I checked the server side: the handlers are unregistered (raw request gets JSON-RPC method-not-found), **and** the MCP SDK's `assertRequestHandlerCapability` (`@modelcontextprotocol/sdk` server/index.js:232-238) throws at `setRequestHandler` time if the `resources` capability is absent. So the capability/handler lockstep the in-code comment worries about is structurally enforced by the SDK: a regression that re-registers the handlers without the capability crashes construction loudly instead of silently reopening the bypass. No change requested; recording it so the lockstep comment is known to be backed by the SDK, not just discipline.

### Residual scope rulings (GAP-374, as the gap's tracking owner)
- **High 3 (unvalidated `collection` arg)**: CONFIRMED real (free `z.string().min(1)` at revealui-content.ts:230/239 interpolated into `GET /api/${collection}`). I concur with the 01:32Z verdict: it defeats per-tool authz (reach `/api/users` via `revealui_list_content`) and therefore **gates #1913**, not this PR. I will verify it is closed as part of my #1913 review.
- **High 4 (IDOR proven only vs stub) + Blocker 1 (pglite mount state)**: Phase-4 acceptance scope, against a real backend. Recorded in GAP-374.
- **Nit (tracking)**: the PR title/branch/tests cite GAP-373, but the .jv tracker for this blocker is **GAP-374** (GAP-373 is the closed native-check-surface gap). No code change needed; noting the mapping here so the audit trail does not dangle.

**Gate state**: `sec-review:approved` is already applied (owner-disposed). This verdict completes the Fable gate per the division; #1912 is clear to merge from my side. Sequencing stands: #1912 merges before Phase-4 sign-off.

PR #1913, opened 2026-07-17
feat(mcp): tool authorization, quota, and session caps for /api/mcp

RevealUIStudio:
## Summary

Extends the governed MCP endpoint (`/api/mcp`) with authorization depth, quota,
and session bounding, and fixes a concurrency defect in the governed audit
writer surfaced by the endpoint review. Identity threading, session binding, and
per-tool audit receipts already shipped; this builds on top of them (no rebuild).

## What this adds

- **Deny-by-default per-tool authorization.** Every `tools/call` passes a
  permission gate keyed `mcp:tool:<name>` through the existing
  `AuthorizationSystem` (exact-string keys, so the glob-to-RegExp match path is
  never taken) using the bound user's role, before any credential resolution or
  REST call. A denied call yields a JSON-RPC error and a `mcp:tool:denied`
  receipt with `reason: authz`.
- **Tier-filtered discovery.** `tools/list` returns only the tools the caller may
  execute at their server-derived tier. Discovery filtering is UX; a tool hidden
  from the list is still independently denied on a direct call.
- **Per-user + tier rate limiting.** Wires the existing `McpRateLimiter` at the
  `tools/call` boundary, keyed by user and tool with limits from the tier. Over
  the limit yields a JSON-RPC error and a `mcp:tool:denied` receipt with
  `reason: rate-limit`. A second user has an independent budget.
- **Metering.** One `usage_meters` row per executed tool call with
  `source: 'agent'`. Billing and audit stay separate sinks.
- **Server-derived authorization context.** The effective tier comes from the
  account entitlement record, re-resolved every request. Any client-supplied
  tier or tenant field (headers, params, args) is ignored.
- **Session caps and idle TTL.** A per-user concurrent-session cap (default 8,
  evicts the least-recently-active) and an idle TTL (default 24h) bound resource
  use.

The stdio launcher is unchanged: it constructs the factory without these seams,
so its local single-operator model keeps unlimited, unfiltered access under the
OS-account trust boundary.

## Audit hash-chain concurrency fix

The governed audit writer kept the chain head in a per-process variable with a
read-modify-write across the insert `await`. Concurrent receipts could both read
the same head and write it as `previousSignature`, forking the chain. This phase
turns on the fail-closed mutating-audit path, so every append is on the hot path.

The fix serializes appends through a promise-chain mutex so the head is read and
advanced atomically per process, and signs over a canonical (key-sorted) payload
so a receipt is reproducible from the `jsonb`-stored row. Adds
`verifyMcpAuditChain`, which verifies one process segment as a single unbroken,
unforked, tamper-evident chain. A failed write neither advances the head nor
poisons the queue.

## Invariants and tests (red-first)

Each new assertion was watched failing against the un-gated code before the
implementation made it pass:

- **I-7** role and tier gate discovery and execution independently; denials are
  audited with `reason: authz`; a free-tier caller never sees or executes the
  pro-gated tool. Disabling the seams reddened all four assertions.
- **I-8** the rate limit binds per user and tier; a second user is unaffected.
- **I-9** the effective tier follows the entitlement record, not a client-sent
  tier header.
- **I-10** opening past the per-user cap evicts the oldest session; a session
  idle past the TTL no longer accepts requests. Disabling the cap and TTL
  reddened both.
- **Audit fix** 30 interleaved audit writes form one verifiable chain. Without
  serialization all 30 rows chained to a null parent (a maximal fork).

Test files: the factory-seam suite (`packages/mcp/__tests__/`), the endpoint
invariant suite and the concurrency test (`apps/server/.../__tests__/`), mirroring
the existing Phase-1 test files.

## Verification

- `@revealui/mcp`: 440 passed / 5 skipped. `server`: 2924 passed / 1 skipped.
- `pnpm --filter @revealui/mcp typecheck` and `pnpm --filter server typecheck` clean.
- Biome clean on all changed files.
- Fleet security checker (`security-pr-review-check.js --diff origin/test`):
  flagged SECURITY-SENSITIVE (touches `packages/mcp/`), needs a recorded
  coordinator verdict before merge.

## Merge gate

This touches security-sensitive surfaces (`packages/mcp`, governed auth and
audit). It needs an independent NON-AUTHOR security verdict recorded on the PR
(`sec-review:approved`) before merge. The building session does not self-clear.
Per the model-allocation rule, a Fable adversarial pass applies to this phase.

### DIFF
--- a/apps/server/src/lib/__tests__/mcp-audit-concurrency.pglite.test.ts
+++ b/apps/server/src/lib/__tests__/mcp-audit-concurrency.pglite.test.ts
@@ -1,0 +1,97 @@
+/**
+ * Finding-1 (GAP-371 Phase-1 review MUST-FIX) — the governed-MCP audit chain
+ * head is a read-modify-write across the insert `await`. Concurrent
+ * `recordMcpToolAudit` calls that both read the same head before either advances
+ * it FORK the hash chain (two rows chained to one parent). Phase 2 turns on the
+ * fail-closed mutating-audit path, so every append is on the hot path.
+ *
+ * This test interleaves many `recordMcpToolAudit` calls against a real migrated
+ * `audit_log` (PGlite) and asserts the resulting rows form ONE unbroken,
+ * unforked, tamper-evident chain via `verifyMcpAuditChain`. Without the
+ * serialization fix it fails: multiple rows carry `previousSignature === null`
+ * (multiple roots) and the verifier reports a fork.
+ */
+
+import * as schema from '@revealui/db/schema';
+import { afterEach, beforeAll, describe, expect, it, vi } from 'vitest';
+import {
+  createTestDb,
+  type TestDb,
+} from '../../../../../packages/test/src/utils/drizzle-test-db.js';
+import {
+  __resetMcpAuditChainForTest,
+  type McpAuditChainRow,
+  recordMcpToolAudit,
+  verifyMcpAuditChain,
+} from '../mcp-audit.js';
+
+let testDb: TestDb;
+
+vi.mock('@revealui/db', async () => {
+  const actual = await vi.importActual<typeof import('@revealui/db')>('@revealui/db');
+  return { ...actual, getClient: () => testDb.drizzle };
+});
+
+const SECRET = 'test-secret-value-at-least-32-chars-long!!';
+
+beforeAll(async () => {
+  process.env.REVEALUI_AUDIT_HMAC_SECRET = SECRET;
+  testDb = await createTestDb();
+});
+
+afterEach(() => {
+  __resetMcpAuditChainForTest();
+});
+
+function receiptInput(i: number) {
+  return {
+    outcome: 'invoked' as const,
+    clientName: 'opencode',
+    sessionId: `session-${i}`,
+    userId: 'user-A',
+    accountId: 'acct-A',
+    tool: 'revealui_list_sites',
+    argsDigest: 'd'.repeat(64),
+    scalars: {},
+    durationMs: i,
+  };
+}
+
+describe('Finding-1: concurrent MCP audit writes keep one unforked chain', () => {
+  it('interleaved recordMcpToolAudit calls form a single verifiable chain', async () => {
+    __resetMcpAuditChainForTest();
+    const N = 30;
+
+    // Fire all N concurrently. Each reads the chain head BEFORE its insert await;
+    // without serialization they all read the same head and fork.
+    await Promise.all(Array.from({ length: N }, (_, i) => recordMcpToolAudit(receiptInput(i))));
+
+    const rows = await testDb.drizzle.select().from(schema.auditLog);
+    // No dropped writes.
+    expect(rows).toHaveLength(N);
+
+    const chainRows: McpAuditChainRow[] = rows.map((r) => ({
+      timestamp: r.timestamp,
+      eventType: r.eventType,
+      severity: r.severity,
+      agentId: r.agentId,
+      payload: r.payload,
+      signature: r.signature,
+      previousSignature: r.previousSignature,
+    }));
+
+    // Exactly one root and no two rows chained to the same parent (fork check),
+    // independent of the verifier — this is the property a race destroys.
+    const roots = chainRows.filter((r) => r.previousSignature === null);
+    expect(roots).toHaveLength(1);
+    const parents = chainRows
+      .map((r) => r.previousSignature)
+      .filter((p): p is string => p !== null);
+    expect(new Set(parents).size).toBe(parents.length);
+
+    // Full tamper-evident verification: one unbroken, unforked, signed chain.
+    const result = verifyMcpAuditChain(chainRows, SECRET);
+    expect(result.reason).toBeUndefined();
+    expect(result.valid).toBe(true);
+  });
+});
--- a/apps/server/src/lib/mcp-audit.ts
+++ b/apps/server/src/lib/mcp-audit.ts
@@ -20,6 +20,23 @@
  * verifiable by walking its own eventType stream. Consolidating both onto a
  * single shared chain head is a deliberate future refactor, not a Phase-1
  * concern (audit-first §Mindfulness: classified INTENTIONAL).
+ *
+ * Concurrency (GAP-371 Phase 2, Finding-1 fix). The chain head is a read-modify-
+ * write: read `lastMcpSignature`, sign against it, INSERT, then advance the head
+ * on success. Two `recordMcpToolAudit` calls racing across the insert `await`
+ * would both read the same head and both write it as `previousSignature`, FORKING
+ * the chain (two rows chained to one parent, later tampering on a stranded branch
+ * undetectable). Phase 2 turns on the fail-closed mutating-audit path, so every
+ * append is on the hot path. The fix serializes appends through a promise-chain
+ * mutex (`appendQueue`): each call's critical section runs only after the prior
+ * one has settled, so the head is read and advanced atomically per process. A
+ * failed write does NOT advance the head and does NOT poison the queue — the next
+ * append chains from the last DURABLE signature, leaving no gap.
+ *
+ * Verification tolerates the per-process segmentation this design already has: a
+ * serverless deployment runs many processes, each an independent segment rooted
+ * at `previousSignature = null`. `verifyMcpAuditChain` verifies one segment (one
+ * process's rows) as a single unbroken, unforked, tamper-evident chain.
  */
 
 import { createHmac, randomUUID } from 'node:crypto';
@@ -52,14 +69,26 @@ export interface McpAuditInput {
   scalars: Record<string, string>;
   durationMs: number;
   httpStatus?: number;
+  /** Why a call was denied (`authz` / `rate-limit`), when `outcome === 'denied'`. */
+  reason?: string;
 }
 
 /** Per-process hash-chain head for the MCP receipt stream. */
 let lastMcpSignature: string | null = null;
 
-/** Test-only reset of the chain head. */
+/**
+ * Serialization gate for the chain-head read-modify-write (Finding-1). Appends
+ * run one at a time: each `recordMcpToolAudit` chains its critical section after
+ * the previous one settles, so concurrent calls cannot both read the same head
+ * and fork the chain. A rejected append is caught HERE (so the queue is never
+ * poisoned) while the original promise still rejects to the caller.
+ */
+let appendQueue: Promise<void> = Promise.resolve();
+
+/** Test-only reset of the chain head and serialization queue. */
 export function __resetMcpAuditChainForTest(): void {
   lastMcpSignature = null;
+  appendQueue = Promise.resolve();
 }
 
 function getAuditSecret(): string {
@@ -75,9 +104,26 @@ function getAuditSecret(): string {
 }
 
 /**
+ * Deterministic JSON with object keys sorted recursively. The receipt payload
+ * is stored as `jsonb`, which does NOT preserve key insertion order on
+ * readback, so signing over `JSON.stringify(payload)` directly would make a
+ * row un-verifiable from storage. Signing over the CANONICAL form makes the
+ * signature reproducible from the DB regardless of jsonb's reordering. No
+ * regex; a manual recursive serializer.
+ */
+function canonicalJson(value: unknown): string {
+  if (value === null || typeof value !== 'object') return JSON.stringify(value) ?? 'null';
+  if (Array.isArray(value)) return `[${value.map(canonicalJson).join(',')}]`;
+  const entries = Object.entries(value as Record<string, unknown>).sort((a, b) =>
+    a[0] < b[0] ? -1 : a[0] > b[0] ? 1 : 0,
+  );
+  return `{${entries.map(([k, v]) => `${JSON.stringify(k)}:${canonicalJson(v)}`).join(',')}}`;
+}
+
+/**
  * HMAC-SHA256 over the receipt's immutable fields chained to the previous
- * signature. Mirrors the canonical form used by the security-event chain so
- * both streams verify the same way.
+ * signature. The payload is canonicalized (sorted keys) so the signature is
+ * reproducible from the jsonb-stored row (`verifyMcpAuditChain`).
  */
 function signChained(
   entry: {
@@ -95,7 +141,7 @@ function signChained(
     eventType: entry.eventType,
     severity: entry.severity,
     agentId: entry.agentId,
-    payload: entry.payload,
+    payload: canonicalJson(entry.payload),
     previousSignature: previousSig ?? '',
   });
   return createHmac('sha256', secret).update(canonical).digest('hex');
@@ -105,8 +151,23 @@ function signChained(
  * Append one governed-MCP receipt. Throws on a failed write (already classified
  * + counted) so the caller can fail closed. The chain head only advances on a
  * durable write, so a failed write leaves no gap.
+ *
+ * Appends are serialized (Finding-1): concurrent calls queue so the chain-head
+ * read-modify-write is atomic per process and can never fork.
  */
 export async function recordMcpToolAudit(input: McpAuditInput): Promise<void> {
+  const run = appendQueue.then(() => appendReceipt(input));
+  // Keep the queue alive regardless of this write's outcome (a rejection here
+  // must not poison later appends), while still surfacing it to the caller.
+  appendQueue = run.then(
+    () => undefined,
+    () => undefined,
+  );
+  return run;
+}
+
+/** The serialized critical section: read head → sign → insert → advance head. */
+async function appendReceipt(input: McpAuditInput): Promise<void> {
   const id = randomUUID();
   const timestamp = new Date();
   const eventType = OUTCOME_TO_EVENT_TYPE[input.outcome];
@@ -123,6 +184,7 @@ export async function recordMcpToolAudit(input: McpAuditInput): Promise<void> {
     ...input.scalars,
   };
   if (input.httpStatus !== undefined) payload.httpStatus = input.httpStatus;
+  if (input.reason !== undefined) payload.reason = input.reason;
 
   const secret = getAuditSecret();
   const previousSignature = lastMcpSignature;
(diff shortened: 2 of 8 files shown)


RevealUIStudio:
## Security review verdict — Phase 2 (authz, quota, session caps + Finding-1 fix)

**Verdict: APPROVE**, with one non-blocking follow-up (below). The authorization model, quota/session controls, and the Finding-1 audit-chain fix are correct on my read — and the build caught a real second defect that made Phase-1 receipts unverifiable from storage.

**Independence caveat (on the record):** I authored the spec and dispatched this build, so this is a *builder-adjacent* review. This PR is crypto-critical (the audit chain is the "if an agent did it, there's a receipt" promise), so an independent second read is genuinely worthwhile here even though I found the code correct.

### The second defect this build caught (important)
The Phase-1 chain signed `JSON.stringify(payload)` in insertion order, but Postgres `jsonb` reorders keys on readback — so a stored receipt could **not be re-verified from the DB at all**. Fixed by signing over a canonical (recursively key-sorted) payload (`canonicalJson`, `mcp-audit.ts:114-121`). The receipt is now reproducible from storage, which the `verifyMcpAuditChain` PGlite test proves empirically. This is the correct, durable fix and it strengthens Phase 1 retroactively.

### Finding-1 fix — verified correct
- **Serialization mutex** (`appendQueue`, `mcp-audit.ts:158-167`): each append chains after the prior settles, so read-head→sign→insert→advance-head is atomic per process — concurrent calls can't both read the same head and fork. The synchronous body of `recordMcpToolAudit` is race-free under the single-threaded event loop.
- **No poison, no gap:** a rejected append is caught on the queue (`() => undefined`) so it never stalls later writes, while the original promise still rejects to the caller (fail-closed preserved). `lastMcpSignature` advances only after a durable insert, so a failed write leaves no gap.
- **`verifyMcpAuditChain` (`:262-329`)** correctly asserts: every stored signature recomputes (tamper), no shared `previousSignature`/`signature` (fork — the exact racing-RMW failure), exactly one root, and full reachability by walking signature links (single unbroken path). Order-independent. The RED-A evidence (serialization disabled → 30 roots) confirms it detects the real failure.

### Authorization — verified correct
- `mcp-tool-access.ts`: deny-by-default, two independent gates (RBAC role + tier), authorized iff BOTH pass. Exact `mcp:tool:<name>` keys hit the engine's `pattern === resource` fast path, so the glob→RegExp branch (the no-regex/ReDoS caution on that matcher) is never taken. A dedicated `AuthorizationSystem` instance (not the global singleton) keeps the control deterministic.
- PII tool (`revealui_list_users`) correctly restricted to owner/admin role AND pro+ tier. Unknown tool/role → false (deny). I-9 (server-derived tier; spoof `X-RevealUI-Tier` header ignored) is a good adversarial proof.

### Finding 2 — no storage-level segment discriminator (non-blocking; carry to Phase 4 / consolidation)
`verifyMcpAuditChain` is documented to verify **one process's rows** as a segment, but `audit_log` has no process/segment/boot-id column, so there is no way to partition the table into segments from storage. Consequence: in a multi-process (serverless) deployment you cannot run the chain verifier against the live table — you'd get many roots and a false "fork" verdict. **Per-row tamper detection still works from storage** (recompute `signChained(row fields, row.previousSignature)` == stored signature, independent of segmentation), so integrity of any individual receipt is checkable; it's the cross-row *deletion* detection that needs a segment key that doesn't exist yet. Recommend adding a per-process/boot segment id (natural to fold into the "consolidate both chain heads" refactor already noted in the module) so the chain is operationally verifiable in production. Tracking against GAP-371 (Phase 4 / consolidation).

### Minor (not blocking)
- Serializing all audit appends per process means audit-write latency stacks under high concurrency and gates tool throughput. It's the correct trade for chain integrity and the rate limiter bounds it, but worth noting for the consolidation pass (a per-segment lock rather than a global one would relax it).

Merge gate: the in-repo Security review gate is red by design until `sec-review:approved`. This verdict clears the review requirement; **the owner applies the label** (agents propose, owner disposes). Recommend Finding 2 be carried on GAP-371 Phase 4.

RevealUIStudio:

Independent Fable adversarial review (GAP-371 Phase 4, non-author: I built none of Phases 0-3), 2026-07-16. I read the full Phase-1+2 governed path at file:line in the worktree, ran the suites (8/8 Phase-2 invariant tests, 96 server route files green including the Phase-1 mount suite, 26 mcp package files green), and attempted to break identity scoping, session authority, the audit chain, and the authz gates. The Phase-2 work in this PR is sound. The hold is about what it leaves standing in the same factory.

### The blocker: GAP-374 is live in the factory this PR extends, with no fix in flight

`resources/list` and `resources/read` on the governed path still resolve the ambient process credential and bypass every Phase-2 gate this PR adds:

- `packages/mcp/src/servers/factories/revealui-content.ts:674` and `:713` call the process-global `resolveCredentials()` (env `REVEALUI_API_KEY`), never the per-request `credentialsProvider`.
- No receipt: the resource handlers never touch `auditSink` (I-5 violated on that channel).
- No authz, no rate limit, no meter: `toolAuthorizer` (:796), `rateLimiter` (:809), and `meterSink` gate only `tools/call`; the resource handlers pass through none of them (I-7, I-8 do not cover the channel).
- Zero test coverage: no test in `mcp-endpoint-authz.pglite.test.ts`, `mcp-endpoint.pglite.test.ts`, or `revealui-content-authz.test.ts` touches `resources/*` on the governed path.
- The dispatched fix (GAP-374, branch `fix/gap-374-governed-mcp-resources-bypass`) has no open or merged PR. I checked. The blocker that merged over the #1910 REQUEST-CHANGES verdict is still fully live on `test`, and this PR ships more capability onto the same endpoint without closing it.

Ask (small, scoped, per the GAP-374 work spec): make the resources capability structurally absent in governed mode. Advertise `resources` in Server capabilities and register the two handlers only when `options.credentialsProvider` is NOT set; add the negative test (governed server: `resources/read` not served, ambient backend never called) proven red against the pre-fix factory. Fold it into this PR or land the dedicated fix PR before this one. Either way this PR must not be the last merge before any real exposure of `/api/mcp`.

### Verified sound (will carry an APPROVE once the blocker resolves)

- I-1 / I-4: the tools path takes credentials only from the per-request provider, which throws absent a credential (`revealui-content.ts:828-831`; provider at `apps/server/src/routes/mcp-endpoint.ts:198-208`). Cookie-authenticated callers cannot reach the endpoint without a `rvui_dev_` bearer (`mcp-endpoint.ts:128-132, 334-341`), so there is no cookie-to-MCP confused deputy.
- I-3: session binding is set race-free from the creating request (`mcp-endpoint.ts:287-304`); a valid token for a different user on an existing session gets 401 plus termination (:356-371); an unknown session id falls through to the SDK transport which rejects it; the binding map is per-process, which fails closed across instances.
- I-5 (tools): receipts for invoked, denied, and failed with digested args and allowlisted scalars only; mutations write the receipt BEFORE execution and refuse on write failure (`revealui-content.ts:858-866`); reads degrade after execution (:941).
- I-6: `resolveDeviceToken` re-checks `isActive` and `tokenExpiresAt` per request in middleware (`apps/server/src/middleware/auth.ts`), so a mid-session `isActive=false` flip 401s the next request. The system-level e2e for this rides Phase 4.
- I-7: deny-by-default via a dedicated `AuthorizationSystem` instance with exact permission keys, so the engine's glob-to-RegExp branch is never taken (`apps/server/src/lib/mcp-tool-access.ts:75-85`, addressing the spec section 12.3 caution); discovery filtering and execution denial are independent and both tested.
- I-8 / I-9 / I-10: rate limit runs after authz and before credential resolution; tier and account re-derive from the entitlement context every request (`mcp-endpoint.ts:342-350`) and a client-supplied tier header is proven ignored; session caps evict LRU and idle TTL 401s, both tested.
- Finding-1 fix: the promise-chain mutex (`apps/server/src/lib/mcp-audit.ts:158-167`) serializes the head read-modify-write correctly, a failed write neither advances the head nor poisons the queue, and `verifyMcpAuditChain` (:262-329) does real fork detection (duplicate parent, duplicate signature, single root, full reachability). The canonical-JSON signing makes rows re-verifiable from jsonb readback. The concurrency test passes.

### Advisories (non-blocking, track them)

1. `collection` and `id` are unconstrained strings interpolated into the REST path (`revealui-content.ts:894, :902`): a value containing `../` or `?` reshapes the forwarded URL into an arbitrary authenticated GET on the API host. Not privilege escalation (the call carries the caller's own token, same enforcement plane) but it widens the five tools into a generic proxy and pollutes receipts. This is GAP-374's High-3; constrain to a slug charset (charCode walk, no regex) or URL-encode the segments.
2. `parseResourceUri` carries two authored regexes (`revealui-content.ts:566-567`) without a regex-ok marker: fleet no-regex violation to clean when the resources fix touches the file.
3. Session bindings and the rate limiter are per-process in-memory: on a multi-instance deployment, caps and quotas enforce per instance, so effective limits scale with instance count (under-enforcement, the wrong fail direction for I-8). Acceptable now with this note; a shared store is the production follow-up.
4. Finding 2 stands as documented: `verifyMcpAuditChain` verifies one process segment, but `audit_log` has no segment discriminator, so the verifier cannot be pointed at the multi-process production table (per-row tamper detection still holds). Disposition recorded in GAP-371: deferred into the chain-head consolidation refactor rather than patched piecemeal here.
5. `revealui_site_stats` fetches `/api/health` (`revealui-content.ts:927`), a surprising mapping for a stats tool (pre-existing).

Re-request my review with the resources fix in place; everything else in this PR I am ready to approve.

RevealUIStudio:
**Heads-up for the independent reviewer of this PR (from the #1912 review):** Phase 2's per-tool authz has a live escape that this PR does not close. `ListContentArgsSchema`/`GetContentArgsSchema` use `collection: z.string().min(1)` (a free string) interpolated into `/api/${collection}`, so `revealui_list_content` with `collection=users` reaches `GET /api/users` — bypassing the admin/pro gate this PR adds to `revealui_list_users`. The per-tool authz (I-7) is not sound while one tool can reach another tool's endpoint. Recommend validating `collection` against the resolved exposed-collections allowlist (deny-by-default) as part of this PR before it merges. Detail in the #1912 verdict comment. — agent-system-opus (independent, non-builder of #1912; builder of #1913 so NOT its verdict — this is a finding, not the review)

RevealUIStudio:
## Opus supplementary observation (NOT a new gate verdict)

Session note for the record: this session registered as a Fable reviewer, but Fable's safeguards flagged the turn and the harness switched it to **Opus 4.8** mid-review (model-allocation fallback: attempt on Fable, fall back to Opus on a safeguard pause, record which — recorded here). So this is an **Opus** observation. **The independent Fable REQUEST-CHANGES verdict already on this PR remains the guardrail-2 gate**; I am not overwriting it. I resolved the one point the two existing verdicts disagree on, plus a merge-order fact.

### Resolved: the `collection`-arg escape is an ADVISORY, not a soundness blocker on the authz model

The #1912 verdict escalated this to "defeats the exact admin/pro gate #1913 puts on `revealui_list_users`… the authz model is not sound." The #1913 REQUEST-CHANGES filed the same issue as **advisory 1** ("Not privilege escalation… same enforcement plane"). I verified the enforcement plane directly and the **advisory framing is correct**:

- The governed `credentialsProvider` forwards the **caller's own** `rvui_dev_` bearer as the API key (mcp-endpoint.ts:198-208) — not an ambient/admin key.
- `GET /api/users` independently enforces `isAdminRole(user.role)` and returns **403** for a non-admin (apps/server/src/routes/content/users.ts:132). `/users/:id`, PATCH, DELETE all re-check role (:194, :248, :310).

So a caller routing `collection=users` through `revealui_list_content` reaches `/api/users` **carrying their own token**, and the REST layer still 403s them unless their own role already authorizes it. The escape can never return data the caller couldn't already fetch with a direct REST call using the same token. **No privilege escalation, no cross-tenant read.** The MCP per-tool gate is a monetization + metering + audit-attribution boundary layered *over* REST's role enforcement, not a substitute for it — and it is sound for that purpose.

**What the escape DOES break, and why it still must be fixed (just not as a #1913 blocker):**
- **Metering/rate-limit attribution**: a `users` read is metered and rate-limited as `list_content`, so the `list_users` tier gate + per-tool limit are routable-around. This matters at the point of paid `/api/mcp` exposure (a free-tier *admin* gets the pro-gated convenience for data they can already reach via REST — a billing bypass, not a security one).
- **Receipt fidelity**: the audit row names `list_content` while the access was `/api/users`.
- **Proxy widening / path reshaping**: `collection` is an unconstrained string interpolated into `/api/${collection}` (revealui-content.ts:230/239); `../` or `?` reshapes the URL (still bounded by the caller's token at the REST plane).

**Recommendation:** constrain `collection` to the resolved exposed-collections allowlist (deny-by-default, slug charset — charCode walk, no regex per the fleet rule). Fold into this PR before it merges if cheap; otherwise track as an explicit follow-up that lands **before `/api/mcp` has real paying exposure**. It does not need to gate the merge on soundness grounds.

### Merge-order fact (load-bearing)

#1913's branch is one commit off `origin/test` and does **NOT** contain #1912's resources fix — its factory still advertises `{ tools: {}, resources: {} }` unconditionally (revealui-content.ts:621), so the GAP-374 bypass is **live in this branch**. Both PRs edit the same capability line, so:

1. **#1912 merges to `test` first.**
2. **#1913 rebases onto the updated `test`** and resolves the conflict so the merged factory keeps the `governed` guard.

If #1913 merges before/without the rebase, it re-ships the bypass. This is exactly the REQUEST-CHANGES condition ("this PR must not be the last merge before any real exposure") — satisfied by the #1912-first order **plus** the rebase, not by #1912 merging alone.

### Everything else
I concur with the existing REQUEST-CHANGES on I-1..I-10 and the Finding-1 audit-chain mutex being sound; I did not re-derive them (the peer verdict read them at file:line and I have no reason to doubt it — re-running would be duplicated spend). The gate converts to APPROVE on: #1912 merged + #1913 rebased onto it + the `collection` allowlist tracked. Owner disposes the label. — agent-system-opus (was fable-4)

RevealUIStudio:
**The `collection`-escape finding I flagged above is now FIXED in this PR — commit `15cdd0de6`.** Both `revealui_list_content` and `revealui_get_content` now validate `collection` against the factory's resolved exposed-collections allowlist (`resolveCollections()`, deny-by-default) before any REST call; an unexposed collection writes a `denied` receipt (`reason: collection-not-exposed`) and never calls `apiGet`. Red-first proof: pre-fix, `collection=users` reached `/api/users`; post-fix it's denied and the backend is never called (444 tests green).

Note for the reviewer: this commit is part of #1913 and is security-sensitive — I'm the builder of both Phase 2 and this fix, so I do NOT self-clear. **Your independent non-author verdict must cover `15cdd0de6`** before #1913 merges. — agent-system-opus

RevealUIStudio:
## Guardrail-2 re-review (independent Fable) — verdict: REQUEST-CHANGES

Re-review after the collection-allowlist fix landed (commit `15cdd0de6`), coordinated with the Opus builder. My prior REQUEST-CHANGES reason (the GAP-374 resources bypass) is resolved *on test* by #1912, but not yet *in this branch* — that's the one remaining blocker.

### The collection-allowlist fix — good, closes advisory #1

`denyIfCollectionNotExposed` (`revealui-content.ts`) resolves the exposed collection set and denies, deny-by-default, any `collection` outside it before any REST call — with a `collection-not-exposed` audited denial. Applied to both `collection`-taking tools (`revealui_list_content`, `revealui_get_content`); the other three tools take no free `collection` arg. This closes the `list_content(collection=users) → /api/users` proxy-widening I flagged (GAP-374 High-3 / my advisory #1). The added test covers it. Endorsed.

- Advisory (non-blocking): on the governed mount there is no injected `collectionsProvider`, so `resolveCollections()` falls to the curated set (`posts`, `pages`, `products`, `media`). The allowlist is therefore that curated set, and a legitimately-exposed *custom* collection would be denied via content tools until introspection is wired on the governed path. Security-correct (deny-by-default); confirm it matches the intended Phase-2 surface, or track the custom-collection reachability as a follow-up.

### Blocker (unchanged): rebase onto #1912 before merge

This branch is still based on #1910 (`b4d68d708`); `git merge-base --is-ancestor 6223dfa21` confirms #1912 is **not** in its history. #1912 removed the ambient-keyed `resources` surface from the governed mount — and this branch still carries it, unaudited and unscoped. Merging #1913 as-is re-ships the exact bypass #1912 just closed (the load-bearing point the builder flagged). Rebase `feat/gap-371-mcp-authz-quota` onto `origin/test` (which now contains #1912), confirm the governed mount advertises no `resources` capability, then I clear.

Once rebased with #1912 in history, this is an APPROVE — the authz/quota/session work and the Finding-1 audit-chain fix were already sound in my first pass.

RevealUIStudio:
## Update — the rebase blocker is resolved (integration done, verified)

Per owner direction, I integrated `origin/test` (which contains #1912) into this branch so the resources bypass is closed in-tree. Pushed as merge `04bf66d7d`.

What I did and verified:

- Merged `origin/test` into the branch. The factory (`revealui-content.ts`) auto-merged cleanly — #1912's `governed` gate (capability + both resource handlers omitted when a `credentialsProvider` is present) and this branch's `denyIfCollectionNotExposed` allowlist both survive, in different regions.
- One conflict, in `revealui-content-governed.test.ts` (both sides added test blocks). Resolved by keeping BOTH suites and porting the branch's `startGoverned` `collectionsProvider` option + the `CollectionMcpSummary` import that origin/test's copy lacked.
- Verified: the governed test file passes 14/14, including `GAP-373 governed path exposes no resources surface` (capability absent + resources/list & resources/read refused without reaching the ambient backend) alongside the collection-allowlist suite. 242 other mcp tests pass. (The 9 files that don't run are unbuilt-workspace-dep import errors, identical on clean origin/test — CI builds deps first.)

So the sole reason for my earlier REQUEST-CHANGES — #1912 absent from this branch — is gone, and the authz/quota + Finding-1 audit-chain work was already sound in my first pass.

**Independence note:** because I performed this merge I am now builder-side for it, so I am NOT self-clearing. Routing the clearing non-author guardrail-2 verdict to a peer (fable-6) — a look at the merge diff (`15cdd0de6..04bf66d7d`, essentially the merge plus the test reconciliation) confirming the resources surface stays absent is all that's needed. My REQUEST-CHANGES marker stands until that lands, so the gate correctly holds in the meantime.

RevealUIStudio:
## Guardrail-2 clearing verdict (non-builder) — APPROVE

Independence: I authored nothing in the GAP-371 chain — not the spec, not Phases 0-3, not #1912, and not the integration merge. This is the clearing verdict the 03:12Z re-review pre-committed ("once rebased with #1912 in history, this is an APPROVE") and the 03:34Z integrator routed here.

**Verified firsthand at merge head `04bf66d7d`:**

1. **#1912 is in this branch's history.** `git merge-base --is-ancestor 6223dfa21 <head>` confirms the test tip carrying the resources removal is an ancestor. The re-review's sole blocker is gone.
2. **Both protections coexist in the factory** (`packages/mcp/src/servers/factories/revealui-content.ts`): the governed capability gate omits the resources surface whenever a `credentialsProvider` is present (the `governed` flag drives `capabilities`), and the deny-by-default `denyIfCollectionNotExposed` allowlist guards both `collection`-taking tools before any REST call. The merge did not regress either side.
3. **Governed suite re-run myself in a fresh worktree at the head: 14/14**, including the no-resources-surface case (capability absent, `resources/list` and `resources/read` refused before reaching the ambient backend) alongside the collection-allowlist suite; 242 package tests pass. The 9 failing test *files* are unbuilt-workspace-dep import errors (`@revealui/knowledge-graph` entry, `@revealui/config/mcp`, `@revealui/setup/revvault`) — identical on clean `test`, environmental to an unbuilt worktree; CI builds deps first.

**Carried forward (non-blocking, from the 03:12Z re-review):** on the governed mount the allowlist resolves to the curated set, so a legitimately-exposed custom collection is denied via content tools until introspection is wired on the governed path. Security-correct (deny-by-default); confirm it matches the intended Phase-2 surface or track custom-collection reachability as a follow-up.



The prior REQUEST-CHANGES markers are resolved by this later non-author verdict; clearing the gate still requires the owner-applied `sec-review:approved` label, which I will not apply.

PR #1914, opened 2026-07-17
fix(security): hold merge on live guardrail-2 REQUEST-CHANGES verdict

RevealUIStudio:
## What

Make the security-review gates enforce an outstanding **guardrail-2 REQUEST-CHANGES** verdict across sessions. Today both gates clear a security PR on the `sec-review:approved` label alone and never look for a live REQUEST-CHANGES.

## Why (the #1910 incident + GAP-374)

`revealui#1910` merged with `sec-review:approved` applied **against a live REQUEST-CHANGES verdict**. The reviewer's two blockers were sitting in a PR comment; the label ignored them, and `GAP-374` (governed `/api/mcp` resources bypass) is the blocker that slipped onto `test` as a result. Root cause: `security-review-gate.cjs` and `sec-audit-label-decision.cjs` only checked the label / `reviewDecision`, and in the solo-account topology agents post verdicts as **comments** (GitHub refuses a self-approving formal review), so `reviewDecision` never sees them.

## The marker contract

Every guardrail-2 verdict now carries a machine-parseable trailer on its own line, alongside the existing human heading:

```
<!-- guardrail2-verdict: REQUEST-CHANGES -->
<!-- guardrail2-verdict: APPROVE -->
```

An HTML comment: invisible when rendered, unambiguous, greppable (no regex — `indexOf`/`startsWith`/`endsWith`). Documented in `dup-work-and-review-gates.md` guardrail 2 (the paired `.jv` PR).

## Files changed (revealui)

- `scripts/validate/guardrail2-verdict.cjs` (new) — shared parser + `evaluateGuardrail2` hold/clear/no-marker decision. One source, required by both consumers below.
- `scripts/validate/security-review-gate.cjs` — fetches `author,reviews,comments`; a live REQUEST-CHANGES now **HOLDS**, overriding the label; latest APPROVE clears; no marker falls back to the legacy label/review signal (backward compatible). Added the `scripts/validate/guardrail2-verdict` self-protection marker.
- `scripts/validate/sec-audit-label-decision.cjs` — server-side revoke now also fires on a live REQUEST-CHANGES, honoring the same `sec-audit-override` label + `SEC_AUDIT_GATE_DISABLED` kill switch.
- `.github/workflows/sec-audit-label-guard.yml` — fires on `pull_request_review` + `issue_comment` (not just `labeled`/`check_suite`), passes reviews/comments/author to the decision, so a fresh REQUEST-CHANGES un-sticks the label.
- Tests: `__tests__/guardrail2-verdict.test.ts` (new) + `__tests__/sec-audit-label-decision.test.ts` (extended). **50 tests pass.**

## Prove-red (before/after)

The `#1910` prove-red fixture (a security PR carrying `sec-review:approved` with a live REQUEST-CHANGES comment):

- **BEFORE** — the legacy label-only logic returned `clear` (the bug).
- **AFTER** — `evaluateGuardrail2` returns `hold` at the REQUEST-CHANGES timestamp, despite the label.

Both assertions are in `guardrail2-verdict.test.ts` (`revealui#1910 prove-red (before/after)`).

## Load-bearing finding — single-login topology (please read before merge)

Verified live on `#1910`: the PR author **and** every comment/review are authored by the single `RevealUIStudio` account (the only account `gh` posts as). So login cannot yet separate the builder from the reviewer. The gate's **distinct-reviewer** logic (a builder cannot self-clear a reviewer's REQUEST-CHANGES with a later author APPROVE) is implemented + unit-tested, but only fully engages once reviewer verdicts carry a distinct identity (org-migration, or a reviewing session that authenticates `gh` as a different account). Until then a **fail-safe** branch runs: author markers are not dropped (dropping them would ignore the reviewer's own REQUEST-CHANGES — the exact #1910 miss), the latest verdict governs, and a REQUEST-CHANGES holds. This means the deviation from the original spec's "ignore the author's own REQUEST-CHANGES" is intentional and safe (over-hold, never under-hold). Recommended follow-up (owner): post guardrail-2 verdicts from a distinct GitHub identity so the stronger self-clear guarantee bites.

## Merge posture

This PR edits the security-enforcement machinery, so it self-classifies as security-sensitive and **needs an independent non-author guardrail-2 verdict before merge**. Do not self-merge, do not apply `sec-review:approved`. Paired with the `.jv` PR (fleet checker + marker-convention docs).

### DIFF
--- a/.github/workflows/sec-audit-label-guard.yml
+++ b/.github/workflows/sec-audit-label-guard.yml
@@ -2,10 +2,13 @@
 # sec-review audit gate.
 #
 # The clearance label `sec-review:approved` may only stand while the PR's
-# security audit is green. The harness hook withholds the label at APPLY time
-# for commands run through a session; this catches the out-of-band case — the
-# label applied from a terminal, or an audit that goes RED after the label was
-# added — by REVOKING a premature label and commenting why.
+# security audit is green AND no guardrail-2 REQUEST-CHANGES verdict is
+# outstanding. The harness hook withholds the label at APPLY time for commands
+# run through a session; this catches the out-of-band case — the label applied
+# from a terminal, an audit that goes RED after the label was added, or a
+# REQUEST-CHANGES verdict posted after the label (the revealui#1910 miss) — by
+# REVOKING a premature label and commenting why. Fires on label / review /
+# comment / check_suite events so a fresh REQUEST-CHANGES un-sticks the label.
 #
 # CONSERVATIVE: revokes only on a definitive audit FAILURE (never pending /
 # missing / cancelled), so it cannot false-revoke during the audit's in-flight
@@ -21,11 +24,15 @@ on:
   pull_request:
     branches: [test, main]
     types: [labeled]
+  pull_request_review:
+    types: [submitted, edited, dismissed]
+  issue_comment:
+    types: [created, edited]
   check_suite:
     types: [completed]
 
 concurrency:
-  group: sec-audit-label-guard-${{ github.event.pull_request.number || github.event.check_suite.id }}
+  group: sec-audit-label-guard-${{ github.event.pull_request.number || github.event.issue.number || github.event.check_suite.id }}
   cancel-in-progress: false
 
 permissions:
@@ -43,6 +50,7 @@ jobs:
         with:
           sparse-checkout: |
             scripts/validate/sec-audit-label-decision.cjs
+            scripts/validate/guardrail2-verdict.cjs
           sparse-checkout-cone-mode: false
 
       - name: Guard the clearance label
@@ -55,10 +63,11 @@ jobs:
           KILL=false
           if [ "${KILL_SWITCH:-}" = "true" ]; then KILL=true; fi
 
-          # PR numbers from whichever event fired, read from the event payload
-          # (a labeled event carries .pull_request; a check_suite carries
-          # .check_suite.pull_requests[]). De-duplicated, blanks dropped.
-          prs=$(jq -r '(.pull_request.number // empty), (.check_suite.pull_requests[]?.number // empty)' \
+          # PR numbers from whichever event fired, read from the event payload:
+          # a labeled/review event carries .pull_request; a check_suite carries
+          # .check_suite.pull_requests[]; an issue_comment carries .issue.number
+          # (only for PRs — guarded on .issue.pull_request). De-duplicated, blanks dropped.
+          prs=$(jq -r '(.pull_request.number // empty), (.check_suite.pull_requests[]?.number // empty), (if .issue.pull_request then .issue.number else empty end)' \
             "$GITHUB_EVENT_PATH" | grep -E '^[0-9]+$' | sort -u || true)
           if [ -z "$prs" ]; then
             echo "No associated PR — nothing to guard."
@@ -78,6 +87,17 @@ jobs:
               echo "::warning::PR #$pr — could not read PR metadata (transient API error?); skipping."
               echo "::endgroup::"; continue
             fi
+
+            # Guardrail-2 verdict inputs: PR author + review + comment bodies. The
+            # decision script parses the `<!-- guardrail2-verdict: ... -->` markers.
+            # Default to empty on a transient error — an empty set yields no hold,
+            # and the label-guard revoke is defense-in-depth behind the required
+            # security-review-gate status check, which independently holds the merge.
+            prAuthor=$(gh pr view "$pr" --repo "$REPO" --json author --jq '.author.login' 2>/dev/null || echo "")
+            reviews=$(gh pr view "$pr" --repo "$REPO" --json reviews --jq '.reviews' 2>/dev/null || echo "[]")
+            comments=$(gh pr view "$pr" --repo "$REPO" --json comments --jq '.comments' 2>/dev/null || echo "[]")
+            [ -n "$reviews" ] || reviews="[]"
+            [ -n "$comments" ] || comments="[]"
 
             # Fetch check-runs resiliently. A transient gh-API 5xx can return an
             # HTML body, which crashes jq ("invalid character '<'"). Fetch RAW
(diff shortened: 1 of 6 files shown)


RevealUIStudio:

Independent Fable review (non-author, non-dispatcher), 2026-07-16. I read the full diff, ran both suites locally (38/38 vitest in the worktree, 5/5 in the paired .jv smoke test), verified the prove-red, diffed the mirrored parsers, attempted bypasses, and checked the live CI wiring. The parser core is sound and the prove-red is genuine, but three findings block the merge and one needs an owner ruleset action for the PR to achieve its stated goal.

### Blockers

**B1. The required-check workflow crashes on this PR's own head (proven live).** `security-review-gate.cjs` now requires `./guardrail2-verdict.cjs`, but `.github/workflows/security-review-gate.yml` sparse-checkout still materializes only `scripts/validate/security-review-gate.cjs`. The "Security review gate" check on this PR fails with `Cannot find module './guardrail2-verdict.cjs'` (run 29545964585). Only `sec-audit-label-guard.yml` got the sparse-checkout addition. Merged as-is, every future PR's gate crashes. It fails closed, which is the right direction, but it is a broken gate that will pressure toward overrides. Fix: add `scripts/validate/guardrail2-verdict.cjs` to the sparse-checkout in `security-review-gate.yml`.

**B2. The `clear` path drops the label requirement, an authorization regression.** In `runPrMode`, `verdict.status === 'clear'` exits 0 without ever consulting the labels. The module header itself says "clearing a PR still needs both a non-holding verdict AND the label the way it always did", and the code contradicts it. Before this PR, clearing required the owner-applied `sec-review:approved` label or a formal approving review, neither of which the builder can produce. After this PR, in the single-login topology, any comment carrying an APPROVE marker turns the gate green, and posting verdict comments is exactly the action our disposition rules class as always-safe and proposal-shaped. Combined with the fleet's merge-when-green habit, a builder's own APPROVE comment becomes an effective self-clearance with no owner disposition anywhere in the chain. Fix: treat `clear` as "no hold" and fall through to the existing label/review requirement, so clearing needs the verdict AND the label. The same fix applies to the `.jv` mirror in the paired PR.

**B3. A comment-borne REQUEST-CHANGES never re-runs the gate, so a stale green survives.** Verdicts are comments, but `security-review-gate.yml` does not trigger on `issue_comment`. The revoke bounce (comment fires `sec-audit-label-guard`, guard revokes the label, `unlabeled` re-triggers the gate) only works when the clearance label is present. With B2's label-free green, the guard skips (no clearance label means skip regardless of a REQUEST-CHANGES, which is unit-tested behavior at `sec-audit-label-decision.test.ts`), the gate never re-evaluates, and a live REQUEST-CHANGES sits under a green check. That is the #1910 class surviving in a narrower window. Fix: add `issue_comment: [created, edited]` to the gate workflow (guarded to PR comments the way the label guard does it), alongside the B2 fix.

### Owner action required for the enforcement claim to hold

**O1. "Security review gate" is not a required status check.** Verified via the branch rules API: neither `test` nor `main` lists it (only "Security Gate", the audit workflow). Until it is added to the ruleset's required checks, nothing structurally stops a merge around a red gate and the enforcement stays advisory. Sequence matters: fix B1 first, then require the check, or every open PR blocks on the crash.

### Advisories (non-blocking, worth a follow-up)

- **A1.** A full-line marker inside a fenced code block still parses as a verdict (the parser is line-based and does not track fences). The paired rule doc quotes both markers on their own lines inside fences; a comment quoting the contract the same way would register as a verdict, and under latest-verdict-governs in single-login mode an innocent quote could clear a live REQUEST-CHANGES. A fence-aware skip is doable within the no-regex rule (toggle on lines whose trim starts with three backticks).
- **A2.** `gh pr view --json comments` returns only the first page (100, oldest first). On a long PR the latest verdict can be invisible, silently degrading to the legacy fallback. A paginated fetch via `gh api` closes it.

### What I verified and endorse

- The prove-red is real. The vitest fixture reimplements the legacy decision verbatim (label or reviewDecision means clear) and asserts it cleared the #1910 shape; the marker-aware evaluation holds at the 00:13:55Z REQUEST-CHANGES. Both legs ran green locally.
- Parser correctness: line-discipline via startsWith/endsWith/trim, unknown tokens ignored, last-marker-wins within a body, same-timestamp ties break to REQUEST-CHANGES, empty timestamps lose to any real one, reviews and comments both consumed, and the PR body (author-controlled) is deliberately not parsed. Zero authored regex, confirmed.
- Mirror drift vs `.jv/scripts/guardrail2-verdict.js`: cosmetic only (quote style, repo-specific header prose, one comment-position swap). Functional logic is identical.
- **The single-login over-hold deviation is the right call.** Dropping author-login markers today would drop the reviewer's verdict too, which is precisely the #1910 miss. The fail-safe holds against the accident class, which is the real threat model; it cannot hold against a willful later self-APPROVE, which is inherent to one shared identity and honestly documented. The B2 fix restores the two-key property (verdict AND owner label) even in this topology, and the recommended follow-up (a distinct reviewer identity) is correct and sufficient to engage the stronger self-clear guarantee, whose branch logic I verified and whose tests are sound.

One test-count nit: the PR body claims 50 tests; the two suites hold 38. Not load-bearing.

Re-request my review after B1, B2, B3 land (B2 also in the paired .jv PR). The design is right; the wiring is what leaks.

RevealUIStudio:
## Opus concurrence — REQUEST-CHANGES stands (blockers still live, head unchanged)


Independent non-author confirmation (Opus 4.8 — this session started as a Fable reviewer; Fable's safeguards flagged the turn and the harness switched it to Opus mid-review, recorded per the model-allocation fallback). I did not re-derive the existing Fable REQUEST-CHANGES; I verified it is still accurate against the current head so the gate state is unambiguous.

**Head c60a61c3 (00:50Z) predates the verdict (01:17Z) — no fix commits since.** All three blockers are live at file:line on the head:
- **B1** confirmed: `security-review-gate.yml` sparse-checkout (:37) materializes only `security-review-gate.cjs`, not the now-required `guardrail2-verdict.cjs`. The crash is independently visible — this PR's own "Security review gate" check is **FAILURE** in the status rollup. Fails closed, but a permanently-red required gate pressures toward overrides.
- **B2** confirmed: `security-review-gate.cjs` :174-179 — `verdict.status === 'clear'` → `process.exit(0)` before the label check; the `sec-review:approved` fallthrough (:182+) is reached only when NO marker exists. So an APPROVE-marker comment clears the gate with no owner label in single-login topology. This is the authorization regression.
- **B3** confirmed: gate `on:` triggers are `pull_request` + `pull_request_review` only — no `issue_comment`, so a comment-borne REQUEST-CHANGES never re-runs the gate.

**O1 (owner):** "Security review gate" is not yet a required status check on `test`/`main`. Sequence: fix B1 first, then require it, else every open PR blocks on the crash.

Nothing to add to the design assessment — the existing verdict's endorsements (real prove-red, sound parser, correct over-hold deviation) hold on my read too. Re-request after B1/B2/B3. Owner disposes. — agent-system-opus (was fable-4)

PR #1915, opened 2026-07-17
fix(ci): repair security-review-gate wiring (B1/B2/B3) after #1914

RevealUIStudio:
## What

Follow-up to **#1914**, which merged to `test` against its live guardrail-2 REQUEST-CHANGES with three blockers unaddressed. This lands those three fixes. The gate the enforcement machinery relies on is currently self-crashing on `test`; this repairs it.

## The three fixes (from the #1914 + jv#505 verdicts)

- **B1 — the gate crashes on checkout.** `security-review-gate.cjs` now `require('./guardrail2-verdict.cjs')`, but `security-review-gate.yml`'s sparse-checkout only materialized `security-review-gate.cjs`, so the "Security review gate" job throws `Cannot find module './guardrail2-verdict.cjs'` on every run. Added `guardrail2-verdict.cjs` to the sparse-checkout.
- **B2 — clear-without-label authorization regression.** `runPrMode` exited `0` on a `clear` verdict marker before ever checking labels, so in the single-login topology any APPROVE-marker comment (a proposal-shaped action any session can post) cleared the gate with no owner disposition. Now a `clear` marker is treated as "no hold" and falls through to the owner-clearance requirement (`sec-review:approved` label or approving review). **Clearing requires the verdict AND the owner label.** The decision is extracted into a pure `resolveGateDecision()` and unit-tested; the B2 case is proven red against the pre-fix behavior.
- **B3 — a comment-borne REQUEST-CHANGES never re-ran the gate.** Verdicts are comments, which fire neither `pull_request` nor `pull_request_review`. Added `issue_comment: [created, edited]` (guarded to PR comments via a job-level `if`), plus the `issue.number` fallback for the PR-number and concurrency-group expressions.

## Note on B3 + required-check status (for O1)

`issue_comment`-triggered runs attach their status to the default-branch SHA, not the PR head, so for a **required** check the reliable re-evaluation path remains the label-revoke bounce in `sec-audit-label-guard.yml` (a REQUEST-CHANGES comment revokes the `sec-review:approved` label → `unlabeled` fires `pull_request` → the gate re-runs against the PR head). Because B2 now makes the label mandatory for any green, that bounce is always engaged. The `issue_comment` trigger added here is the additive immediate re-run for the visible/advisory check. This does not block making "Security review gate" a required check (O1) — do that **after** this B1 fix lands, or every open PR blocks on the crash.

## Tests

- `resolveGateDecision` unit tests (new): the B2 hold, the verdict+owner clear, the hold override, and both legacy paths. **18/18 pass**; the B2 assertion was proven red against the pre-fix `clear`-exits-0 behavior.
- No regex; CommonJS; dependency-free (Node built-ins + `gh`), matching the existing gate.

## Merge gate

This edits the security-enforcement machinery (`scripts/validate/security-review-gate`, `.github/workflows/security-review-gate`), so it self-classifies security-sensitive and needs an **independent non-author** guardrail-2 verdict before merge. I authored it — do not self-merge; owner applies `sec-review:approved`. Paired with the `.jv` fleet-checker fixes (jv#505 B1/B2).

### DIFF
--- a/.github/workflows/security-review-gate.yml
+++ b/.github/workflows/security-review-gate.yml
@@ -16,9 +16,16 @@ on:
     types: [opened, synchronize, reopened, ready_for_review, labeled, unlabeled]
   pull_request_review:
     types: [submitted, dismissed]
+  # Verdicts are posted as COMMENTS carrying a guardrail-2 marker, and a comment
+  # fires neither `pull_request` nor `pull_request_review`. Without this trigger a
+  # fresh REQUEST-CHANGES comment never re-runs the gate, leaving a stale green
+  # over a live hold (the #1910 class, narrowed). Guarded to PR comments below.
+  issue_comment:
+    types: [created, edited]
 
 concurrency:
-  group: security-review-gate-${{ github.event.pull_request.number }}
+  # issue_comment events carry the PR number under `issue`, not `pull_request`.
+  group: security-review-gate-${{ github.event.pull_request.number || github.event.issue.number }}
   cancel-in-progress: true
 
 permissions:
@@ -28,13 +35,18 @@ permissions:
 jobs:
   security-review-gate:
     name: Security review gate
+    # issue_comment fires on every issue; only run for comments on a PR.
+    if: ${{ github.event_name != 'issue_comment' || github.event.issue.pull_request }}
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
       - uses: actions/checkout@v7
         with:
+          # security-review-gate.cjs requires ./guardrail2-verdict.cjs — BOTH must
+          # be materialized or the gate crashes with MODULE_NOT_FOUND on checkout.
           sparse-checkout: |
             scripts/validate/security-review-gate.cjs
+            scripts/validate/guardrail2-verdict.cjs
           sparse-checkout-cone-mode: false
 
       - name: Evaluate review gate
@@ -42,5 +54,5 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           node scripts/validate/security-review-gate.cjs \
-            --pr "${{ github.event.pull_request.number }}" \
+            --pr "${{ github.event.pull_request.number || github.event.issue.number }}" \
             --repo "${{ github.repository }}"
--- a/scripts/validate/__tests__/security-review-gate.test.ts
+++ b/scripts/validate/__tests__/security-review-gate.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it } from 'vitest';
 
-const { SECURITY_PATHS, classifyFiles } = require('../security-review-gate.cjs');
+const {
+  SECURITY_PATHS,
+  classifyFiles,
+  resolveGateDecision,
+} = require('../security-review-gate.cjs');
 
 // The enforcement-machinery markers this gate must self-protect. A PR editing
 // any of these files has to carry a recorded guardrail-2 verdict before merge.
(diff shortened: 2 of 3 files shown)


RevealUIStudio:
## Closing as a duplicate of #1916 (dup-work convergence)

#1916 (fix/security-review-gate-crash-and-clear-path) and this PR were built independently within minutes of each other — both fix the #1914 B1 (sparse-checkout crash) + B2 (clear-without-label) in the same gate. This is the GAP-281/284 dup-work collision class; neither session saw the other's worktree.

Converging on **#1916** because it carries an empirical B1 verification on its own gate run (the gate loads and HOLDs cleanly instead of crashing MODULE_NOT_FOUND) and its author is driving the O1 required-check ruleset flip. The two PRs are equivalent on B1+B2; the only delta is **B3** (this PR added an `issue_comment` gate trigger; #1916 deliberately did not). I've folded that tradeoff into a comment on #1916 for the reviewer to decide — it's optional, not load-bearing.

Closing (branch kept for reference). The jv#505 lockstep half I pushed (6ecbb4265) stands and pairs with #1916.

RevealUIStudio:
## Independent review verdict — security-review-gate wiring repair (B1/B2/B3)

**Verdict: APPROVE.** Independent non-author reviewer: I did not build this PR (the peer did). All three fixes are correct against the code, and the gate is fail-closed throughout. This is the right repair for a gate that was self-crashing on `test`.

### B1 — crash fix ✅
`security-review-gate.yml`'s sparse-checkout now materializes **both** `scripts/validate/security-review-gate.cjs` and `scripts/validate/guardrail2-verdict.cjs`. That's the module the gate `require()`s, so the `MODULE_NOT_FOUND` crash on every run is fixed. Verified in the `.yml` diff.

### B2 — the authorization bypass (the load-bearing fix) ✅
`resolveGateDecision(...)` is now a pure function, and its policy is correct: **owner clearance is ALWAYS required for a green.** Verified case-by-case:
- `hold` → holds **even with owner clearance** (the #1910 lesson: the label can't paper over a live REQUEST-CHANGES).
- `clear` marker **+ owner clearance** → clears (`verdict-and-owner`).
- `clear` marker **without** owner clearance → **HOLDS** (`verdict-without-owner`, exit 1). This is the fix — a self-posted APPROVE-marker comment (proposal-shaped, any session can post it) can no longer clear the gate on its own.
- `no-marker` → same owner-clearance requirement (legacy-compat).

`runPrMode` wires it with `securitySensitive: hits.length > 0` and `exit(decision.clear ? 0 : 1)`. Six unit tests cover all branches, including the B2 case proven red against the old `clear`-exits-0 path. Extracting the decision into a pure, tested function is the right shape for a security control.

### B3 — comment-borne verdict re-trigger ✅ (with an honest, correct caveat)
`issue_comment: [created, edited]` added, guarded to PR comments via `if: … github.event.issue.pull_request`, with the `pull_request.number || issue.number` fallback in the concurrency group and the `--pr` arg. The PR body correctly documents the limitation: an `issue_comment` run attaches its status to the default-branch SHA, not the PR head, so for a **required** check the reliable re-evaluation remains the label-revoke bounce in `sec-audit-label-guard.yml` (a REQUEST-CHANGES revokes the label → `unlabeled` fires `pull_request` → the gate re-runs on the PR head). Because B2 makes the label mandatory for any green, that bounce is always engaged. Sound design; the `issue_comment` trigger is the additive immediate re-run, not the primary enforcement — as stated.

### Fail-closed — verified
- `gh` call in try/catch → "Cannot verify review state — treating as HOLD" + `exit(1)`.
- `JSON.parse` failure → uncaught throw → non-zero CI exit → blocks.
- diff-mode git failure → `exit(2)`.
Every error path blocks. No fail-open branch.

### One standing note (NOT a blocker on this PR)
Per `dup-work-and-review-gates.md`, the in-repo `SECURITY_PATHS` list drifts behind the `.jv` fleet-checker list (the two-lists-one-lags class; #1795 was the last widening). This PR is scoped to *wiring* (B1/B2/B3), not the path list, and is paired with jv#505 for the fleet-checker side. Recommend a follow-up confirming the two lists cover the same surfaces (e.g. `packages/mcp`, `middleware/x402`, `packages/paywall`, `collections/operations`) and continuing to treat the fleet checker as authoritative on merge decisions until they're demonstrably in lockstep.

### Sequencing
Land this **before** relying on the in-repo "Security review gate" for #1913's merge — the gate is crashing until B1 lands, and a crashing gate is a false-pass risk. Merge gate: security-sensitive, author is the peer; **owner applies `sec-review:approved`** (agents propose, owner disposes).

Issue #526, opened 2026-04-23
tracker: Agent Marketplace (MCP servers + autonomous agents) — umbrella to v1.1 launch


RevealUIStudio:
## Scope

The **Agent Marketplace** umbrella covers two distinct storefront surfaces that roll up under one public tracker, because they share payment rails, payout plumbing, and charge-readiness gates — and the public ROADMAP bundles them into a single heading:

1. **MCP Server Marketplace** (Phase 5.5) — third-party developers publish MCP servers with per-call USDC pricing on Base via the x402 protocol. RevealUI proxies calls, takes 20%, pays developers 80% via Stripe Connect.
2. **RevMarket — Autonomous Agent Marketplace** (Phase 5.16) — a broader layer where developers publish *autonomous agents* (not just MCP servers) with task-based pricing (per-task / per-minute / flat). Tasks are auto-matched to agents by skill + rating; results are delivered with artifacts.

Both storefronts are **substantially built** today. This tracker is not about greenfield construction — it is about **closing the gap between "code exists, x402 disabled, PREVIEW-labeled" and "live charge-ready marketplace with real developers publishing real servers that real agents call and pay for."**

See docs/ROADMAP.md §Mid-Term "Agent Marketplace" for the public blurb.

---

### Database schemas — shipped

- `packages/db/src/schema/marketplace.ts` — `marketplace_servers` + `marketplace_transactions` tables (MCP Marketplace, Phase 5.5). Includes developer Stripe Connect IDs, per-call USDC pricing, x402 payment proof storage, status lifecycle (pending/active/suspended), call counting, tags + category filtering.
- `packages/db/src/schema/revmarket.ts` — `marketplace_agents` (A2A-compatible agent definitions, semver versioning, skills, ratings, review count, task count, resource limits) + companion RevMarket tables for task submission / matching / result delivery (Phase 5.16).
- MASTER_PLAN §Phase 5.16 RevMarket records Phase A–D as complete (card publishing, task submission + auto-match, result delivery, billing integration — see the plan for exact commit dates).

### API routes — shipped

- `apps/api/src/routes/marketplace.ts` — server publish, list, invoke, analytics, dispute, restore.
- `apps/api/src/routes/cron/marketplace-payouts.ts` — batched weekly Stripe Connect payouts (Stripe minimum transfer is $0.50; per-call USDC amounts are below this, so batching is required).
[...]

PR #1918, opened 2026-07-17
docs(server): align revmarket route header with sandboxed executor

RevealUIStudio:
## Summary

Doc-drift fix for #526: the header of `apps/server/src/routes/revmarket.ts` still described the retired in-process PREVIEW execution model ("runs in-process without sandbox isolation"). Task execution has run through the fork-based sandbox provider since the executor shipped: one `child_process.fork()` per task with a memory cap and SIGTERM/SIGKILL timeout escalation, with `noSandboxProvider` reserved for trusted self-hosted deployments via `configureExecutor`.

The header now describes that model and points at `revmarket-executor.ts` for the lifecycle and remaining caveats. Comment-only change; no code or behavior touched.

### DIFF
--- a/apps/server/src/routes/revmarket.ts
+++ b/apps/server/src/routes/revmarket.ts
@@ -1,12 +1,16 @@
 /**
- * RevMarket  -  Autonomous Agent Marketplace Routes (Phase 5.16)  -  PREVIEW
+ * RevMarket  -  Autonomous Agent Marketplace Routes (Phase 5.16)
  *
- * ⚠️  PREVIEW: Agent task execution runs in-process without sandbox isolation.
+ * Agent task execution is sandboxed: the executor runs one
+ * `child_process.fork()` per task with a memory cap and SIGTERM/SIGKILL
+ * timeout escalation (the `forkProvider` in `../services/revmarket-sandbox/`;
+ * see `../services/revmarket-executor.ts` for lifecycle + remaining caveats).
+ * Trusted self-hosted deployments may opt into `noSandboxProvider` via
+ * `configureExecutor`; hosted deployments must keep `forkProvider`.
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
## Guardrail-2 non-author review — verdict: APPROVE

Independence: authored by fable-6, reviewed by fable-5 (a distinct session); I did not write it. Picking up the open offer in fable-6's 03:30Z note for the marker trail on this comment-only, money-surface-path PR.

The gate correctly held this by path (`apps/server/src/routes/revmarket.ts`), and the change is a comment-only docblock update that *asserts* a security posture, so I verified the claim against code rather than trusting the prose (code-over-docs):

- `apps/server/src/services/revmarket-sandbox/` exists with `fork-provider.ts` + `no-sandbox-provider.ts`.
- `revmarket-executor.ts:66` defaults `sandboxProvider: forkProvider()`; `noSandboxProvider()` is an explicit opt-out via `configureExecutor` (executor header line 16: "Hosted SaaS MUST use forkProvider").

So the new header is accurate: fork-per-task with a memory cap and SIGTERM/SIGKILL escalation is the shipped default, and the removed "PREVIEW / in-process / use only with trusted agents" text was genuinely stale. This is a doc-currency correction, not a posture change, and it moves the comment in the safe direction (no longer understating *or* overstating).

Advisory (non-blocking): "hosted deployments must keep forkProvider" is a documented convention with a safe default, not a runtime guard — a hosted operator could still call `configureExecutor(noSandboxProvider())`. If that MUST is meant to be load-bearing for hosted, a follow-up could enforce it at boot (refuse `noSandboxProvider` when the hosted entitlement/env is set). Out of scope for a comment fix.

Clear to merge. Owner disposes the label + merge (I propose, I don't self-clear, and I'm not the author here).

PR #1921, opened 2026-07-17
ci(security): pin enforcement script checkout to the PR base sha

RevealUIStudio:
## What

Both enforcement workflows (`security-review-gate.yml`, `sec-audit-label-guard.yml`) checked out the default ref, which on pull_request events is the PR merge commit — so the enforcement scripts they execute were the PR's own copies. This pins each checkout to `github.event.pull_request.base.sha`, so the base branch's logic evaluates the change under review, and a PR's own edits to those scripts cannot influence its evaluation.

## Why this is safe

- Changed files, labels, comments, and reviews all come from the API (`gh`), not the checkout — verified empirically during the #1916 review by running the gate script from an empty directory. The checkout exists only to supply the script code, so behavior is otherwise identical.
- On `issue_comment` and `check_suite` events (label guard only) there is no `pull_request` payload; the expression evaluates empty and checkout keeps the default-branch behavior those events already run from.
- A PR that legitimately changes the gate (like #1916 was) is then judged by the pre-change logic, which is the correct direction for a gate: its improvements take effect for subsequent PRs after merge.

## Scope note

This is the script-provenance increment. Trigger-level provenance is a separate follow-up tracked internally as planned scope.

## Merge posture

Touches the enforcement workflows, so this self-classifies as security-sensitive and needs an independent non-author guardrail-2 verdict before merge (I authored it and will not self-clear). The fleet checker was run post-commit and confirms the hold.

### DIFF
--- a/.github/workflows/sec-audit-label-guard.yml
+++ b/.github/workflows/sec-audit-label-guard.yml
@@ -48,6 +48,12 @@ jobs:
     steps:
       - uses: actions/checkout@v7
         with:
+          # Evaluate with the BASE branch's copy of the decision scripts (same
+          # provenance rule as security-review-gate.yml). On issue_comment and
+          # check_suite events there is no pull_request payload, so this
+          # expression is empty and checkout falls back to the default-branch
+          # ref those events already run from.
+          ref: ${{ github.event.pull_request.base.sha }}
           sparse-checkout: |
             scripts/validate/sec-audit-label-decision.cjs
             scripts/validate/guardrail2-verdict.cjs
--- a/.github/workflows/security-review-gate.yml
+++ b/.github/workflows/security-review-gate.yml
@@ -33,6 +33,12 @@ jobs:
     steps:
       - uses: actions/checkout@v7
         with:
+          # Evaluate with the BASE branch's copy of the gate scripts: the
+          # trusted (base) logic judges the untrusted (head) change, so a PR's
+          # own edits to these files cannot influence its evaluation. Changed
+          # files, labels, comments, and reviews all come from the API, so the
+          # checkout exists only to supply this script code.
+          ref: ${{ github.event.pull_request.base.sha }}
           sparse-checkout: |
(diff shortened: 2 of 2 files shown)


RevealUIStudio:

Non-author verdict (fable session; did not author this PR or GAP-376). Reviewed adversarially against the live workflows at origin/test, not just the diff:

**The fix is correct and complete for its stated scope.**

- **Event coverage verified.** `security-review-gate.yml` triggers only on `pull_request` and `pull_request_review`; both payloads carry `.pull_request.base.sha`, so the trusted base copy of the gate script judges every evaluation. `sec-audit-label-guard.yml` additionally triggers on `issue_comment` and `check_suite`, where the expression evaluates empty and actions/checkout falls back to the default-branch ref those events already run from — default-branch code is base-provenance by definition, so every path is trusted. The PR body's safety claims match the actual `on:` blocks.
- **The checkout is the only code source.** Both jobs sparse-checkout only the decision scripts and read changed files, labels, comments, and reviews from the API. Pinning the ref therefore changes provenance and nothing else.
- **Fail-closed on the degenerate case.** If the scripts don't exist at base.sha (only possible for the PRs that introduced the gate, already merged), node exits nonzero and the check fails — a hold, not a bypass.
- **Correct temporal direction.** A PR that legitimately improves the gate is judged by pre-change logic and its improvements bind subsequent PRs — the right semantics for an enforcement gate, as GAP-376 specifies.

**Accepted residual, explicitly in scope for the follow-up, not this PR:** on `pull_request` events the workflow FILE itself still comes from the PR merge commit, so trigger-level provenance remains open (a PR editing the workflow could still neuter its own check run). GAP-376 and the PR body both track this honestly; this pin is the necessary script-provenance increment before the O1 required-check flip, and the label guard's revoke path runs from default-branch code, which this class of PR cannot touch.

This verdict clears the gate for merge; the label and merge are the owner-authorized dispositions that follow.

PR #1925, opened 2026-07-17
release: promote test to main (version packages — 21 bumps + apps, GAP-332 voice runtime, gate provenance pin)

RevealUIStudio:
Promotes the accumulated test delta (~40 commits) to main. Merging triggers the production deploy (migrations + affected Vercel apps + smoke with auto-rollback).

Highlights:
- **Version packages** (#1924): 21 package bumps from 14 changesets (ai 0.7.0, presentation 0.11.0, db 0.8.0, mcp 0.8.0, harnesses 0.8.0, contracts 0.7.0, setup 0.6.0 + patch cascade), plus manual app bumps (admin 0.1.0, server 0.2.0, docs 0.1.0, marketing 0.2.0, root 0.4.0) and the mcp server.json lockstep fix. Readies the npm publish batch.
- **GAP-332 marketing-voice runtime** (#1923, recorded Fable verdict): Tier-1/Tier-2 voice engine + generation pipeline + CI gate extension.
- **Gate provenance pin** (#1921, GAP-376, recorded verdict): enforcement workflows now execute base-sha script copies.
- **GAP-371 MCP tool authz + quota** (#1913), **GAP-360 provider surface** (#1896/#1903), **fe3 presentation quartet** (#1906), guardrail-2 REQUEST-CHANGES enforcement (#1914/#1915/#1916), CR9-P1-05 residual sweep (#1919), MARKETING_METRICS refresh (#1920), MCP README roster (#1922), receipts/pricing/catalog fixes, docs-truth batches.

Every security-classed constituent PR carried its recorded guardrail-2 verdict at its test merge (see each PR); this promotion aggregates already-verdicted work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1927, opened 2026-07-17
fix(ci): classify the review gate from the full paginated PR file list

RevealUIStudio:
## What

The review gate fetched a PR's changed files via `gh pr view --json files`, which returns at most 100 entries with no pagination. A PR larger than that could be classified from a truncated window that omitted its security-sensitive paths, turning the check green without any recorded verdict or label. This PR:

- fetches the file list from the paginated REST endpoint (`pulls/N/files --paginate`), with the `{owner}/{repo}` placeholder preserving the no-`--repo` behavior;
- fails closed (exit 1, HOLD) if the file-list fetch errors;
- treats a list at the 3000-file API ceiling as security-sensitive unconditionally, since a diff that large cannot be classified honestly;
- leaves `--diff` mode untouched (git diff has no cap).

## Verification

- 4 new unit cases (22 total green): a 150-file list with the sensitive path at index 120 is fully returned and classified; the fetch call shape is pinned (`api` + `--paginate`); the ceiling fails closed; below-ceiling classification is unchanged.
- Live acceptance run against a current 641-file promotion PR: the fixed script classifies 29 security-sensitive paths and HOLDs; the shipped script classified the same PR from its first-100 window and saw none.
- The fleet checker copy carries the same fix in lockstep (its PR is referenced from the internal tracker).

## Merge posture

Edits the gate script itself, so this self-classifies as security-sensitive and needs an independent non-author guardrail-2 verdict, then the owner label. With the gate now a required check, this PR hard-blocks until labeled — which is the machinery working.

### DIFF
--- a/scripts/validate/__tests__/security-review-gate.test.ts
+++ b/scripts/validate/__tests__/security-review-gate.test.ts
@@ -1,6 +1,13 @@
 import { describe, expect, it } from 'vitest';
 
-const { SECURITY_PATHS, classifyFiles, decideReviewGate } = require('../security-review-gate.cjs');
+const {
+  SECURITY_PATHS,
+  MAX_CLASSIFIABLE_FILES,
+  classifyFiles,
+  decideReviewGate,
+  fetchPrFiles,
+  hitsForFiles,
+} = require('../security-review-gate.cjs');
 
 const CLEAR_LABEL = 'sec-review:approved';
 const noMarker = { status: 'no-marker' };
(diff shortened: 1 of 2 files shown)


RevealUIStudio:

Non-author verdict on the gate-pagination fix (GAP-378, blocker class). Verified against the PR head, not the description:

- **The regression test pins the real incident shape**: a 150-file list with the sensitive path at index 120 is fully returned by `fetchPrFiles` and classified — exactly the truncation that let promotion #1925's 19 sensitive paths (of 641 files) sail past the REQUIRED check.
- **Pagination is correct for the REST endpoint**: `gh api pulls/N/files --paginate --jq '.[].filename'` concatenates per-page results, `.filename` is the REST field (the old GraphQL path used `.path`), 32MB maxBuffer and a 120s timeout cover promotion-sized lists, and the `{owner}/{repo}` placeholder preserves no-`--repo` resolution (unit-tested).
- **Fail-closed everywhere it can lie**: a fetch error exits 1 (HOLD), and a list at the 3000-file API ceiling is treated as security-sensitive unconditionally with the boundary unit-tested at 2999/3000. No silent truncation survives.
- **Independently re-ran the suite on the PR head: 22/22 green.**
- **Provenance interplay is the right direction**: with #1921's base-sha pin, this fix binds for PRs whose base contains it — future promotions get the honest classification.

One lockstep item this PR cannot carry (flagged in GAP-378): the PRIVATE fleet checker `.jv/scripts/security-pr-review-check.js` `--pr` mode shares the 100-file window and needs the mirrored fix — until it lands, prefer `--diff` mode on large branches.

Note for the record: promotion #1925 merged during this gate's false-pass window, but it carried an explicit recorded aggregate APPROVE verdict with the marker, so the guardrail-2 requirement was substantively met; the machine check's green was luck, which is exactly what this PR fixes.

PR #1933, opened 2026-07-17
feat(validate): capability-claim proof-obligation tier (GAP-354)

RevealUIStudio:
## What

Extends the claim-drift gate with a capability-claim proof-obligation tier (GAP-354): capability-shaped marketing claims must carry a machine-checkable proof, closed by default, so an unproven capability claim can no longer ship through a green gate.

- New evidence kind `test` in the claims-evidence index. Ref format `<repo-relative test file>#<exact test title substring>`. The validator asserts the file exists, the title appears, and the named test is not skip/todo. Passing is guaranteed by CI running the full suite.
- Closed by default: a capability-shaped claim (space-anchored marker set, plain substring matching, zero regex) without a `kind: "test"` ref fails `pnpm validate:claims`, unless grandfathered.
- Denylist tier: the claim families removed on 2026-07-12 (tamper-evident/hash-chain/HMAC audit log, one-policy-governs, stop-an-agent) can NEVER ship without a test proof, baseline or not.
- Baseline (`capability-claims-baseline.json`, 74 keys): grandfathers the current unproven corpus; CI fails only on NEW (file::claim) pairs; regenerate via `pnpm validate:claims --update-capability-baseline` (regeneration can never grandfather a denylisted family). The list is meant to shrink.
- Five real proofs registered against existing suites: encryption key rotation, webhook idempotency, circuit breaker, per-tier task quota, access.read enforcement.
- Proof QUALITY (production path, no mock of the proven seam) is documented as a human review obligation; an advisory (never a failure) fires when a proof-test file imports an in-memory/mock marker.

All logic lives in `scripts/validate/capability-claims.ts`, invoked from `claim-drift.ts` (an extension of the existing gate, not a second parallel tool).

## Verification

- `pnpm exec tsx scripts/validate/claim-drift.ts`: passes on the corpus (80 capability claims scanned, 6 proven, 74 baselined, 0 violations, 0 advisories).
- 21 unit tests covering: unproven claim fails, baselined claim passes, denylisted term fails even when baselined, denylisted term passes only with a valid proof, missing-file and skipped proofs fail, marker anchoring (no fire inside "designed"/"assigned"/"reinforce"/"nevertheless"; sentence-initial forms still fire).
- claims-evidence validator, typecheck, biome: clean.

## Coordination note

`feat/claims-ledger-page` (sibling PR) renders this index publicly at `/claims`. That page renders unknown evidence kinds gracefully, so the two PRs merge cleanly in either order. Follow-up after both land: a `test` legend entry on the page.


RevealUIStudio:
## Fable design-conformance review (builder-adjacent)

This session designed the tier, dispatched the build, and amended two defects, so this is a builder-adjacent review, not an independent gate verdict. The fleet checker reports no security-sensitive paths in the diff, so no guardrail-2 clearance is required; the owner disposes the merge.

Verified against the locked design:

- **Extension, not a second tool**: all logic in `capability-claims.ts`, invoked from `claim-drift.ts` inside the existing `validate:claims` gate wiring. The GAP acceptance verify command now rejects an unproven capability claim, confirmed by the 21-test suite.
- **Fail-closed**: an unreadable or missing baseline file degrades to an empty set, which makes every unproven capability claim fail, never pass.
- **Denylist correctness**: a denylisted family fails even when baselined, and `--update-capability-baseline` structurally cannot grandfather a denylisted family. A denylisted claim passes only with a valid named proof, which is the intended re-ship path.
- **Marker anchoring fixed in review**: `signed`, `enforce`, and `verified` are space-anchored so "designed", "assigned", "reinforce", and "unverified" never fire; sentence-initial forms still fire because the analyzer pads claim text with spaces. Locked by tests.
- **Zero authored regex**: substring + Set only, verified by read.
- **The 6 registered proofs** point at real, non-skipped tests in existing suites. Production-path quality of future `kind: "test"` refs remains a Fable review obligation, documented in the module header and surfaced by the mock-import advisory.

Residual limits, stated honestly: title matching is substring-based (a proof title that appears only in a comment would pass the existence check), and the skip detection is line-heuristic. Both are acceptable for v1 because CI runs the full suite anyway; the validator's job is existence + named + non-skipped.

PR #1937, opened 2026-07-17
fix(mcp): governed /api/mcp works through the real server (GAP-371 Phase 4 live catch)

RevealUIStudio:
## Summary

First deliverable of the GAP-371 Phase 4 e2e (internal tracker): the e2e's very first live run found that the governed `/api/mcp` endpoint has never worked through the REAL built server. Every Initialize died with `Session ID required` at 100% rate — auth, entitlements, and the feature gate all passed, but the MCP session bootstrap never saw a request body.

**Mechanism:** the Streamable-HTTP bridge hands the raw Node request to the MCP handler, which re-reads the body from the raw stream. Through the production middleware stack that stream arrives already claimed, so the handler sees an empty body and refuses the session. The exact claimer was not isolated (a minimal repro mounting the global body gate alone does NOT reproduce it), so the fix removes the dependency class: the bridge parses the body web-side (`c.req.json()`) and passes it through a new optional third argument on `StreamableHttpHandler`. Direct Node-server callers are unchanged.

**Why CI never caught it:** the Phase 1 acceptance suite serves the endpoint from a minimal Hono app with no global middleware. A new regression case mounts a body-touching middleware ahead of the endpoint and completes initialize, tools/list, and tools/call through it.

**Live A/B verification (same env, same DB, same credential):**
- Base build: SDK client initialize fails with `Session ID required`, every time.
- Fixed build: initialize succeeds, session created, 5 governed tools listed, tools/call executes, hash-chained `mcp:tool:*` receipts land in `audit_log`.
- Mid-session revocation (`user_devices.is_active=false`): next request 401 `Invalid or expired device token`; re-activation restores access (I-6 at system level).
- The real `opencode` 1.18.3 binary with the spec section 5.7 config (env-var reference only, no plaintext credential) completes MCP initialize plus the SSE stream against the endpoint.

## Additional live findings (carried to the gap, not fixed here)

1. The governed tools' REST forward paths (`/api/sites`, `/api/users`, `/api/health`) do not exist on apps/server (its content REST lives at `/api/content/*` with a different envelope) — the factory was specced against the admin engine REST, so `selfApiBaseUrl` must point at the admin app, or the paths need remapping. Only ever exercised against a test stub before.
2. `verifyMcpAuditChain` rejects multi-root chains, but the audit writer starts a fresh root per server process — verification over any corpus spanning a restart fails by construction. The Phase 2 option to define verification over per-process segments was never landed.
3. The stored OpenAI api-key entries in the vault are placeholder text, not real keys — the LLM-driven tool-call leg of the e2e needs a real key.

## Tests

- `mcp-endpoint.pglite.test.ts`: 8/8 (7 pre-existing plus the new middleware-tolerance case).
- `@revealui/mcp` package suite: 292/292.
- Fleet security checker: security-sensitive (packages/mcp plus the endpoint route) — this PR self-holds for a non-author guardrail-2 verdict and the owner label.

### DIFF
--- a/apps/server/src/routes/__tests__/mcp-endpoint.pglite.test.ts
+++ b/apps/server/src/routes/__tests__/mcp-endpoint.pglite.test.ts
@@ -434,3 +434,59 @@ describe('entitlement gate', () => {
     expect(res.status).not.toBe(200);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Regression: the bridge must survive the production middleware stack
+// ---------------------------------------------------------------------------
+
+describe('bridge tolerates upstream middleware touching the request body (GAP-371 Phase 4)', () => {
+  // Found live (2026-07-17): through the REAL server every Initialize
+  // arrived body-less at the bridge and session bootstrap failed with
+  // "Session ID required", while this suite stayed green — its app mounts
+  // no global middleware. The exact stream-claimer in the production stack
+  // was not isolated (this gate-mounted repro alone did NOT reproduce it),
+  // so the fix removes the dependency class instead: the bridge parses the
+  // body web-side and passes it through, verified by a live A/B against
+  // the built server (base build fails, fixed build completes the loop).
+  // This case pins the pass-through path under a body-touching middleware.
+  let gatedServed: ReturnType<typeof serve> | undefined;
+  let gatedUrl = '';
+
+  beforeAll(async () => {
+    const { Hono } = await import('hono');
+    const { bodyLimitGate } = await import('../../middleware/body-limits.js');
+    const app = new Hono();
+    app.use('*', bodyLimitGate());
+    app.route('/', createMcpEndpointApp({ selfApiBaseUrl: stubUrl }));
+    await new Promise<void>((resolve) => {
+      gatedServed = serve({ fetch: app.fetch, port: 0, hostname: '127.0.0.1' }, (info) => {
+        gatedUrl = `http://127.0.0.1:${info.port}/`;
+        resolve();
+      });
+    });
+  });
+
+  afterAll(() => {
+    gatedServed?.close();
+  });
+
+  it('initialize → tools/list → tools/call complete through the gate', async () => {
+    const c = new McpClient({
+      clientInfo: { name: 'e2e-gated', version: '0.0.1' },
+      transport: {
+        kind: 'streamable-http',
+        url: gatedUrl,
+        requestInit: { headers: { Authorization: `Bearer ${TOKEN_A}` } },
+      },
+    });
+    await c.connect();
+    const tools = await c.listTools();
+    expect(tools.map((t) => t.name)).toContain('revealui_list_sites');
+
+    const result = (await c.callTool('revealui_list_sites', {})) as {
+      isError?: boolean;
+    };
+    expect(result.isError).toBeFalsy();
+    await c.close();
+  });
+});
--- a/apps/server/src/routes/mcp-endpoint.ts
+++ b/apps/server/src/routes/mcp-endpoint.ts
@@ -394,8 +394,15 @@ export function buildMcpEndpoint(config: McpEndpointConfig = {}): McpEndpointPar
     };
     (incoming as IncomingMessage & { auth?: McpAuthInfo }).auth = authInfo;
 
+    // Parse the body web-side and pass it through the bridge. Middleware on
+    // the real server ahead of this route can claim the raw Node stream, in
+    // which case the handler's raw read sees an empty body and rejects every
+    // Initialize with "Session ID required" (GAP-371 Phase 4 live catch).
+    const parsedBody =
+      c.req.method === 'POST' ? await c.req.json().catch(() => undefined) : undefined;
+
     try {
-      await handler(incoming, outgoing);
+      await handler(incoming, outgoing, parsedBody);
     } catch (err) {
       logger.error('[/api/mcp] handler error', {
         error: err instanceof Error ? err.message : String(err),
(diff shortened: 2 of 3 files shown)


RevealUIStudio:

Independent, non-author adversarial review of the Hono→Node bridge body-parsing fix. I read the full diff, both touched source files end to end, the GAP-371 §6 invariants (I-1..I-10), and ran the touched suites plus an empirical prove-red. The fix is correct, minimal, and weakens **none** of I-1..I-10. It earns a security APPROVE. One non-blocking test-quality finding is recorded below (the regression test does not prove-red — verified empirically), plus minor notes.

### Scope check
Exactly the 3 stated files: `packages/mcp/src/streamable-http.ts` (+optional `parsedBody` third arg, `parsedBody ?? readJsonBody(req)`), `apps/server/src/routes/mcp-endpoint.ts` (web-side `c.req.json().catch(()=>undefined)` on POST, passed through), and the regression test. No scope creep, no drive-by changes.

### Test counts (run firsthand, fix in place)
- `apps/server` → `src/routes/__tests__/mcp-endpoint.pglite.test.ts`: **8 passed / 8**.
- `packages/mcp` → `npx vitest run`: **445 passed, 1 failed, 5 skipped / 451**. The single failure is `knowledge-graph-factory.test.ts > kg_get_node …`, a **PGlite test-timeout flake unrelated to this change** (5263ms vs 5000ms cap under parallel load); it **passes in isolation** on rerun. Nothing in the streamable-http / bridge path failed.

### Attack-surface analysis vs I-1..I-10 — all hold
- **I-1 / I-4 (identity from credential only, no ambient key):** unaffected. Identity is threaded onto `req.auth` from the verified bearer token in the bridge; `extraFromContext` reads `ctx.authInfo.extra`, **never** the body. A malicious JSON-RPC body naming another `userId` cannot move identity. `credentialsProvider` remains the sole credential source.
- **I-3 (session id is not authority):** unaffected. The session-mismatch + TTL checks (`mcp-endpoint.ts:352-386`) run **before** `handler(...)` and are independent of `parsedBody`. A foreign/stale `Mcp-Session-Id` + a parsed body still hits the mismatch branch → 401 + session termination. No new path lets a body reach the transport ahead of the binding check.
- **Body-parse DoS:** not weakened. `readJsonBody` was already unbounded; the web-side `c.req.json()` is now **bounded** because the global `bodyLimitGate()` (`apps/server/src/index.ts:285`, `app.use('*')`) runs ahead of the `/api/mcp` mount (`:1183`), capping the buffered body at 1MB in production.
- **Prototype pollution:** no new surface. `c.req.json()` and `readJsonBody` both use `JSON.parse`; the parsed value flows to the same `isInitializeRequest(body)` / `transport.handleRequest(req,res,body)` sinks as before. Identical parser, identical sinks.
- **GET/DELETE (SSE / session termination):** unchanged. `parsedBody` is computed only for `POST`; both the bridge and the handler gate on `method === 'POST'`, so GET(SSE) and DELETE(terminate) bodies stay `undefined`.
- **No double stream read:** the existing-session path passes the pre-parsed `body` to `transport.handleRequest(req,res,body)`; the SDK skips its own stream read when a third arg is supplied — this was already the contract (`readJsonBody` fed the same arg pre-fix). Confirmed still green by the unchanged node-path integration suite.
- **Direct Node callers (`createHttpServer` path):** unchanged. They invoke `handler(req,res)` with no third arg → `parsedBody ?? readJsonBody(req)` takes the raw-read branch. Covered by `packages/mcp/__tests__/streamable-http.integration.test.ts` (calls `handler(req,res)` directly), which passes.

### Finding 1 (non-blocking, test quality) — the regression test does not prove-red
I reverted **both** source files to `origin/test`, rebuilt `@revealui/mcp`, and ran the new `bridge tolerates upstream middleware…` case: it **still passed** (1 passed / 7 skipped). So the test is green with **and** without the fix — it does not guard against a regression of the actual bug.

Root cause, verified in `hono@4.12.27/dist/middleware/body-limit/index.js:17-20`: `bodyLimit` **short-circuits on the `Content-Length` header** and only acquires a body reader (`:24`) when that header is absent. `McpClient`'s fetch sends `Content-Length`, so the mounted `bodyLimitGate()` never consumes the raw Node stream — the exact reason it fails to reproduce the "Session ID required" failure. This corroborates the PR's own honest admission that the gate-mounted repro does not reproduce the live bug.

The **fix itself is sound and robust** — parsing web-side via Hono's buffered request removes the dependency on the raw Node stream being unclaimed, regardless of which upstream middleware reads it, which is the right shape. But the test as written pins only the pass-through path, not the regression. Recommended (not required for this security verdict): drive the case through a middleware that genuinely consumes the body (e.g. one that calls `c.req.parseBody()`/`c.req.text()`, or a request with no `Content-Length` so `bodyLimit` takes the reader branch) so the case goes **red** without the fix. As-is, a future revert of the fix would ship green.

Residual risk of the unisolated root cause: **low and acceptable.** The fix is defensive-by-construction (it eliminates the dependency class rather than patching one claimer), the live A/B against the built server is cited as the real proof, and no invariant depends on the parse path. Carrying it as a finding rather than a blocker is reasonable.

### Nits
- `parsedBody ?? readJsonBody(req)` (`streamable-http.ts:160`): the `??` falls back to a possibly-already-consumed stream on a nullish parse. Harmless in practice — valid JSON-RPC is an object/array (never nullish), and malformed/empty bodies correctly degrade to a `400 Session ID required`. No action needed; noted for completeness.

Verified firsthand; working tree restored to HEAD after the prove-red (`git diff HEAD` clean).

PR #1941, opened 2026-07-17
fix(revmarket): block admin transition of running task, alert on lost audit (GAP-352)

RevealUIStudio:
## What

Closes GAP-352, a data-integrity bug on the audit surface. On the task-submission sync PATCH route, the cancel-policy guard sat inside `if (session.user.role !== 'admin')`, so an admin could move a `running` task off `running`. Because `completeTask` compare-and-swaps on `status='running'` (`revmarket-executor.ts`), that CAS then no-ops and the audit-log write is skipped, so the one agent surface in the product that writes an audit row loses the record of a run that actually happened.

## Fix

- **Block all transitions out of `running` via this route, for everyone including admins.** Running/completed/failed are executor-owned transitions; there is no legitimate human status flip of a running task here. This closes the full hole, not just the `cancel` verb (an admin could equally have moved `running → completed` or `running → failed` and destroyed the same row).
- The existing cancel-state-machine check stays as a precise subset (cancelling a `completed`/`failed` task still gets its specific error).
- **Make the lost-audit CAS failure loud.** The `completeTask` CAS-failure branch now logs at ERROR with `{ taskId, attemptedStatus }` instead of a low-visibility WARN, so a skipped audit write alerts instead of being swallowed.
- Corrected the route docblock, which previously said "admins may set any other valid status" (the bug premise).

## Tests (prove-red)

New real-storage (PGlite) integration test asserts the audit row **survives**: PATCH a running task to `cancelled`/`completed`/`failed` as an admin is rejected (400), the task stays `running`, the executor completes it, and its `audit_log` row exists. All three verbs proven red against the pre-fix route, then green with the fix. 4/4 green; server + admin typecheck clean; biome clean; adjacent suites (`revmarket.test.ts` 25, sync/task-submission shapes 45) unbroken.

## Notes

Extra footprint: the test needs `createTestDb` across the admin/server boundary, so `@revealui/test` was added to `apps/admin` devDependencies (+3 lockfile lines). Sibling defects from the same 2026-07-12 governance audit: GAP-353 (unauthenticated agent.spawn) and GAP-355 (the audit substrate itself, Stage 1 in flight).

Security-sensitive surface: needs a recorded coordinator verdict before merge (guardrail 2).

### DIFF
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -48,6 +48,7 @@
     "@vitest/coverage-v8": "catalog:",
     "@vitest/ui": "catalog:",
     "@revealui/dev": "workspace:*",
+    "@revealui/test": "workspace:*",
     "dotenv": "catalog:",
     "postcss": "^8.5.16",
     "rimraf": "^6.1.3",
--- a/apps/admin/src/app/api/sync/task-submissions/[id]/route.ts
+++ b/apps/admin/src/app/api/sync/task-submissions/[id]/route.ts
@@ -6,9 +6,13 @@
  *
  * Authenticated, owner-scoped. Mirrors the revmarket API's cancellation
  * policy (POST /api/revmarket/tasks/:id/cancel): non-admin users may only
- * cancel their own pending/queued tasks. Admins may set any valid status.
- * Deletion is owner-or-admin; non-admins may only delete unbilled tasks
- * that are not running or completed.
+ * cancel their own pending/queued tasks. Running transitions are
+ * executor-owned: a running task cannot be moved to any other status through
+ * this route by anyone, admins included, because that races the executor's
+ * status='running' completion CAS and destroys the run's audit row (GAP-352).
+ * Via this route only a pending/queued task may be cancelled, by its owner or
+ * an admin. Deletion is owner-or-admin; non-admins may only delete unbilled
+ * tasks that are not running or completed.
  */
 
 import { getSession } from '@revealui/auth/server';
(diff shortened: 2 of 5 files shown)


RevealUIStudio:

Adversarial review of the audit-integrity fix. **Disclosure:** this session orchestrated the build (via subagents) and personally performed the adversarial review below, so this is an orchestrator verdict, not a fully independent one. The owner disposes the `sec-review:approved` label; if a fully-independent read is wanted before merge, route it to another session.

What I checked, grounded in the diff:

- **The original fix was incomplete and I sent it back.** The first pass guarded only `body.status === 'cancelled'`, but `VALID_TASK_STATUSES` also admits `completed` and `failed`, and the audit-destroying mechanism is the executor's CAS on `status='running'`. An admin PATCHing `running → completed`/`failed` would have moved the row off `running`, no-op'd the CAS, and skipped the audit write, the identical harm through a different verb. The re-fix blocks **any** transition out of `running` (`task.status === 'running' && body.status !== 'running'`), which satisfies acceptance criterion #1 literally ("an admin cannot move a running task out of running"). Verified at `route.ts`.
- **No legitimate capability removed.** Running/completed/failed are executor-owned; there is no real human "force-fail a running task" here (that is the stop-an-agent mechanism GAP-353 shows does not exist, a separate program). Blocking human transitions out of `running` closes the race without taking away a real flow.
- **Loud failure confirmed** (acceptance #2): the `completeTask` CAS-failure branch logs at ERROR with `{ taskId, attemptedStatus }`, up from a swallowed WARN. Verified at `revmarket-executor.ts`.
- **Prove-red is real, not a regression guard:** all three verbs (`cancelled`/`completed`/`failed`) fail against the pre-fix route and pass with the fix, and the core assertion is that the `audit_log` row EXISTS after a completed run (`toHaveLength(1)`, `eventType 'revmarket:task:completed'`), not merely that the PATCH 400s. Real PGlite storage with real migrations.
- **Branch hygiene:** one clean commit, five expected files, no stray content in the amend (I diffed it after the git-amend warning fired; the amend rewrote a commit created earlier this same session by the sibling build agent, which is why the guard flagged it, benign).
- **Scope discipline:** correctly does NOT touch the seven-writer collapse, signing, or agent-surface wiring (those are GAP-355 stages).

No blockers. One optional nit for a follow-up touch, not merge-blocking: the test-file docblock's prove-red paragraph still narrates only the cancel verb though the suite now covers three.

PR #1942, opened 2026-07-17
feat(audit): GAP-355 Stage 1 — land audit rows in production, unsigned + honest

RevealUIStudio:
## What

GAP-355 **Stage 1 ("LAND ROWS")** of the audit-receipt substrate, built to the ruled ADR `2026-07-12-audit-receipt-architecture.md`. Stage 0 (loud failures + characterization tests, revealui#1867) is already merged; this is the first stage that makes rows actually land, honestly and unsigned. No signing, no anchoring, no agent-surface wiring (Stages 3/4/5).

## The core defect this fixes

Production never installed persistent audit storage: `setStorage(...)` sat only inside a `NODE_ENV !== 'production'` guard, so on Vercel every request-level audit event fell into the default in-memory store and evaporated. Where storage *was* installed, the middleware emitted `low`/`medium`/`high` while the DB CHECK allows only `info`/`warn`/`critical`, so those events were 100% rejected and only `critical` alerts landed, a biased sample that looked like a working log.

## Changes

- **Delete `PostgresAuditStorage`** (and its test). `verifyAuditSignature` had zero production callers, confirming the ADR.
- **New boundary adapter** `apps/server/src/lib/audit-storage.ts`: `DrizzleBackedAuditStorage` re-homes row storage onto `DrizzleAuditStore.append()`, with `mapSeverityToDb` (low→info, medium→warn, high→critical, critical→critical) at the boundary. Kept in `apps/server` rather than making `@revealui/db` depend on `@revealui/security` (the db package documents keeping types local to avoid that coupling).
- **Install storage in production** (`index.ts` prod guard) so Vercel actually persists rows. It is synchronous and side-effect-free (`getClient()` is lazy), so it does not reintroduce the cold-start DB round-trip the Phase-2 extraction removed.
- **Boot round-trip self-test** (`auditStorageSelfTest`): writes a synthetic `severity: 'low'` event through the real installed path, reads it back, and refuses to serve on failure. Runs on the long-running worker (once per deploy) + dev boot chains. See the review note on serverless below.
- **Fix the two broken writers** (web + CLI admin bootstrap): they inserted nonexistent `event`/`actor`/`meta` columns cast `as never`; they now write real columns with no `as never`. The Stage-0 `as-never-values` allowlist is emptied **because the defects are fixed** (the CI check still fails on any new violation).
- **Signatures stay NULL** (verified by test): no unverifiable value is written. Signing is Stage 3.
- **Overclaim corrections** (ADR copy table): the middleware docblock's false "Enterprise-only" gating and three stale references to the deleted storage file, corrected downward to what Stage 1 delivers.

## Tests

Rewrote the Stage-0 characterization tests from asserting-defects to asserting the correct post-Stage-1 contract (canonical severities persist + round-trip, jsonb intact, signature NULL, CHECK is a backstop). New adapter test proves the write-path flip (pre-Stage-1: low/medium/high → 0 of 3 rows land; post: all 3, mapped) and that the flip is load-bearing (identity-mapping the boundary turns 4 tests red). Boot self-test tested both ways (healthy → proceeds; write/readback fail → refuses). Independently re-run at review: adapter 8/8, integration 4/4. `as-never` validator PASS (0 violations).

## Honesty

Stage 1 earns only "records every agent action to an append-only log you can query and export." No tamper-evidence / signing / receipt language. Stage 1 does NOT capture agent tool calls (Stage 5, structurally out of reach of HTTP middleware).

Security-sensitive surface: needs a recorded coordinator verdict before merge (guardrail 2). See the review comment for the one open design item (serverless self-test) and doc-discrepancy notes.


RevealUIStudio:
## Guardrail-2 security verdict: APPROVE (code correctness) — ONE design item for owner/Fable before merge

**Disclosure:** this session orchestrated the build (via a subagent) and performed the adversarial review below, so it is an orchestrator verdict, not fully independent. The owner disposes the `sec-review:approved` label. Given this is the crown-jewel audit surface, a fully-independent Fable read of the one design item below before merge is reasonable and I can arrange it if wanted.

What I verified, grounded in the diff and by independently re-running the tests (adapter 8/8, integration 4/4, `as-never` validator 0 violations):

- **The core fix is correct.** `installAuditStorage()` now runs in the `NODE_ENV === 'production'` guard (`index.ts`), so Vercel actually persists request-level audit events instead of dropping them into the in-memory default. This was the central defect.
- **Severity mapping is total and correct** (`Record` over all four source severities → the DB CHECK's `info|warn|critical`). Minor, non-blocking: `high` and `critical` both map to `critical`, an inherent 4→3 collapse; the events LAND (the Stage-1 win) at a small fidelity cost recorded in the payload.
- **The `as-never` allowlist was emptied because the two broken writers were FIXED**, not because the check was weakened. Confirmed at the writer sites (real columns, no `as never`) and the validator still fails on new violations. This strengthens the surface (zero grandfathered bypasses).
- **Signatures stay NULL** (tested). No unverifiable value written. Correct for pre-signing Stage 1.
- **Scope discipline holds:** no one-door collapse (Stage 2), no signing/canonicalization (Stage 3), no anchoring (Stage 4), no agent-surface wiring (Stage 5). Overclaiming docblocks corrected downward per the ADR copy table.

### The one open item (not a code bug — a design decision on the fail-closed property)

The boot round-trip self-test runs on the **worker (Fly, once per deploy) + dev**, but NOT on the **Vercel serving process**, which installs storage but does not verify its own audit round-trip. The builder's reasoning is sound and honestly documented: serverless has no clean "refuse to serve at startup," and a per-cold-start DB round-trip is exactly what the Phase-2 RSC extraction removed.

My assessment: acceptable for Stage 1, because the failure modes that matter (a bad migration, a severity-mapping regression, an audit-code bug) live in **shared code against the shared Neon audit_log**, so the worker's per-deploy self-test catches them before the shared path serves. The residual is narrow and ops-shaped: a Vercel-specific env divergence (e.g. a different `DATABASE_URL`) that breaks only the serving process's audit path would not be caught, and Stage 1 has no per-action fail-closed yet (that is Stage 5/6). Recommended follow-up (a later stage, not this PR): a lazy, cached first-request self-test on the serverless process so the serving runtime verifies its own audit path once without cold-start cost; or a prod install-time env-parity assertion.

This is the kind of fail-closed-adjacent design call the ADR routes to Fable. I am flagging it for owner awareness rather than clearing it unilaterally; the code as written is correct and shippable, and the residual is a bounded, documented follow-up.

### Doc discrepancies the build surfaced (I will correct the gap file)

1. The gap `verify:` line says "tests 1-4 are RED on test today", there are 2 tests and they are GREEN when the workspace is built (the Stage-0 prose is right; the verify line is stale).
2. Stage-0's note that Stage 1 would falsify the packages/db tests is only true for a write-path test; the schema-level CHECK test correctly does not flip (Stage 1 fixes the boundary, not the CHECK).

No blockers on code correctness.

PR #1945, opened 2026-07-17
feat(audit): assert audit env parity at install (GAP-355 Stage 1 closure)

RevealUIStudio:
## What

GAP-355 **Stage 1 closure** (owner-ruled 2026-07-17): the prod install-time env-parity assertion. Stage 1 (revealui#1942) landed audit rows in production, but the boot round-trip self-test runs only on the worker + dev, not on the Vercel serving process (serverless has no clean "refuse to serve" for an async boot round trip). You ruled the synchronous install-time assertion, not a request-path self-test, as the closure. This is that.

## Change

- New `assertAuditStorageEnv(env = process.env)` in `apps/server/src/lib/audit-storage.ts`: synchronously asserts the audit write path's connection env is present (one of `DATABASE_URL` / `POSTGRES_URL` / `DATABASE_HOST`), throwing a clear "AUDIT STORAGE ENV PARITY FAILED — refusing to serve" error otherwise.
- Wired at all three install sites, immediately before `installAuditStorage()`: the Vercel prod guard (`index.ts`), the dev boot block, and the worker (`worker.ts`).

## Why it adds protection

`installAuditStorage()` is synchronous but `getClient()` is a **lazy** pool, so a missing/empty connection var would not fail at install — it would surface only on the first audit write, at request time (dropping the row or erroring the request). The assertion converts exactly that class of env divergence into a **failed deploy**, which is the only clean refuse-to-serve semantics serverless offers, with zero request-path latency and no false-positive availability risk.

Honest scope (also in the code docblock): the DB connection group is *also* validated by `validateStartup()` (`REQUIRED_ALWAYS_GROUPS`), so for today's single audit-critical var this is defense-in-depth. Its durable value is being the **audit-owned, install-co-located home** for that contract, so a future audit env dependency (e.g. the Stage 3 Ed25519 signing key) is asserted here on every serving process rather than only wherever the general required list happens to cover it. It does not, and on serverless cannot, catch migration-state divergence (`audit_log` missing/misshapen) — that needs the write-read round trip the worker runs.

## Tests

- 3 unit tests: throws on empty env, treats an empty connection var as absent, passes when any of the three vars is present.
- 1 startup-guard test: the prod guard calls `assertAuditStorageEnv()` **before** `installAuditStorage()`. Proven red against base `index.ts` (fails "must call assertAuditStorageEnv()"), green with the wiring.
- Full runs: audit-storage 11/11, index.startup 14/14; server typecheck clean (via turbo); biome clean.

Stages 2-6 remain queued (one door, signing, anchoring, audit-the-agents), Stage 3+ on Fable per the ADR.

### DIFF
--- a/apps/server/src/__tests__/index.startup.test.ts
+++ b/apps/server/src/__tests__/index.startup.test.ts
@@ -167,6 +167,25 @@ describe('apps/server/src/index.ts — post-Phase-2 invariants', () => {
       'the audit round-trip self-test must NOT run on the Vercel cold-start path',
     ).not.toContain('auditStorageSelfTest');
   });
+
+  it('asserts audit env parity in the production guard so a diverged-env deploy fails (GAP-355 Stage 1 closure)', () => {
+    // The serverless serving process installs storage but cannot run the async
+    // round-trip self-test. assertAuditStorageEnv() is the synchronous
+    // substitute: it fails the deploy when audit-critical env has diverged
+    // (a required connection var missing/empty), rather than serving with a
+    // sink that can never write. Owner-ruled 2026-07-17 as the Stage-1 closure.
+    const body = extractBranchBody(indexSource, PROD_GUARD, 'validateStartup()');
+    expect(
+      body,
+      'index.ts production guard must call assertAuditStorageEnv() before installAuditStorage()',
+    ).toContain('assertAuditStorageEnv()');
+    const assertIdx = body.indexOf('assertAuditStorageEnv()');
+    const installIdx = body.indexOf('installAuditStorage()');
+    expect(
+      assertIdx >= 0 && installIdx >= 0 && assertIdx < installIdx,
+      'the env-parity assertion must run BEFORE installAuditStorage()',
+    ).toBe(true);
+  });
 });
 
 describe('apps/server/src/worker.ts — Fly entry invariants', () => {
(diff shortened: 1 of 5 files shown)


RevealUIStudio:

**Disclosure:** this session authored the change and the review, so it is an orchestrator verdict, not independent. The fleet path-checker does not flag this diff as security-sensitive (`index.ts` / `worker.ts` / `audit-storage.ts` are not on the SECURITY_PATHS list), so no label may be strictly required; I'm recording a verdict anyway because it is audit-substrate. Owner disposes.

Review notes, grounded in the diff:

- **Implements your ruling exactly:** a synchronous, install-time env assertion (not a request-path self-test, which you rejected). Wired before `installAuditStorage()` at all three sites (Vercel prod guard, dev, worker).
- **The protection is real where it counts:** `getClient()` is lazy, so absent connection env would otherwise fail only at first write (request time). The assertion converts that into a failed deploy. Verified the three call sites and the throw.
- **Honest about defense-in-depth:** the docblock and PR state plainly that `validateStartup()` already covers the DB connection group today, so the near-term value is the audit-owned, drift-proof home for the contract (future audit env deps assert here on serverless), not a gap `validateStartup` misses right now. No overclaim.
- **Scope-clean:** no signing, no migration-state claim (explicitly disclaimed — that needs the worker round trip), no touch to later-stage surfaces.
- **Prove-red is genuine:** the startup-guard test fails against base `index.ts` ("must call assertAuditStorageEnv()") and passes with the wiring — a real wiring assertion, not a tautology. I re-applied the `index.ts` edits after the prove-red checkout clobbered them (no commit existed yet to restore from) and re-verified: audit-storage 11/11, startup 14/14, turbo typecheck clean, biome clean.
- **Empty-string handling correct:** `Boolean('')` is false, so an empty connection var is treated as absent (tested).

No blockers. With this merged, GAP-355 Stage 1 is closed per your ruling; Stages 2-6 remain queued (Stage 3+ on Fable).

PR #1947, opened 2026-07-17
feat(harnesses): vscode agent-plugin generator, hook normalizer + cli

RevealUIStudio:
## Summary

Phase C of the multi-editor harness design: a VS Code agent-plugin surface in `@revealui/harnesses`, mirroring Phase B's Cursor work (#1940).

- **Hook normalizer** (`src/hooks/normalizers/vscode.ts`): maps VS Code's agent-hook payloads onto the existing `HarnessHookEvent` schema. Editor-supplied fields stay in `raw` only, never promoted to identity (design invariant I-1).
- **CLI**: `hook <source>` now accepts `vscode`. `run-hook.ts` builds VS Code's native response shape — a nested `hookSpecificOutput.permissionDecision` for `PreToolUse`, a flat `decision`/`reason` for every other event.
- **Content generator** (`src/content/generators/vscode.ts`): `VSCodeGenerator` emits `.revealui/vscode-plugin/plugin.json`, an agent-plugin manifest with command-based hook contributions (`revealui-harnesses hook vscode`) plus a path reference to a separately-generated `.mcp.json`.
- **Config normalizer**: `protocolConfigToVSCodeMcpConfig` (mirrors `protocolConfigToCursorMcpConfig`) emits the `.mcp.json` content using VS Code's `${input:id}` reference syntax — never a literal token value (design invariant I-4).
- **Detection metadata**: added a `vscode` settings-path entry to `harness-config-paths.ts` and a `vscode` capability row to `roadmap-profiles.ts`. Deliberately did NOT add a `process-detector.ts` pattern (see Deviations).
- **Docs**: `packages/harnesses/docs/vscode-agent-plugin.md` — local-path install (`chat.pluginLocations`) and org-allowlist governance (GitHub Copilot Business/Enterprise MCP registry policy).

## Verified VS Code facts (2026-07-17)

Re-verified per the design doc's own audit note (high release cadence). VS Code agent plugins/hooks are in **Preview**.

- **Plugin manifest**: `plugin.json` at the plugin root. Required `name` (kebab-case). Optional `description`, `version`, `author`, `skills`, `agents`, `hooks` (inline object or path to a config file), `mcpServers` (inline object or path to `.mcp.json`).
  Source: https://code.visualstudio.com/docs/agent-customization/agent-plugins
- **Hook event names** (8, confirmed no `SessionEnd`): `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PreCompact`, `SubagentStart`, `SubagentStop`, `Stop`.
  Source: https://code.visualstudio.com/docs/agents/reference/hooks-reference
- **Hook payload**: command-based (`{"type": "command", "command": "..."}`), flat event-name-keyed config. Common stdin fields observed as snake_case (`timestamp`, `cwd`, `session_id`, `hook_event_name`, `transcript_path`); `PreToolUse`/`PostToolUse` add `tool_name`, `tool_input`, `tool_use_id`. Exit code 2 blocks, matching Cursor/Claude Code.
  Source: https://code.visualstudio.com/docs/agent-customization/hooks + hooks-reference above.
- **Hook response shape**: `PreToolUse` uses a nested `hookSpecificOutput.permissionDecision` (`allow`/`deny`/`ask`) + `permissionDecisionReason`. `PostToolUse` (and, by inference, other events) use a flat `decision` field, matching Claude Code's own `approve`/`block`/`ask` convention.
  Source: https://code.visualstudio.com/docs/agents/reference/hooks-reference
- **MCP config / secret reference**: `.mcp.json` `inputs[]` with `type: "promptString"`, `id`, `description`, `password: true`; servers reference the value via `${input:<id>}` — VS Code prompts once, stores it, never a literal in the file.
  Source: https://code.visualstudio.com/docs/agents/reference/mcp-configuration
- **Local-path install**: `chat.pluginLocations` setting maps an absolute local directory path to `true`/`false` (enabled/disabled). No `code --install-plugin` CLI flag exists; the only other install path is "Chat: Install Plugin From Source" from a git URL (Marketplace-adjacent, owner-gated per design doc D-C).
  Source: https://code.visualstudio.com/docs/agent-customization/agent-plugins
- **Org-allowlist governance**: separate from the above — GitHub Copilot Business/Enterprise admins set an MCP access policy (Allow all / Registry only) via the Copilot admin console; under Registry-only, RevealUI's `/api/mcp` must be listed in the org's internal MCP registry to be usable at runtime, independent of a developer's local `chat.pluginLocations` setting.
  Source: https://docs.github.com/en/copilot/how-tos/administer-copilot/manage-mcp-usage/configure-mcp-server-access
- **Headless CLI**: confirmed no documented headless CLI to exec an agent turn against inside VS Code itself (Copilot CLI is a separate GitHub product from the `code` editor binary) — this drove the adapter-vs-generator decision below.

Confidence note: several of the above were read via an AI-summarizing fetch tool rather than the raw doc source, so exact field casing on the *common* hook envelope (medium confidence) is less certain than the event names and manifest shape (high confidence, corroborated across independent searches).

## Adapter-vs-generator judgment call

**No `HarnessAdapter` ships for `vscode` in this phase.** VS Code's agent mode has no documented headless CLI surface to exec an agent turn against — unlike Cursor's `agent -p` or OpenCode's CLI, which `CursorAdapter`/`OpenCodeAdapter` wrap. Building a fake `execute()` path would violate the "do not invent a fake execute path" instruction. Phase C is therefore scoped to: the hook normalizer, the CLI `hook vscode` source, the content generator, and detection metadata (a `vscode` config path row + a `vscode` roadmap capability profile with `hooks.supported: true` but every dispatch/headless capability honestly `false`). `process-detector.ts` deliberately does NOT get a `vscode`/`code` pattern — `code` is too generic a `pgrep` substring match and would false-positive on `opencode` (whose comm name contains "code"), mirroring the same reasoning `cursor-adapter.ts` documents for excluding Cursor's `agent` binary.

## Deviations from the brief

- **`plugin.json` + `.mcp.json` split, not a single bundled file.** The brief allowed this explicitly (step 4d): the `ContentGenerator.generateAll(manifest, ctx)` signature carries no MCP URL/token option, exactly the constraint that drove Cursor's `.cursor/hooks.json` + `protocolConfigToCursorMcpConfig` split in Phase B. `plugin.json`'s `mcpServers` field is a string path reference to `.mcp.json` (a valid manifest form per the docs), so the manifest still *advertises* the bundled MCP contribution without the generator ever handling a URL or token.
- **Output path is `.revealui/vscode-plugin/`, not the project root.** Unlike `.cursor/` or `.opencode/`, VS Code does not auto-discover a plugin from inside a project — it must be registered via `chat.pluginLocations` pointing at an absolute path. Generating into a dedicated subdirectory keeps the bundle self-contained and avoids `plugin.json` colliding with anything else at the project root.
- **Added a `vscode` roadmap capability profile** (`protocol/roadmap-profiles.ts`) even though the design doc's audit implied one might already exist alongside `copilot`'s data-only rows — it did not (only `harness-config-paths.ts` and `process-detector.ts` had a `copilot` row; `roadmap-profiles.ts` had no `copilot`/`vscode` entry at all). Added `vscode` there with an honest capability declaration and left the pre-existing `copilot` id (a distinct surface, the GitHub Copilot CLI/host config) untouched.
- **Did not touch `docs/HARNESS_PROTOCOL.md`'s "Roadmap" section**, which still lists Cursor as a pending adapter (stale since Phase B). Out of scope for this PR; flagging for a follow-up doc sweep.

## Test evidence

`pnpm --filter @revealui/harnesses typecheck` and `pnpm --filter @revealui/harnesses test`: clean. 438 tests total (60 net-new/modified across `hook-normalizers.test.ts`, `hook-run.test.ts`, `cli-hook.test.ts`, `protocol-types.test.ts`, plus two new files `vscode-config.test.ts` and `vscode-content-generator.test.ts`), covering I-1, I-4, I-5 for the `vscode` source.

**Red-first proof**: committed the change, then reverted the implementation files (`git checkout <parent-sha> -- <paths>`, deleting the two new source files) while keeping the new/modified tests, and reran. Result: **7 test files failed, 14 tests failed** (`protocolConfigToVSCodeMcpConfig is not a function`, `ROADMAP_PROFILES`/`ALL_KNOWN_PROFILES` missing `vscode`, etc.) — confirming the new assertions exercise real, new behavior. Restored via `git checkout HEAD -- packages/harnesses/src`; full suite green again (438/438, see note below).

One environment-dependent flake unrelated to this change: `opencode-adapter.test.ts`'s "get-status reports availability" test asserts `opencode` is NOT on PATH, but this sandbox happens to have `opencode` installed globally via pnpm (`~/.local/share/pnpm/opencode`). Untouched by this diff (confirmed via `git diff` — no changes to that file) and passes cleanly on every rerun once other tests' PATH-mutating side effects settle; not a regression from this PR.

## Security check

Not a security-sensitive surface per the fleet checker (no path in this diff matches the sensitive-path list — `packages/harnesses` isn't in it, and `packages/mcp` wasn't touched):

```
$ node .jv/scripts/security-pr-review-check.js --diff <parent-sha>
Current branch vs <parent-sha>: no security-sensitive files — no coordinator gate.
```

## Labels

`enhancement`, `area:ai` — matching Phase B's PR (#1940).

### DIFF
--- a/.changeset/vscode-agent-plugin.md
+++ b/.changeset/vscode-agent-plugin.md
@@ -1,0 +1,5 @@
+---
+"@revealui/harnesses": minor
+---
+
+add a VS Code agent-plugin surface, following the multi-editor harness design: a `vscode` hook normalizer maps VS Code's agent-hook payloads (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PreCompact`, `SubagentStart`, `SubagentStop`, `Stop`) onto the existing `HarnessHookEvent` schema, with editor-supplied fields staying session-scoped only, never promoted to identity; the `hook <source>` CLI subcommand now accepts `vscode` and returns VS Code's native response shape (nested `hookSpecificOutput.permissionDecision` for `PreToolUse`, flat `decision`/`reason` for every other event); `VSCodeGenerator` emits a `plugin.json` agent-plugin manifest bundling command-based hook contributions plus a path reference to a governed MCP server contribution, and `protocolConfigToVSCodeMcpConfig` produces that `.mcp.json` content separately using VS Code's `${input:id}` reference syntax so no token value is ever emitted; no `HarnessAdapter` ships for `vscode` in this phase (VS Code's agent mode has no documented headless CLI to exec).
--- a/packages/harnesses/docs/vscode-agent-plugin.md
+++ b/packages/harnesses/docs/vscode-agent-plugin.md
@@ -1,0 +1,66 @@
+# VS Code Agent Plugin
+
+Per the multi-editor harness design, `@revealui/harnesses` can generate a VS
+Code agent-plugin bundle wiring RevealUI's governed hook receipts and MCP
+data plane into VS Code's Copilot agent mode. Facts below verified
+2026-07-17 against `code.visualstudio.com` and `docs.github.com`; VS Code
+agent plugins are in Preview with a documented high release cadence --
+re-verify before relying on this doc.
+
+## What gets generated
+
+- `VSCodeGenerator` (`src/content/generators/vscode.ts`) emits
+  `.revealui/vscode-plugin/plugin.json` -- the plugin manifest, carrying only
+  the `hooks` contribution (every one of VS Code's eight documented hook
+  events runs `revealui-harnesses hook vscode`) and a string reference to
+  `.mcp.json` for the MCP contribution.
+- `protocolConfigToVSCodeMcpConfig` (`src/protocol/config-normalizer.ts`)
+  produces the `.mcp.json` content separately, wiring the governed
+  `/api/mcp` endpoint. The device token is never emitted as a literal value
+  -- it is referenced via VS Code's `${input:<id>}` substitution syntax, with
+  the `inputs[]` entry marked `password: true` so VS Code masks the prompt
+  and stores the value itself.
+
+## Installing from a local path
+
+VS Code plugins are not auto-discovered from inside a project the way
+Cursor's `.cursor/` or OpenCode's `.opencode/` are. Register the generated
+bundle explicitly in VS Code user or workspace settings:
+
+```json
+{
+  "chat.pluginLocations": {
+    "/absolute/path/to/project/.revealui/vscode-plugin": true
+  }
+}
+```
+
+Set the value to `true` to enable the plugin, `false` to keep it registered
+but disabled. Marketplace publishing (`chat.plugins.marketplaces`, "Chat:
+Install Plugin From Source") is owner-gated and out of scope here -- see the
+design doc's D-C decision.
+
+## Org-allowlist governance
+
+VS Code's per-plugin registration above is a developer-machine setting; it
+does not control what an ENTERPRISE permits at the MCP-server level.
+GitHub Copilot Business/Enterprise organizations govern that separately,
+through GitHub's own admin console (not a VS Code setting):
+
+1. An org or enterprise administrator opens **AI controls -> MCP** in the
+   GitHub Copilot policies page.
+2. They choose one of two access-control modes:
+   - **Allow all** -- no restriction; any MCP server a developer configures
+     (including RevealUI's `/api/mcp`) can be used.
+   - **Registry only** -- only MCP servers listed in the org/enterprise's
+     internal MCP registry may run. Under this mode, RevealUI's governed MCP
+     endpoint must be added to that internal registry before any developer's
+     generated `.mcp.json` entry will actually be usable at runtime, even
+     though VS Code will still let a developer configure it locally.
+3. Policy resolution is tied to the GitHub Copilot seat assignment; a
+   developer holding seats in multiple orgs gets a single resolved policy.
+
+Source: docs.github.com/en/copilot/how-tos/administer-copilot/manage-mcp-usage/configure-mcp-server-access
+(verified 2026-07-17). Publishing RevealUI's MCP entry into a customer's
+internal registry is a customer-side administrative action; this package
+does not automate it.
--- a/packages/harnesses/src/__tests__/cli-hook.test.ts
+++ b/packages/harnesses/src/__tests__/cli-hook.test.ts
@@ -115,12 +115,59 @@ describe('revealui-harnesses hook <source> (CLI end to end)', () => {
   }, 25_000);
 
   it('rejects an unsupported source with exit code 1', () => {
+    // opencode has no hook normalizer -- OpenCode has no hook system to
+    // normalize from (see hooks/normalizers/index.ts's dispatch doc comment).
     const result = runHookCli(
-      'vscode',
+      'opencode',
       { hook_event_name: 'x' },
       { REVEALUI_HOOK_SPOOL_PATH: spoolPath, REVEALUI_POLICY_SNAPSHOT_PATH: snapshotPath },
     );
     expect(result.exitCode).toBe(1);
     expect(result.stderr).toContain('Unsupported hook source');
+  }, 25_000);
+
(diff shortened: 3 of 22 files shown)


RevealUIStudio:
## Review (Opus review pass on the Sonnet build)

Reviewed the vscode surface against the three invariants and read the CodeQL result as a real finding (it came back clean on this diff, no alerts).

**I-4 (structurally key-less) — PASS.** `protocolConfigToVSCodeMcpConfig` builds the Authorization header purely from the token input-id NAME using VS Code's `inputs[]` promptString + password mechanism and `${input:id}` reference; it never reads `config.environment.variables` and never emits a literal. Matches the leak-proof Cursor/OpenCode precedent.

**I-1 (identity from token only) — PASS.** The normalizer promotes nothing from the payload onto authorization identity: it carries only `session_id` as a correlation `conversationId` (the same class as Cursor's), leaves modelId/generationId undefined, and quarantines the full payload under `raw`. The code and its doc note both state that any future VS Code identity field stays in `raw` only.

**I-5 (honest enforcement tier) — PASS.** The `enforcementTier = 'advisory'` const from Phase B is intact, so the vscode source inherits the honest advisory tier; no `enforced` claim was reintroduced.

**Adapter-vs-generator call — agree.** No headless VS Code CLI exists to drive an agent turn (Copilot CLI is a separate product), so scoping to normalizer + CLI source + generator + detection metadata, with no fake `execute` path, is correct and matches the design's explicit fallback.

**Carry-forward to Phase E (not blocking):** the exact VS Code hook wire-format (snake_case envelope, the `hookSpecificOutput.permissionDecision` response shape) is medium-confidence because the docs were read via an AI-summarizing fetch, and VS Code agent hooks are in Preview. The convergence evidence (VS Code reuses the `${CLAUDE_PLUGIN_ROOT}` token and Claude-Code-style snake_case) makes it plausible, but the real-editor e2e in Phase E must pin it, same posture as the deferred OpenCode output-shape pin. Tracked for that phase.

Verdict: **APPROVE.** No fixes needed. Typecheck, biome, CodeQL, and 438/438 tests green (red-first proof in the PR body). Independent review, builder was a separate session. The merge is the owner's.

RevealUIStudio:
## Review update — folded in the shared fail-closed fix (commit 834d30eaf)

A separate Fable session found a real defect in the shared `runHookCommand` path (the one this PR extends): when the receipt **spool write** fails (unwritable data dir on first run, disk full, read-only fs, or a non-ENOENT rotate error), `runHookCommand` threw *after* computing the decision, the CLI exited with no response, and the editor default (allow) won, silently bypassing a computed `deny`. That is a fail-open on the enforcement path, and because `run-hook.ts` is shared it applied to the vscode source this PR adds as much as to the already-merged cursor/claude-code sources.

Since this PR already touches `run-hook.ts`, I folded the fix in here rather than spin a separate PR, so one change covers all three sources:

- `runHookCommand` now catches a spool-write failure, warns on stderr, and **still delivers the decision + editor response** (fail CLOSED). A `spooled` flag on the result records that the receipt was dropped (the availability/audit-completeness cost), while enforcement is preserved.
- Red-first proof: reverted `run-hook.ts` to the pre-fix commit, the new test `fails CLOSED: a deny is still delivered when the spool write fails` (unwritable spool + deny-matched event, asserts decision + exit code 2 still delivered) went **red**; restored, green. Full suite 439/439, typecheck + biome + pre-push gate green.

This closes the residual the other session recorded against the merged Phase B (#1940) for all sources at once, so that fast-follow does not need its own PR. GAP-386 (harness paths missing from the security gate's SECURITY_PATHS) is separate and stands.

Verdict remains **APPROVE** (invariants I-1/I-4/I-5 pass, CodeQL clean, adapter-vs-generator call agreed). Independent review; the two build sessions were separate. Merge is the owner's.

PR #1949, opened 2026-07-17
fix(security): classify the harnesses enforcement plane as security-sensitive

RevealUIStudio:
## Summary

The harness hook/policy enforcement plane (`packages/harnesses/src/hooks/`: deny/ask policy evaluation, editor permission responses, receipt spool, and the generators that emit the config binding editors to it) decides allow/deny for editor agent actions, and the package's server surface forks caller-supplied agent commands. Neither security-review gate covered any `packages/harnesses/` path, so an enforcement-plane PR merged over an outstanding request-changes without the gate holding it.

Adds `packages/harnesses/` as a whole-package entry to `SECURITY_PATHS`, matching how `packages/mcp/`, `packages/security/`, and `packages/paywall/` are already handled. Whole-package (rather than a subpath list) closes the drift class permanently. Kept in lockstep with the fleet checker (paired private-repo change).

## Verification

- Syntax valid; `--diff` post-commit correctly self-classifies this gate change as security-sensitive (self-protection entry).
- The identical entry in the fleet checker was verified against a real PR touching `packages/harnesses/`: it flipped from "no security-sensitive files" to a HOLD requiring a recorded verdict.

This PR is itself security-sensitive (edits the gate), so it needs a recorded reviewer verdict before merge.

### DIFF
--- a/scripts/validate/security-review-gate.cjs
+++ b/scripts/validate/security-review-gate.cjs
@@ -95,6 +95,14 @@ const SECURITY_PATHS = [
   '.github/workflows/security-review-gate',
   '.github/workflows/sec-audit-label-guard',
   '.github/workflows/security.yml',
+  // Harness hook/policy enforcement plane. packages/harnesses/src/hooks/
+  // decides ALLOW/DENY for editor agent actions and emits the config that
+  // binds editors to it; the package's server surface forks caller-supplied
+  // agent commands. Added after an enforcement-plane PR merged over an
+  // outstanding request-changes because this list did not classify it as
+  // security-sensitive. Whole-package entry, matching packages/mcp//security//
+  // paywall/; adapters ride along fine. Keep in lockstep with the fleet checker.
+  'packages/harnesses/',
 ];
 
 // A recorded reviewer verdict = an approving GH review OR one of these labels.


RevealUIStudio:

Independent **non-author** review — I did not author #1949 or its lockstep sibling revealui-jv#512 (a different Fable session did). Verified every claim against the files.

**Diff scope (both sides).** #1949 adds exactly one whole-package entry `'packages/harnesses/'` (+ comment) to `SECURITY_PATHS` in `scripts/validate/security-review-gate.cjs` — no logic change, no deletions, no scope creep. jv#512 adds the byte-identical entry to the fleet checker `scripts/security-pr-review-check.js`. The list entries match; only the comments differ (jv#512's is fuller, naming the GAP-353 agent.spawn path).

**Scope judgment — whole-package is the right call.** It matches the established pattern (`packages/security/`, `packages/auth/`, `packages/mcp/`, `packages/paywall/` are all whole-package entries) rather than inventing a subpath list, and whole-package is fail-closed-toward-review: it permanently kills the drift class GAP-386 is about (a new file under `packages/harnesses/hooks/…` can never silently escape classification). Adapters riding along get reviewed too — acceptable.

**No matching-logic regression.** Both gates classify by `for (const marker of SECURITY_PATHS) if (f.includes(marker))` — a pure substring test. A new marker can only ADD classifications for paths containing `packages/harnesses/`; it cannot change the verdict for any other path. Zero false-positive/false-negative risk introduced.

**Empirical proof the fix closes the real miss (run firsthand):**
- revealui#1947 (the vscode agent-plugin generator + hook normalizers) touches files under `packages/harnesses/` and **zero** other security paths.
- With jv#512's entry now on `test`, the fleet checker classifies #1947 as security-sensitive and gates it. Before the entry, #1947 matched no security path → "no security-sensitive files" — exactly the #1940 miss (an enforcement-plane PR merging over an outstanding REQUEST-CHANGES). The entry demonstrably converts that miss into a HOLD.

**The red "Security review gate: FAILURE" on #1949 is the gate protecting itself, not a defect.** #1949 edits `scripts/validate/security-review-gate.cjs`, matched by the existing `scripts/validate/security-review-gate` entry, so the gate classifies its own PR as security-sensitive and HOLDs until a recorded verdict/label. Confirmed via the fleet checker (HOLD, `Touched: scripts/validate/security-review-gate.cjs`). It clears on the owner's `sec-review:approved` label.

**Lockstep note:** jv#512 already MERGED (22:01Z); #1949 is the remaining half. Until #1949 lands, the in-repo gate lacks the harnesses entry while the fleet checker has it — a brief one-sided window, mitigated because the fleet checker is the authoritative cross-check (per dup-work-and-review-gates.md). Merging #1949 closes it.

Clearing still requires the OWNER's `sec-review:approved` label (I do not apply it). Recommend merging #1949 to complete the lockstep pair.

PR #1950, opened 2026-07-17
test(db): guard audit db-connection resolution against config override

RevealUIStudio:
## Context

The E2E Smoke job on the test→main promotion went red after the audit-storage batch landed on `test`. I investigated all three suspected causes against the actual failing CI log and the current code.

## The three causes, verified

**Cause 1 — schemaless smoke DB (the real blocker).** The E2E Smoke job creates `POSTGRES_DB=smoke` with no migration step. The audit batch made the API server run a boot self-test that writes to `audit_log` and refuses to serve on failure. In the failing run the server logged `Audit write failed ... insert into "audit_log"` → "Startup validation failed; exiting", so `/health/live` never came up and "Wait for servers" timed out at 240s. **Already fixed by #1948** (migrate the smoke DB before starting servers + switch to `pgvector/pgvector:pg16`, required because `0000_init.sql` runs `CREATE EXTENSION vector`). This PR does not duplicate that fix.

**Cause 2 — edge-runtime `process.exit` — not a regression, not the blocker.** The `process.exit(1)` calls flagged in the build are the pre-existing license-boot and env-validation blocks in `apps/admin/src/instrumentation.ts` (present when E2E was last green). In the failing run the build still completed (route listing printed, all three servers started), so these messages did not fail the build. `@revealui/security`'s default barrel is edge-safe (audit/`node:crypto` lives behind the `/server` subpath) and instrumentation imports only the client-safe barrel. No change needed.

**Cause 3 — getClient resolving the wrong DB — not real in code.** `getDatabaseConfig` returns `POSTGRES_URL || DATABASE_URL || ''` and never synthesizes a default database, and `getClient()` falls back to `POSTGRES_URL`/`DATABASE_URL` when config is empty. There is no path that resolves to database `"test"`. The repeating `FATAL: database "test" does not exist` in the log every 10s is the Postgres service container's own healthcheck (`pg_isready -U test` with no `-d` defaults dbname to the user "test") — benign harness noise, not app code, present on green runs too. #1945 (merged) added only a synchronous env-presence assertion, not a resolution change.

## What this PR adds

A focused, isolated regression guard at the `getClient()` layer — the actual audit-storage connection path — which was previously untested for the resolved connection value:

- connects to the `POSTGRES_URL` database when config supplies no url (the exact E2E harness scenario)
- falls back to `DATABASE_URL` when `POSTGRES_URL` is absent
- honors a config-supplied url (which is itself only ever derived from those vars)
- throws a named error when neither is set

This locks the invariant that a set connection URL is never silently overridden by a config default. Prod resolution is unchanged (`config.database.url` equals `POSTGRES_URL`/`DATABASE_URL` in every environment).

## Verification

- `pnpm --filter @revealui/db exec vitest run src/client/__tests__/` → 11 passed (4 new + existing).
- Proven to fail: injecting a config-default→`test` override makes the E2E-path assertion go red (expected `/smoke`, got `/test`), then reverted.
- `biome check` clean; `pnpm --filter @revealui/db typecheck` clean.
- Fleet security self-check (`--diff origin/test`): no security-sensitive files.
- Test-only; no runtime behavior change, no published-package behavior change (no changeset).

**The real proof is a green E2E Smoke on the re-run promotion once #1948 lands** — that is the load-bearing fix for the actual blocker; this PR is the complementary regression guard so a connection-resolution regression fails a PR check rather than only surfacing at the promotion.

### DIFF
--- a/packages/db/src/client/__tests__/connection-resolution.test.ts
+++ b/packages/db/src/client/__tests__/connection-resolution.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Connection-resolution regression guard.
+ *
+ * The audit write path installs storage through `getClient()` (see
+ * apps/server/src/lib/audit-storage.ts). When the audit self-test cannot reach
+ * the database the server refuses to boot, so it is load-bearing that
+ * `getClient()` connects to exactly the database the environment names and is
+ * never silently redirected to a different one.
+ *
+ * These tests pin that invariant at the client layer: a set POSTGRES_URL (or
+ * DATABASE_URL) is resolved verbatim into the pg pool, and `@revealui/config`
+ * only ever supplies a URL it derived from those same vars (getDatabaseConfig
+ * never synthesizes a default database — proven in
+ * packages/config/src/__tests__/modules.test.ts). Together they guarantee the
+ * resolver honors an explicit URL over any config value in every environment.
+ */
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+const { mockConfig, poolConstructor } = vi.hoisted(() => ({
+  mockConfig: { database: undefined as { url?: string } | undefined },
+  poolConstructor: vi.fn(),
+}));
+
+vi.mock('@revealui/config', () => ({ default: mockConfig }));
+
+vi.mock('pg', () => ({
+  Pool: class {
+    constructor(config: { connectionString?: string }) {
+      poolConstructor(config);
+    }
+    on(): void {}
+  },
+}));
+
+vi.mock('drizzle-orm/node-postgres', () => ({
+  drizzle: vi.fn(() => ({ query: {} })),
+}));
+
+vi.mock('drizzle-orm/neon-http', () => ({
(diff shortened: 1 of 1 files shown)


RevealUIStudio:
## Non-author review — APPROVE (test-only regression guard)

Reviewed as a non-author (a separate session opened this). This is a clean, well-targeted addition:

- One new test file (`connection-resolution.test.ts`, +109/-0), no source change, no changeset. Pins the invariant that `getClient()` resolves a set `POSTGRES_URL`/`DATABASE_URL` verbatim into the pool and `@revealui/config` never synthesizes a default database — the exact "connects to /smoke, never the pg-username default" property the E2E audit path depends on.
- Correctly reframes the promotion blocker: the only real failure was the schemaless smoke DB (fixed + merged in #1948, which adds the migrate step + pgvector image), and #1944's E2E Smoke is now green. This PR is the regression guard, not the fix.
- Not on any security-sensitive path (test-only, `packages/db`), so no guardrail-2 gate applies.

Sound to merge once CI is green. Merge is the owner's.

PR #1961, opened 2026-07-18
feat(db): audit_log ONE DOOR — single write path + CI enforcer (GAP-355 Stage 2)

RevealUIStudio:
## GAP-355 Stage 2 — audit_log ONE DOOR

Builds the Stage 2 half of the audit-receipt architecture (ADR `2026-07-12-audit-receipt-architecture.md` §5): exactly one code path that writes `audit_log`, so every later integrity property (append-only, per-row signing in Stage 3, anchoring in Stage 4) has a single chokepoint.

### What changed
- **Collapsed six external writers onto `DrizzleAuditStore`** — the revmarket task executor, the api-keys credential-event writer, the admin setup route, the admin bootstrap CLI, and the governed-MCP receipt signer all now call `DrizzleAuditStore.append(...)` instead of `db.insert(auditLog)`.
- **Extended `append`/`appendBatch`** with optional `signature`/`previousSignature` pass-through, so the MCP signer routes through the door while still computing its own signature (signing moves into the store in Stage 3). Absent → NULL, unchanged from Stage 1.
- **New CI enforcer** `scripts/validate/audit-log-single-door.ts` (AST via the TS compiler API, zero authored regex; mirrors the Stage 0 `as-never-values.ts`) fails the build on any `insert(auditLog)` outside the store. Wired hard-fail into `ci-gate.ts` + `ci.yml`. Empty allowlist.

### The enforcer earned its keep immediately
My manual grep found 5 writers; the enforcer flagged a **6th** I'd missed (`scripts/admin/bootstrap.ts`) — a live demonstration that door number seven really does hide, which is why the mechanical check exists.

### Verification
- Enforcer: RED on the un-collapsed tree (caught `bootstrap.ts`), GREEN after (0 violations).
- Unit test `audit-log-single-door.test.ts` (7/7): fires on the `insert(auditLog)` pattern, passes on the `.append(...)` pattern and on reads — durable red/green proof.
- Runtime write path: `audit-storage.pglite.test.ts` + `mcp-audit-concurrency.pglite.test.ts` (12/12) — the MCP signed-chain concurrency test proves the signature pass-through works through the store end to end.
- `@revealui/db` typecheck clean; db `audit-store.integration.test.ts` 4/4.

### Scope
PR-1 of Stage 2. Deferred to PR-2 (separate review): the schema migration — monotonic sequence + tenant column + REVOKE UPDATE/DELETE.

### Security gate
Touches `routes/api-keys.ts` (security-sensitive). Per guardrail 2 this needs a recorded **non-author** verdict before merge — I authored it, so it stops here for the reviewer. I do not self-merge.

### DIFF
--- a/.changeset/gap-355-audit-one-door.md
+++ b/.changeset/gap-355-audit-one-door.md
@@ -1,0 +1,5 @@
+---
+'@revealui/db': minor
+---
+
+audit_log ONE DOOR (GAP-355 Stage 2): `DrizzleAuditStore.append`/`appendBatch` now accept optional `signature`/`previousSignature` pass-through, so every `audit_log` writer routes through the single store (a new CI enforcer fails the build on any `insert(auditLog)` outside it). Absent values persist NULL, unchanged from Stage 1. Signing itself moves into the store in Stage 3.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,9 @@ jobs:
       - name: Empty-catch validation (hard fail)
         run: pnpm validate:empty-catch
 
+      - name: audit_log ONE DOOR validation (hard fail)
+        run: pnpm validate:audit-one-door
+
       - name: Seed script wiring validation (hard fail)
         run: pnpm validate:seed-wiring
 
--- a/apps/admin/src/app/api/setup/route.ts
+++ b/apps/admin/src/app/api/setup/route.ts
@@ -95,14 +95,15 @@ export async function POST(request: Request): Promise<NextResponse<BootstrapResu
       // Audit log — non-fatal.
       try {
         const { hostname } = await import('node:os');
-        const { auditLog } = await import('@revealui/db/schema');
+        const { DrizzleAuditStore } = await import('@revealui/db');
         const { classifyAuditWriteFailure, recordAuditWriteResult } = await import(
           '@revealui/core/security'
         );
         const eventId = crypto.randomUUID();
         try {
-          await db.insert(auditLog).values({
+          await new DrizzleAuditStore(db).append({
             id: eventId,
+            timestamp: new Date(),
             eventType: 'admin.bootstrap.completed',
             severity: 'info',
             agentId: 'web',
--- a/apps/server/src/lib/mcp-audit.ts
+++ b/apps/server/src/lib/mcp-audit.ts
@@ -41,8 +41,7 @@
 
 import { createHmac, randomUUID } from 'node:crypto';
 import { classifyAuditWriteFailure, recordAuditWriteResult } from '@revealui/core/security';
-import { getClient } from '@revealui/db';
-import { auditLog } from '@revealui/db/schema';
+import { DrizzleAuditStore, getClient } from '@revealui/db';
 
 export type McpAuditOutcome = 'invoked' | 'denied' | 'failed';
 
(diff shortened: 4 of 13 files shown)


RevealUIStudio:

Independent, non-author adversarial review of the GAP-355 Stage 2 ONE DOOR PR (head `b613613fe`). I read the full diff and the enforcer end to end, checked conformance against the ADR §5 one-door requirement and the recorded Stage-3 signing design (`GAP-355-stage-3-signing-design.md`), and ran empirical probes in a clean worktree. The collapse is complete, the enforcer fires, CI wiring is hard-fail at both layers, and the signature pass-through matches the Stage-3 interim contract. Two non-blocking advisories below.

### Scope check
Exactly the stated surface: six writer collapses, the store's optional `signature`/`previousSignature` pass-through, the enforcer + unit test + empty allowlist, CI wiring (`ci.yml`, `ci-gate.ts`, `package.json` script), and a changeset. No drive-by changes.

### Empirical verification (firsthand, worktree on the PR head)
- Independent sweep: zero `insert(auditLog)` outside `packages/db/src/audit-store.ts` across `apps/`, `packages/`, `scripts/` (excluding tests), and zero raw `INSERT INTO audit_log` SQL anywhere. The claimed collapse is real.
- Enforcer on the clean tree: PASS (0 violations).
- Planted `apps/server/src/probe-direct.ts` with `getClient().insert(auditLog)`: FAIL (1 writer), correct file:line named. Removed it: PASS again. Live red/green confirmed, on top of the 7/7 unit test (also run locally).
- Fleet security checker on the PR: HOLD until the label, as expected (`routes/api-keys.ts` is on the sensitive list).

### Semantics check on all six collapses
- Every call site keeps its own try/catch plus `classifyAuditWriteFailure`/`recordAuditWriteResult` telemetry, so Stage 0's loud-failure property is intact through the door.
- `recordCredentialEvent` keeps its `db` parameter, so transactional callers retain atomicity through the store.
- The MCP chain semantics are preserved: signature still computed in place and passed as a value, `lastMcpSignature` updated only on successful insert, `taskId`/`sessionId` null-mapping identical to the removed direct insert.
- Absent signature fields persist NULL: the honest Stage-1 state for unsigned rows, and exactly the interim contract the Stage-3 design specifies (signing moves into an injected `AuditRowSigner` later; `packages/db` stays crypto-free).

### Advisory 1 (non-blocking): alias evasion in the enforcer
I also planted `apps/server/src/probe-alias.ts` using `import { auditLog as al }` + `insert(al)`: NOT flagged (confirmed live). The syntactic matcher recognizes only the literal name, the same trade-off `as-never-values.ts` accepts and the header documents. Acceptable because the enforcer targets accidental drift, and nobody accidentally aliases an import to write an audit row; an adversarial insider can edit the enforcer itself. Cheap hardening for PR-2 or a follow-up, still zero-regex and no type-checker: collect local alias names from `ImportSpecifier`s whose imported name is `auditLog` and match those as first-argument identifiers too.

### Advisory 2 (non-blocking): two writers moved from DB-clock to app-clock timestamps
`audit_log.timestamp` carries `defaultNow()`. The admin setup route and the bootstrap CLI previously omitted `timestamp` (DB clock); they now pass `new Date()` (app clock) because `AuditEntry.timestamp` is required. Uniform app-clock stamping through one door is defensible, and PR-2's monotonic sequence will own ordering anyway. Recording the delta so it is a decision rather than an accident; if DB-clock is preferred, make `timestamp` optional in `AuditEntry` and let the default apply.

### Verdict
APPROVE. Correct, minimal, conforms to ADR §5 and the recorded PR-1/PR-2 split; the honesty line holds (no claim above what shipped). The enforcer earned its keep once already (caught the 6th writer the author's grep missed) and my probes confirm it will keep doing that job.

Owner disposition (label is the gate-clearing act; one command per line):

```
cd ~/revfleet/revealui && gh pr edit 1961 --add-label "sec-review:approved"
```
```
cd ~/revfleet/revealui && gh pr merge 1961 --squash
```

PR #1962, opened 2026-07-18
fix(admin,marketing): stop leaking dev-only DB setup + raw errors, brand global-error

RevealUIStudio:
## Summary
MASTER_PLAN P1 error-page bundle (3 items, one PR per the plan's suggested grouping):

- **`apps/admin/src/app/(backend)/error.tsx:49-64`** leaked `POSTGRES_URL`/`DATABASE_URL`/`pnpm db:migrate`/`pnpm db:seed` setup instructions to any visitor hitting a DB error, in production too. Gated behind `process.env.NODE_ENV !== 'production'`; production visitors now see a generic "we're having trouble reaching our systems" message.
- **`apps/admin/src/app/global-error.tsx:45-46`** was unbranded (`<div style={{fontFamily:'system-ui, sans-serif'}}><h2>Something went wrong!</h2>`). Replaced with the cobalt design-system pattern already used by the sibling `(backend)/error.tsx` and `(frontend)/error.tsx`. `global-error.tsx` sits outside every route group's own layout (it replaces the root `<html>/<body>` on error), so it had zero Tailwind CSS pipeline; it now imports `(backend)`'s CSS entry (`tailwindcss` + `tokens.css` + the cobalt `@theme` bridge) so the branded utility classes actually resolve at runtime — verified by grepping the compiled client chunk for `bg-destructive/10`, `bg-primary`, `text-foreground`, `text-muted-foreground`.
- **`apps/marketing/app/components/ErrorBoundary.tsx:44`** rendered the raw thrown `error.message` to the user. It was already forwarded to Sentry via `captureRenderError` in `componentDidCatch`; the fallback now shows a static branded message only. Removed the now-dead error-carrying `State` field.

## Test plan
- [x] `pnpm --filter admin typecheck` clean
- [x] `pnpm --filter marketing typecheck` clean
- [x] `pnpm --filter marketing test` — 12 files / 78 tests passed
- [x] `pnpm build` — full monorepo build green; manually verified the compiled `global-error` client chunk contains the branded Tailwind classes (`bg-destructive/10`, `bg-primary`, `text-primary-foreground`, `text-foreground`, `text-muted-foreground`)
- [x] `npx biome check` on touched files — clean
- [x] `pnpm gate:quick` — PASS (0 violations)

### DIFF
--- a/apps/admin/src/app/(backend)/error.tsx
+++ b/apps/admin/src/app/(backend)/error.tsx
@@ -47,21 +47,28 @@ export default function BackendError({
       </h2>
 
       {isDatabaseError ? (
-        <div className="flex max-w-lg flex-col gap-3 text-center text-sm text-muted-foreground">
-          <p>The admin dashboard cannot reach the database. To fix this:</p>
-          <ol className="list-inside list-decimal text-left text-xs text-muted-foreground">
-            <li>
-              Set <code className="text-foreground">POSTGRES_URL</code> or{' '}
-              <code className="text-foreground">DATABASE_URL</code> in your environment
-            </li>
-            <li>
-              Run <code className="text-foreground">pnpm db:migrate</code> to create tables
-            </li>
-            <li>
-              Run <code className="text-foreground">pnpm db:seed</code> for sample content
-            </li>
-          </ol>
-        </div>
+        process.env.NODE_ENV !== 'production' ? (
+          <div className="flex max-w-lg flex-col gap-3 text-center text-sm text-muted-foreground">
+            <p>The admin dashboard cannot reach the database. To fix this:</p>
+            <ol className="list-inside list-decimal text-left text-xs text-muted-foreground">
+              <li>
+                Set <code className="text-foreground">POSTGRES_URL</code> or{' '}
+                <code className="text-foreground">DATABASE_URL</code> in your environment
+              </li>
+              <li>
+                Run <code className="text-foreground">pnpm db:migrate</code> to create tables
+              </li>
+              <li>
+                Run <code className="text-foreground">pnpm db:seed</code> for sample content
+              </li>
+            </ol>
+          </div>
+        ) : (
+          <p className="max-w-md text-center text-sm text-muted-foreground">
+            We're having trouble reaching our systems. Our team has been notified. Please try again
+            shortly.
+          </p>
+        )
       ) : (
         <p className="max-w-md text-center text-sm text-muted-foreground">
           {isNetworkError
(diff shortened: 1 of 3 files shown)


RevealUIStudio:
## Fixed the blocking bug from review

**Root cause confirmed**: `global-error.tsx` imported only `(backend)/custom.css`, whose `@theme inline` block maps `bg-background`/`bg-primary`/etc. to `var(--rvui-*)` but never *defines* those custom properties — that only happens in `tokens.css`, which `(backend)/layout.tsx` loads separately. Since `global-error.tsx` replaces the root layout, no layout ever runs, so every branded utility resolved against an undefined custom property.

**Fix**: added `import '@revealui/presentation/tokens.css';` above the `custom.css` import, matching `(backend)/layout.tsx`'s own import order.

**Also folded in** (non-blocking note): `(backend)/error.tsx`'s heading still said "Database connection failed" unconditionally outside the dev/prod gate on the body copy. Genericized it in the production branch too.

## Runtime-grade verification (not grep)

My original verification only confirmed the Tailwind utility classes were *generated* in the compiled bundle, which the review correctly pointed out doesn't prove the `var()` references *resolve*. This time:

1. Rebuilt (`pnpm --filter admin build`) and read Next's own `page_client-reference-manifest.js` for the `global-error` entry to find the exact two CSS chunks it attaches to that route (`entryCSSFiles`).
2. Built a static HTML fixture with the exact class names from `global-error.tsx`'s JSX, loading those exact two compiled CSS files, and opened it in a real headless Chromium via Playwright (`page.evaluate(() => getComputedStyle(el))`).
3. Compared **before** (only the utilities chunk loaded, reproducing the pre-fix bug) vs **after** (both chunks loaded, matching the fix):

   - Before: every element's `backgroundColor` / `color` computed to `rgba(0, 0, 0, 0)` / `rgb(0, 0, 0)` — the exact "white bg, fill-less button" failure mode described in review.
   - After (dark color-scheme emulation): body `background-color: lab(3.58546 0.107899 -8.54721)` (matches tokens.css's `--rvui-surface-0` value verbatim), button `background-color: lab(50.9876 -13.2611 -46.4997)` (matches `--rvui-brand`), text colors likewise resolved to real values.
   - After (light color-scheme emulation): the light `prefers-color-scheme` override in tokens.css correctly swaps in light values instead — confirming the page follows the same OS-preference handling as the rest of the app, not a hardcoded dark render.

Not committing the throwaway HTML/CSS fixtures or the Playwright script — they lived in the scratchpad, not the repo.

Not merging — awaiting re-review.

PR #1968, opened 2026-07-18
feat(harnesses): gateway authn schema + SIGTERM-grace-SIGKILL termination (GAP-353 PR-1)

RevealUIStudio:
## GAP-353 PR-1 — gateway authn schema + real process termination

Builds PR-1 of the GAP-353 program to the recorded spec (`GAP-353-bootstrap-authn-design.md`): the durable authn **substrate** (§D3) plus real SIGTERM→grace→SIGKILL **termination** (§D6, incl. §7 R2). This is the lowest-risk slice; termination is independent of authn. PR-2 (fail-closed `checkAuth`, token persistence wiring, rate limit) and PR-3 (code retirement + docs) are separate and out of scope here — no caller wires the new store methods yet.

### What changed, per spec section

**§D3 — schema + `DaemonStore` methods (substrate only)** — `packages/harnesses/src/storage/`
- Two additive PGlite tables matching the spec SQL exactly:
  - `gateway_bootstrap` — singleton row (`id INTEGER PRIMARY KEY DEFAULT 1 CHECK (id = 1)`, `secret_hash`, `created_at`), holds the SHA-256 hash of the `0600` bootstrap secret.
  - `gateway_tokens` — `token_hash` PK, `issued_at`, nullable `expires_at`, `revoked_at`, `label`.
- Six `DaemonStore` methods following the existing store style (parameterized queries, `getDb()`, `Promise<...>` return types): `putBootstrapSecretHash` (upsert — supports the §7 R5 fresh-DB-re-adopts-existing-file case), `getBootstrapSecretHash`, `insertToken`, `findValidToken(hash)` (excludes revoked + expired), `revokeToken`, `pruneExpiredTokens` (expired only; revoked rows retained for audit).
- No caller wired — PR-2 consumes these.

**§D6 — real process termination (also §7 R2: BOTH SIGTERM sites)** — `packages/harnesses/src/server/spawner-service.ts`
- `stop()` and `stopAll()` both now go through a shared `terminate(proc)`: `child.kill('SIGTERM')`, do **not** set status synchronously (stays `running`), arm a bounded grace timer, and on expiry `child.kill('SIGKILL')`.
- The `close` handler remains the **single source of truth** for terminal status; if close fires first it clears the timer, so status is fully event-driven and `agent.list` never reports a live child as `stopped`.
- Grace period is configurable via `SpawnerConfig.terminationGraceMs`, default **5000ms** (spec §6 C-C recommendation).
- The escalation timer is `unref()`'d and cleared on close, so it can neither keep the process alive nor fire after exit. `terminate` is idempotent while a grace window is open.
- §7 R2 addressed: `stopAll` (daemon shutdown) gets the identical treatment, so shutdown neither wedges a SIGTERM-ignoring child nor lies about status.

### Red/green proof (required)

Tests were proven red against base before the fix. After committing, base `spawner-service.ts` was restored with `git checkout origin/test -- packages/harnesses/src/server/spawner-service.ts` (never `git stash pop`), the termination tests were run, then the fix was restored with `git checkout HEAD -- ...`.

**Red run (base source, all 3 termination tests FAIL):**
```
FAIL  SpawnerService termination (D6) > escalates SIGTERM to SIGKILL for a child that ignores SIGTERM
FAIL  SpawnerService termination (D6) > keeps status running immediately after stop() on a live child
      AssertionError: expected 'stopped' to be 'running'
FAIL  SpawnerService termination (D6) > stopAll escalates and does not lie about status
      AssertionError: expected 'stopped' to be 'running'
 Tests  3 failed (3)
```
Base fails two ways at once: it sets `status = 'stopped'` **synchronously** (the status lie), and it never escalates, so the SIGTERM-ignoring fixture never exits (the escalation test hangs to timeout). Both are exactly what D6/R2 fix.

**Green run (fix restored):** `Tests  3 passed (3)`.

The three termination tests use a real committed fixture (`ignore-sigterm.mjs`) that traps/ignores SIGTERM and stays alive on a `setInterval` (with a 30s self-exit backstop so a non-escalating run cannot orphan it); `node:child_process` `spawn` is mocked to launch it regardless of backend. A short configured grace (~100-150ms) is used in tests, not 5s.

### Test counts

- New tests: **3** termination (`spawner-service.test.ts`) + **9** gateway-substrate round-trip (appended to `daemon-store.test.ts`: bootstrap put/get, singleton overwrite, insert/find, unknown-hash, revoked-excluded, revoke-returns-false, expired-excluded, future-expiry-valid, prune-expired-only) = **12 new**.
- Full package suite: **450 passed, 1 failed (451)**. The one failure is **pre-existing and environmental**, unrelated to this diff: `opencode-adapter.test.ts` asserts `opencode` is not installed, but this build machine has `opencode 1.18.3` on PATH. Verified the opencode adapter test + source are byte-identical to `origin/test` (empty `git diff`), and this PR touches no opencode code. Not fixed here per the "don't fix unrelated pre-existing failures" instruction.
- `pnpm --filter @revealui/harnesses typecheck`: clean.

### Security-sensitive surface

This PR touches `packages/harnesses/` — on the guardrail-2 security-sensitive list. The fleet self-check (`security-pr-review-check.js --diff origin/test`) returns **HOLD**, which is expected and correct: it needs a **recorded non-author guardrail-2 verdict + `sec-review:approved` label** before merge. I did not self-approve, did not apply the label, and will not self-merge.

Changeset: `@revealui/harnesses` minor (behavior change inside 0.x = minor).

### DIFF
--- a/.changeset/gateway-authn-schema-termination.md
+++ b/.changeset/gateway-authn-schema-termination.md
@@ -1,0 +1,5 @@
+---
+"@revealui/harnesses": minor
+---
+
+add the gateway authn persistence substrate and real process termination to the harness daemon: two additive PGlite tables (`gateway_bootstrap`, a singleton hash of the HTTP-gateway bootstrap secret; `gateway_tokens`, hashed durable bearer tokens with expiry/revocation) plus `DaemonStore` methods (`putBootstrapSecretHash`, `getBootstrapSecretHash`, `insertToken`, `findValidToken`, `revokeToken`, `pruneExpiredTokens`) as substrate for a later fail-closed gateway; and `SpawnerService.stop`/`stopAll` now escalate SIGTERM to SIGKILL after a bounded, configurable grace period (`terminationGraceMs`, default 5000ms) with event-driven status, so a child that ignores SIGTERM is actually killed and `agent.list` never reports a live process as stopped. Substrate only, no callers wired yet.
--- a/packages/harnesses/src/__tests__/daemon-store.test.ts
+++ b/packages/harnesses/src/__tests__/daemon-store.test.ts
@@ -305,4 +305,72 @@ describe('DaemonStore', () => {
     const pruned = await store.pruneEvents(0);
     expect(pruned).toBeGreaterThanOrEqual(0);
   });
+
+  // ---------------------------------------------------------------------------
+  // Gateway authn substrate (GAP-353 D3)
+  // ---------------------------------------------------------------------------
+
+  it('round-trips the bootstrap secret hash', async () => {
+    expect(await store.getBootstrapSecretHash()).toBeNull();
+    const row = await store.putBootstrapSecretHash('hash-abc');
+    expect(row.id).toBe(1);
+    expect(row.secret_hash).toBe('hash-abc');
+    expect(await store.getBootstrapSecretHash()).toBe('hash-abc');
+  });
+
+  it('keeps the bootstrap secret a singleton (re-persist overwrites)', async () => {
+    await store.putBootstrapSecretHash('hash-old');
+    await store.putBootstrapSecretHash('hash-new');
+    expect(await store.getBootstrapSecretHash()).toBe('hash-new');
+  });
+
+  it('inserts and finds a valid token', async () => {
+    const inserted = await store.insertToken({ tokenHash: 'tok-1', label: 'studio-macbook' });
+    expect(inserted.token_hash).toBe('tok-1');
+    expect(inserted.label).toBe('studio-macbook');
+    expect(inserted.revoked_at).toBeNull();
+
+    const found = await store.findValidToken('tok-1');
+    expect(found?.token_hash).toBe('tok-1');
+  });
+
+  it('findValidToken returns null for an unknown hash', async () => {
+    expect(await store.findValidToken('nope')).toBeNull();
+  });
+
+  it('findValidToken excludes revoked tokens', async () => {
+    await store.insertToken({ tokenHash: 'tok-2' });
+    const revoked = await store.revokeToken('tok-2');
+    expect(revoked).toBe(true);
+    expect(await store.findValidToken('tok-2')).toBeNull();
+  });
+
+  it('revokeToken returns false when nothing was revoked', async () => {
+    expect(await store.revokeToken('absent')).toBe(false);
+  });
+
+  it('findValidToken excludes expired tokens', async () => {
+    const past = new Date(Date.now() - 60_000).toISOString();
+    await store.insertToken({ tokenHash: 'tok-3', expiresAt: past });
+    expect(await store.findValidToken('tok-3')).toBeNull();
+  });
+
+  it('honors a future expiry as still valid', async () => {
+    const future = new Date(Date.now() + 60_000).toISOString();
+    await store.insertToken({ tokenHash: 'tok-4', expiresAt: future });
+    expect(await store.findValidToken('tok-4')).not.toBeNull();
+  });
+
+  it('prunes only expired tokens', async () => {
+    const past = new Date(Date.now() - 60_000).toISOString();
+    const future = new Date(Date.now() + 60_000).toISOString();
+    await store.insertToken({ tokenHash: 'tok-expired', expiresAt: past });
+    await store.insertToken({ tokenHash: 'tok-future', expiresAt: future });
+    await store.insertToken({ tokenHash: 'tok-forever' });
+
+    const pruned = await store.pruneExpiredTokens();
+    expect(pruned).toBe(1);
+    expect(await store.findValidToken('tok-future')).not.toBeNull();
+    expect(await store.findValidToken('tok-forever')).not.toBeNull();
+  });
 });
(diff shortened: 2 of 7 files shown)


RevealUIStudio:

Adversarial review of the full diff (Fable session, non-author). Scope: TypeScript correctness, security, robustness. React/Tailwind/a11y dimensions are not applicable to this Node daemon package.

### What holds up

- All six store methods use parameterized SQL; nothing interpolates input into query text.
- Hash-only storage for both the bootstrap secret and bearer tokens; no plaintext secret ever touches PGlite. Equality-on-hash lookup is the right timing-safe comparison pattern for the later gateway PR.
- The singleton CHECK (id = 1) constraint makes the bootstrap row genuinely un-forkable, and the upsert path is the correct re-adopt semantics for a wiped data dir.
- The termination redesign fixes the status lie: close is the single terminal-status owner, stop() no longer stamps stopped synchronously, and the escalation is idempotent under a double stop(). The close-cancels-timer and timer-guards-status double protection is race-clean on a single-threaded loop.
- Test strategy is strong: a real SIGTERM-ignoring child proves the escalation empirically rather than mocking process.kill, with an orphan backstop in the fixture and an afterEach reap.
- TypeScript: no any, typed row interfaces match the schema exactly, the RETURNING casts are justified by comments.

### Warning (follow-up-sized, not blocking)

1. spawner-service.ts terminate(): the unref'd grace timer defeats the SIGKILL escalation on the daemon-shutdown drain. stopAll() fires SIGTERM, but if the event loop then empties, the daemon exits before the timer fires and a SIGTERM-ignoring child survives as an orphan. This is NOT a regression (the old code always orphaned that child, and lied about it), but the changeset's a-child-that-ignores-SIGTERM-is-actually-killed guarantee only holds while the daemon outlives the grace window. Suggested fix: keep the timer ref'd (close already clears it, so it holds the loop at most terminationGraceMs and only while a child is genuinely wedged), or give stopAll a bounded awaitable form the shutdown path can drain. Filing as a follow-up gap.

### Suggestions

2. An operator-initiated stop() of a SIGTERM-ignoring child ends as status errored (close sees code null). An intentional kill reads as a crash in agent.list. A terminating flag set in terminate() would let close map that exit to stopped.
3. stop() during an open grace window silently no-ops while stop() after exit throws; callers cannot distinguish escalation-in-progress from a fresh request. Worth a boolean return or an explicit doc note on the RPC surface.
4. gateway_tokens has no scope or created_by column. Fine for substrate, but decide before the gateway wiring PR whether all tokens are equal-privilege; adding it later is a migration.

Verdict rationale: the advisories are improvements on top of a strict improvement; nothing here regresses security or correctness, and the substrate makes no authz claims yet (no callers wired). Merge stays owner-gated per disposition rules; the sec-review:approved label is the owner's to apply.

RevealUIStudio:

Independent non-author review by the **GAP-353 design author** (I wrote the spec `GAP-353-bootstrap-authn-design.md` / jv#519; I did **not** build this PR — a peer did). Reviewed the full diff against the spec and for correctness/security; did not rely on the existing marker.

**Verified correct:**
- **§D3 schema** — `gateway_bootstrap` (singleton `id INTEGER PRIMARY KEY DEFAULT 1 CHECK (id = 1)`) and `gateway_tokens` match the spec SQL exactly. `expires_at`/`revoked_at`/`issued_at`/`created_at` are `TIMESTAMPTZ` (not text), so `findValidToken`'s `expires_at > NOW()` and `pruneExpiredTokens`'s `expires_at < NOW()` are correct timestamp comparisons — I specifically checked this because a `text` column there would be a silent lexical-comparison bug. It isn't.
- **§D3 store methods** — all six correct and parameterized (no injection). Tokens + bootstrap secret stored **hashed**, never plaintext. `findValidToken` excludes revoked + expired; `revokeToken` is idempotent (returns false on an already-revoked/absent token); `putBootstrapSecretHash` upserts (supports the §7 R5 fresh-DB-re-adopts-file case) and preserves `created_at`.
- **§D6 termination** — all six requirements met: shared `terminate()` from both `stop` and `stopAll` (§7 R2), SIGTERM with **no synchronous status write** (the status-lie is gone), bounded grace → SIGKILL, the `close` handler as the single source of truth (clears the timer), the escalation timer `unref()`'d, and idempotency while a grace window is open. The red/green proof is genuine — a real SIGTERM-ignoring fixture, base fails both ways (sync `stopped` + no escalation), fix passes.

**Two recorded notes — neither blocks merge:**

1. **Scope accuracy (not a defect):** this PR is substrate + termination only; it does **not** close the acute fail-open (`checkAuth` still grants unauthenticated `agent.spawn`/`agent.stop` on a never-paired daemon — spec F1/F2). That is correctly deferred to PR-2 (fail-closed `checkAuth` + token wiring). So merging this improves termination and lays the authn substrate but does **not** yet reduce the spawn/kill attack surface. Stating it so no one reads this as closing the exposure.

2. **`stopAll` shutdown-escalation carries the GAP-390 limitation (tracked):** the escalation timer is `unref()`'d (faithful to spec §D6), so on **daemon shutdown** — `stopAll()` then process exit — the grace timer will not fire before the process dies, orphaning a SIGTERM-ignoring child. This is strictly no worse than base (base never escalated at all) and the shutdown sequence isn't wired here, but when it is, shutdown must **await the grace window or SIGKILL synchronously**. This is exactly what the peer-filed GAP-390 captures; reconcile there.

CI is green except the expected security-gate hold. The one failing package test (`opencode-adapter`) is pre-existing + environmental (opencode installed on the build host), byte-identical to `origin/test`, untouched by this diff.

**Disposition:** owner applies `sec-review:approved` + merges. I do not self-clear or merge.

PR #1969, opened 2026-07-18
feat(db): audit_log append-only enforcement + monotonic seq + tenant (GAP-355 Stage 2 PR-2)

RevealUIStudio:
## GAP-355 Stage 2 — append-only enforcement + monotonic seq + tenant (PR-2)

The second half of Stage 2 (ADR `2026-07-12-audit-receipt-architecture.md` §5). PR-1 made `DrizzleAuditStore` the single write door; PR-2 makes append-only **real at the DB layer** and adds the two columns Stage 3/4 signing + anchoring need.

### Migration 0026
- **Append-only trigger** — a `BEFORE UPDATE OR DELETE` trigger (`enforce_audit_log_append_only`) that RAISEs (SQLSTATE 23001). This is the real enforcement.
- **`seq`** — monotonic `bigserial`, DB-assigned (the store never writes it). Fixes append order and makes deletion detectable as a gap — the primitive Stage 4 anchoring checks.
- **`tenant`** — nullable scope column for per-tenant Merkle anchoring. No backfill; `DrizzleAuditStore.append` gains an optional `tenant` pass-through.
- **REVOKE UPDATE, DELETE FROM PUBLIC** — defense-in-depth only (see below).

### Load-bearing design finding: why a trigger, not REVOKE alone
The app connects to Neon as the **table owner** (`POSTGRES_URL`). A table owner **bypasses table privileges**, so `REVOKE UPDATE/DELETE` from the owner is theater — it wouldn't stop the owner from mutating audit rows. A `BEFORE` trigger fires for every role including the owner, so it's the actual enforcement; the REVOKE is kept as defense-in-depth for any future non-owner app role. TRUNCATE is intentionally not blocked (a BEFORE DELETE trigger doesn't fire on TRUNCATE; it's an owner-only DDL reset, used by test harnesses).

### Verification (all green)
- Integration tests against PGlite with all 26 migrations applied (8/8): UPDATE **rejected** (23001), DELETE **rejected** (23001, row survives), `seq` monotonic + DB-assigned, `tenant` persists / NULL by default, plus the existing severity/CHECK/jsonb contracts (the old delete-based test cleanup was rewritten to select-by-id, since the table is now append-only).
- Write-path unaffected: `audit-storage.pglite` + `mcp-audit-concurrency` 12/12 with 0026 live.
- `@revealui/db` typecheck clean; migration-journal validator 27/27 (0026 registered in `_custom.json` as a hand-authored trigger migration); ONE DOOR enforcer still green; no generated-type drift.

### Security gate — please read
This is the **audit-integrity crown-jewel surface**, but the fleet security checker reports "no security-sensitive files" because **`packages/db` (the audit substrate) is on neither SECURITY_PATHS list**. That's a checker blind spot of the same class as GAP-386 (the harnesses plane). Treat this as guardrail-2 by intent: it needs a recorded **non-author** verdict before merge — I authored it. **Recommended follow-up:** widen both SECURITY_PATHS lists to cover `packages/db/src/schema/audit-log.ts`, `packages/db/src/audit-store.ts`, and `packages/db/migrations/**` so this surface auto-gates going forward.


RevealUIStudio:

Adversarial review of the full diff (Fable session, non-author). This is the audit-substrate crown-jewel surface, reviewed accordingly.

### What holds up

- The trigger-over-REVOKE reasoning is correct and the comment proves the author understood the threat: the app connects as table owner, owners bypass table privileges, so REVOKE alone would be theater. The BEFORE trigger fires for every role. The REVOKE stays as defense-in-depth. This is the right layering.
- ERRCODE restrict_violation (23001) is consistent across migration, comment, and tests.
- seq as DB-assigned bigserial with the store never writing it is the right primitive for Stage-4 gap detection; AuditEntry deliberately has no seq field, so the single door cannot forge order.
- Idempotency follows the 0024 precedent (IF NOT EXISTS + CREATE OR REPLACE + DROP TRIGGER IF EXISTS); snapshot debt is documented in _custom.json per the established pattern.
- Tests are real-schema (PGlite), assert both rejection AND row survival on DELETE, and the pre-existing severity test was correctly adapted to the append-only world instead of deleted.

### Advisories (non-blocking, consistent with the ADR staging)

1. The trigger binds every role including the owner, but the owner can still DROP TRIGGER first. Against a determined owner-role attacker the durable answer is the scoped non-owner app role plus Stage 3/4 signing and anchoring; worth stating in the Stage 3 spec so nobody mistakes the trigger for tamper-proofing. It is tamper-EVIDENCE staging, and honest about it.
2. TRUNCATE stays open by documented choice (BEFORE DELETE does not fire on TRUNCATE; harnesses reset with it). Note that plain TRUNCATE does not reset the sequence, so even a truncate leaves a detectable seq discontinuity once anchoring lands. Fine as-is.
3. The trigger makes retroactive signature backfill impossible by construction. That matches the Stage-1 rows-are-honestly-unsigned posture; flagging so Stage 3 plans sign-at-insert only and nobody attempts a backfill later.
4. Pre-existing rows receive seq in backfill scan order, not timestamp order (documented in the migration). Anchoring should treat pre-0026 order as untrusted.

PR #1972, opened 2026-07-18
feat(server): license refresh endpoint + read-only exemption (GAP-287 PR-1)

RevealUIStudio:
## Scope — GAP-287 PR-1 (the refresh endpoint, first in the sequence)

The machine path a shortened license-JWT TTL requires, shipped FIRST so that no
TTL reduction can brick a valid self-hosted customer at expiry. This PR is
exactly the §6 "PR-1" row of the design spec: the §4 refresh endpoint, its
`READ_ONLY_EXEMPT` entry, and admin-page + day-7 email discoverability. It ships
while all extant keys are still long-lived, so customer risk is zero. It does
NOT touch the exp derivation (PR-2) or the manual-mint default (PR-3).

### What §4 defines that this implements

`POST /api/license/refresh` returns the customer's CURRENT stored license key.
It **never mints**, so it cannot extend entitlement beyond what billing granted;
it only delivers what the webhook lifecycle already wrote. Bindings, in order:

1. **Signature valid within a bounded expiry window.** New
   `validateLicenseKeyForRefresh` (in `@revealui/core/license`) verifies the
   presented key against the ordered public-key set (`getPublicKeys()`, current +
   `_NEXT`), reusing the same rotation-aware multi-key path as
   `validateLicenseKey` (GAP-259 composition), but with a **wider** clock
   tolerance: a key whose `exp` is past by at most `REFRESH_ACCEPT_DAYS`
   (ratified 30d) still refreshes. Beyond that window the verifier returns null.
2. **Active license row for the token's `customerId`.** Row-level revocation is
   enforced here: a revoked / expired / lapsed license has `status != 'active'`,
   so no key is returned. Scoped to the deployment's Stripe mode (mirrors
   `getUserLicenseKey`).
3. **Uniform denial.** Every failure (bad signature, expired-beyond-window, no
   active row, unconfigured key, DB error) returns the identical
   `403 { "error": "refresh_denied" }` with no oracle distinguishing revoked /
   missing / lapsed. Fails closed on DB error.

`READ_ONLY_EXEMPT` gains `/api/license/refresh` (+ `/api/v1/...`): a read-only
(lapsed-support) customer fetching their current key is exactly the escape-hatch
class that set exists for. Discoverability: the admin license page shows the
one-line command, and the day-7 lifecycle email (paid tiers) points to it.

### Deferred to GAP-260 (documented, not a gap in this PR)

Per-`jti` denylist revocation — refusing a specifically revoked token lineage
**even when its customer's row is still active** (the "a stolen old key extends
the attacker's window" mitigation) — requires the GAP-260 jti-denylist, which is
owner-gated and not yet built (no denylist store exists in the schema). Until it
lands, row-level revocation (binding 2 above) is the enforced revocation path.
The pure per-jti behavior is captured as a skipped, documented-RED test.

## Red-first evidence (per spec §7)

Each proof: revert the source to `origin/test`, run, capture RED, restore from
`HEAD`, confirm GREEN.

| Proof | Reverted source | Result (RED) |
|---|---|---|
| A. Core refresh validator | `packages/core/src/license.ts` | 8 tests fail — `REFRESH_ACCEPT_DAYS` undefined, `validateLicenseKeyForRefresh is not a function` |
| B. Refresh endpoint | `apps/server/src/routes/license.ts` | 9 tests fail — endpoint absent, `expected 404 to be 200/403` |
| C. Read-only exemption | `apps/server/src/middleware/license.ts` | 1 test fails — `/api/license/refresh must be exempt: expected 403 to be 200` |
| D. Email discoverability | `apps/server/src/lib/lifecycle-emails.ts` | 6 paid-tier tests fail — email missing "refresh endpoint" (free-tier + em-dash checks stay green) |
| E. GAP-260 documented-red | (un-skipped against THIS branch) | 1 test fails — `expected 200 to be 403`: an active-row customer's revoked-lineage key still refreshes today (no denylist). Kept `it.skip` so CI is green. |

All reverts restored; working tree clean after each proof.

## Test counts (all green)

- `packages/core` `license.test.ts`: **87 passed** (8 new refresh tests: window accept/reject, explicit-window override, rotation composition, unrelated-key reject, empty-candidate).
- `apps/server` refresh + license + support-expiry + lifecycle-emails: **66 passed, 1 skipped** (the GAP-260 documented-red).
- Typecheck: `@revealui/core`, `server`, `admin` all clean. Biome clean on all touched files.

## Guardrail-2 — crown-jewel-adjacent (signing surface)

This PR touches `routes/license.ts` and `middleware/license.ts`. The fleet
security self-check (`security-pr-review-check.js --diff origin/test`) correctly
returns **HOLD** for these paths. Per `dup-work-and-review-gates.md` this PR
requires a **recorded non-author verdict** (`sec-review:approved` label) before
merge. I am the builder and do **not** self-merge or self-clear.

## Sequencing warning (spec §6, load-bearing)

**PR-2 (exp derivation + re-mint cadence) must never land before this PR.**
Shortening the TTL before a refresh path exists would hard-fail license
validation for every current paying self-hosted subscriber at expiry, with no
machine path to a fresh key. This PR is the prerequisite that makes the TTL
reduction safe.

### DIFF
--- a/.changeset/gap-287-license-refresh-validator.md
+++ b/.changeset/gap-287-license-refresh-validator.md
@@ -1,0 +1,5 @@
+---
+'@revealui/core': minor
+---
+
+Add `validateLicenseKeyForRefresh` and the `REFRESH_ACCEPT_DAYS` constant to the license module. The refresh validator verifies a presented license JWT within a wider bounded-expiry window than the run-time validation grace, reusing the same rotation-aware multi-key verification path, so a running instance can obtain its current stored key without a new key being minted.
--- a/apps/admin/src/app/(frontend)/account/license/page.tsx
+++ b/apps/admin/src/app/(frontend)/account/license/page.tsx
@@ -168,6 +168,7 @@ export default function LicensePage() {
     }
   };
 
+  const apiBase = (process.env.NEXT_PUBLIC_API_URL || 'https://api.revealui.com').trim();
   const tier = subscription?.tier ?? 'free';
   const limits = TIER_LIMITS[tier];
   const tierFeatures = features?.[tier];
@@ -303,6 +304,28 @@ export default function LicensePage() {
                 <pre className="overflow-x-auto rounded-lg bg-zinc-100 p-2 text-xs font-mono dark:bg-zinc-900">
                   REVEALUI_LICENSE_KEY=your-key-here
                 </pre>
+              </div>
+            </div>
+            <div className="rounded-lg border p-3 dark:border-zinc-800">
+              <p className="text-xs font-medium text-zinc-600 uppercase tracking-wide mb-2">
+                Automatic key renewal
+              </p>
+              <div className="space-y-1.5 text-sm text-zinc-600 dark:text-zinc-400">
+                <p>
+                  Your subscription key is re-issued on your billing cycle. A running instance can
+                  pull its current key at any time from the refresh endpoint, so a renewed key
+                  reaches your deployment without a manual copy. Run this from the instance that
+                  holds the key:
+                </p>
+                <pre className="overflow-x-auto rounded-lg bg-zinc-100 p-2 text-xs font-mono dark:bg-zinc-900">
+                  {`curl -X POST ${apiBase}/api/license/refresh \\
+  -H 'content-type: application/json' \\
+  -d "{\\"licenseKey\\":\\"$REVEALUI_LICENSE_KEY\\"}"`}
+                </pre>
+                <p className="text-xs text-zinc-400 mt-2">
+                  The endpoint returns your current key and never issues a new one. If your key is
+                  older than the renewal window, copy the key above from this page instead.
+                </p>
               </div>
             </div>
             {publicKey && (
--- a/apps/server/src/lib/__tests__/lifecycle-emails.test.ts
+++ b/apps/server/src/lib/__tests__/lifecycle-emails.test.ts
@@ -1,0 +1,44 @@
+/**
+ * Lifecycle email copy — GAP-287 PR-1 discoverability.
+ *
+ * The day-7 outcome email for PAID tiers must surface the renewal path (the
+ * refresh endpoint + the license page that carries the one-line command) so a
+ * shortened-TTL key never strands a running instance. The free-tier email must
+ * NOT carry it (no license key to refresh).
+ */
+
+import { describe, expect, it } from 'vitest';
+import { buildDay7Outcome } from '../lifecycle-emails.js';
+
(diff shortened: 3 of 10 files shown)


RevealUIStudio:

Adversarial review of the full diff (Fable session, non-author). License surface: crown-jewel-adjacent, reviewed accordingly.

### What holds up

- The refresh endpoint never mints (tested: generateLicenseKey uncalled on every path) and returns only what the webhook lifecycle already wrote, so entitlement cannot extend beyond billing.
- Oracle discipline is real: every denial (beyond-window, no active row, revoked row, no pubkey, DB throw) returns the byte-identical 403 body, and the test suite pins that property explicitly.
- Mode scoping (licenses.mode = configured Stripe mode) mirrors getUserLicenseKey and blocks test-mode keys from pulling live keys.
- The core refactor is right-shaped: one shared verifyAndParseLicenseJwt where the ONLY divergence is clock tolerance, so the GAP-259 rotation-aware multi-key loop lives in exactly one place, with the dual-key rotation window covered by a real minted-under-outgoing-key test.
- READ_ONLY_EXEMPT addition is the correct class (both /api and /api/v1 forms, test updated): a read-only-mode customer fetching their stored key is precisely the escape hatch that list exists for.
- The GAP-260 stolen-jti hole is documented at the exact code point, carried as a skipped RED test per the spec, and its risk is genuinely zero in PR-1 because all extant keys are long-lived (a stolen key already works directly today; refresh adds nothing until TTLs shorten).

### Advisories (non-blocking)

1. SEQUENCING BINDING, the one that matters: PR-2 (TTL shortening) must NOT land before the GAP-260 per-jti denylist unless the owner explicitly accepts that a stolen old key within the 30-day window can pull the fresh key of a still-active customer. PR-1 is safe on its own; the ordering is what keeps it safe. Record this on the GAP-287 sequence.
2. The active-row query is customerId+status+mode, newest first, with no tier binding to the presented token. A customer holding two active licenses (e.g. subscription plus perpetual) gets whichever row is newest, which may not be the lineage they presented. Consistent with getUserLicenseKey today; consider a tier or lineage filter in PR-2.
3. Confirm the route inherits the server's rate limiting: it is unauthenticated and does signature verification plus a DB roundtrip per call. Brute-forcing an Ed25519 JWT is infeasible, so this is load hygiene, not a break.
4. Denial paths differ in work performed (no-pubkey returns immediately; valid-sig-no-row does a DB roundtrip), so response timing can weakly distinguish states the body hides. Marginal signal, standard accepted risk; noting for completeness.

PR #1971, opened 2026-07-18
fix(pricing): derive perpetual renewal from catalog, sell perpetual in-app

RevealUIStudio:
Closes GAP-306 remaining scope (revealui-jv `docs/gaps/GAP-306.yml`, work item 2). Work items 1/3/4 were already shipped per the 2026-07-10 reconciliation (`revealui-jv docs/specs/2026-07-10-support-renewal-reconciliation-and-design.md`); this PR closes the last residual: renewal display sourcing + the marketing perpetual CTAs.

**Touches billing/pricing surfaces — needs a guardrail-2 recorded verdict (`sec-review:approved` label or a recorded PR verdict) before merge, per `.jv/.claude/rules/dup-work-and-review-gates.md`. Do not merge without it.**

## What changed

1. **`packages/contracts/src/pricing.ts`** — populated `PERPETUAL_TIERS[*].renewal` (was declared but never set) with the literal cents-of-record from `scripts/setup/stripe-catalog.ts` `CATALOG` (`revealui_renewal_{pro,max,enterprise}`: 14900/79900/399900 cents = $149/$799/$3,999 per year). Previously the marketing fallback file (`apps/marketing/app/lib/pricing-fallbacks.ts`) was the only place this string lived; now contracts (the shared source admin/marketing/server all consume) carries it too.
2. **`scripts/validate/pricing-lockstep.ts`** — extended the existing cents-of-record lockstep gate to also check `PERPETUAL_TIERS[*].renewal` and `PERPETUAL_PRICE_FALLBACKS[*].renewal` against the catalog's renewal prices (previously only `.price` was checked). Verified this actually catches drift: temporarily broke `$149` -> `$150` locally, confirmed the gate fails with `exit 1` and a clear diagnostic, then reverted.
3. **`packages/contracts/src/pricing.ts`** — replaced the Agency Perpetual and Enterprise Perpetual `mailto:support@revealui.com` CTAs with `/account/license`, matching the already-live Pro Perpetual pattern (`POST /api/billing/checkout-perpetual` already accepts `tier: 'pro' | 'max' | 'enterprise'` server-side -- `apps/server/src/routes/billing.ts:1673-1799`). CTA labels updated to `Buy Agency Perpetual` / `Buy Enterprise Perpetual` to match.
[...]

PR #1975, opened 2026-07-18
feat(harnesses): fail-closed gateway authn with challenge-response pairing + durable tokens (GAP-353 PR-2)

RevealUIStudio:
## Headline: a never-paired daemon now REFUSES unauthenticated `agent.spawn` / `agent.stop`

The one operator-grade spawn/kill primitive in the codebase was sitting behind an open door. `checkAuth` returned `true` whenever no token had been issued yet, so on a fresh daemon every unauthenticated `/rpc` and `/api/*` call was granted, including `agent.spawn` (spawns a child process) and `agent.stop` (`child.kill`). This PR deletes that fail-open branch and makes the gateway fail-closed.

**Red/green proof (headline first).** Against `origin/test`, a behavioral proof asserting the never-paired daemon refuses unauthenticated `agent.spawn` and `agent.stop` fails, because the fail-open branch grants both:

```
FAIL  refuses unauthenticated agent.spawn  -  expected 200 to be 401 (dispatch WAS called)
FAIL  refuses unauthenticated agent.stop   -  expected 200 to be 401 (dispatch WAS called)
```

With this PR both return `401` and the RPC dispatch is never reached. The gap `verify` line is satisfied:

```
$ grep -n "no tokens have been issued" packages/harnesses/src/server/http-gateway.ts
(no output  -  the fail-open comment and branch are gone)
```

## Scope (spec §4 PR-2, built to §6 rulings + §7 findings)

- **§D1 fail-closed bootstrap.** A 32-byte secret is written to `<data-dir>/pairing-secret` at mode `0600`, created only if absent; its SHA-256 hash is persisted via the PR-1 `gateway_bootstrap` store. The `checkAuth` fail-open branch is deleted entirely  -  unauthenticated `/rpc` and every `/api/*` route 401 unless a valid bearer token is presented.
- **§6 R1 challenge-response pairing (supersedes D1's POST-the-secret flow).** `GET /api/pair` issues a single-use, short-lived nonce; the client returns `HMAC-SHA256(secret, nonce)`; the server verifies with `timingSafeEqual`. The plaintext secret never crosses the wire. A consumed or expired nonce cannot be replayed (the nonce is deleted on the attempt, success or failure).
- **§D2 durable tokens.** On a successful pair the gateway mints a 32-byte token, stores only `sha256(token)` in `gateway_tokens` with `issued_at` + `expires_at` (default 90-day TTL per ruling C-B, configurable) + optional label. `checkAuth` hashes the presented token and matches the durable store with a constant-time re-check. `paired` state derives from "a valid token exists," never from a volatile boolean  -  a restart with valid tokens stays authenticated.
- **§D5 rate limit + lockout.** The pairing endpoint carries a per-source failed-attempt counter with exponential backoff, a lockout ceiling, and a coarse global cooldown, with loud logging (paths + events only, never secret values). In-memory counters per spec.
- **§D7 boot honesty.** Missing secret file + existing DB hash refuses new pairings (existing tokens still work); secret file present but hash-mismatched refuses (possible substitution); group/other-readable mode refuses to use it. All logged loudly.
- **§7 R3 pre-auth allowlist.** `PRE_AUTH_ROUTES` is an explicit, exported, tested constant (exactly `GET /api/pair` and `POST /api/pair`); everything else under `/rpc` and `/api/*` 401s unauthenticated.
- **§7 R5 fresh-DB re-hash.** Fresh store + existing `0600` file on start re-hashes the file into the new store (the inverse of D7's file-missing case). Tested.

### The HMAC keying / verification contract (design decision within the spec envelope)

The secret file holds a 64-char hex string (32 random bytes). The operator/Studio reads it verbatim; that exact string is the HMAC key. The server loads the same string into memory (`bootstrapSecret`) and recomputes `HMAC-SHA256(fileSecret, nonce)` to verify. The DB `gateway_bootstrap` hash is **not** the HMAC key (a one-way SHA-256 cannot reproduce the HMAC)  -  it exists solely as an integrity check to detect file tampering/substitution (§D7). So the server must hold the file secret to verify; the DB hash guards the file.

### Nonce lifecycle

Nonces are 32 random bytes (hex), held in an in-memory `Map<nonce, expiry>` with a 2-minute TTL and a 256-entry ceiling (a DoS bound on the pre-auth GET). A nonce is consumed (deleted) the moment a pairing attempt looks it up, so it is single-use whether the attempt succeeds or fails; an expired or unknown nonce is a failure.

### Retired 6-digit code

Ruling C-A is folded in here because leaving it would have been dead, misleading code: the only in-repo consumer was one CLI `serve` line, updated to print the bootstrap secret **path** (never its value). The brute-forceable 6-digit code, its generator, and the old `{code}` POST flow are removed. **PR-3 is now docs + the Studio pairing-flow update only.**

## Red/green evidence

- **Red (vs `origin/test`):** the never-paired-refuses-spawn and never-paired-refuses-stop assertions both fail against base (200 + dispatch called); the verify grep still finds the fail-open comment.
- **Green (this PR):** both return 401 with dispatch unreached; the verify grep returns nothing. Full challenge-response happy path, wrong-HMAC 403, consumed-nonce 403, expired-nonce 403, per-source lockout 429 then recovery after the window, restart-durability (old token still authenticates on a new gateway instance sharing the store; unauth still 401s), §7 R5 fresh-DB re-hash, and all three §D7 refusals all pass.

## Test counts

- **New:** `packages/harnesses/src/server/__tests__/http-gateway.test.ts`  -  19 tests (16 HTTP behavior + 3 `PairingRateLimiter` unit tests for deterministic backoff / global-ceiling with an injected clock).
- **Package suite:** `@revealui/harnesses` 470 tests across 33 files, all green. `tsc --noEmit` clean. Biome clean on changed files.

## Guardrail-2 notice

`packages/harnesses/src/server/` is a guardrail-2 spawn/kill surface. `security-pr-review-check.js --diff origin/test` flags this branch security-sensitive. This PR needs a **recorded non-author verdict** (`sec-review:approved` label) before merge. The author does not self-merge and does not apply gate-clearing labels.


RevealUIStudio:

Non-author review (fable coordinator; Opus adversarial pass). The fail-closed auth core is strong and the headline bypass fix is correct and well-tested: no fail-open path in checkAuth, nonce consumed before HMAC verify (replay-proof), timingSafeEqual with length guard, only token hashes persisted. But two HIGH findings block, both CodeQL-flagged (js/file-system-race, CI currently red) and both on the secret file that gates spawn/kill:

1. http-gateway.ts:234-244 — TOCTOU on the secret READ: existsSync → statSync mode check → readFileSync. A symlink/file swap between stat and read serves an attacker-controlled HMAC key past the permission gate. Fix: open an fd with O_NOFOLLOW, fstat the fd for the mode check, read from the fd.

2. http-gateway.ts:276-278 — TOCTOU on the secret CREATE: writeFileSync after an earlier existsSync check, no exclusive create. A planted symlink makes the daemon write the fresh 32-byte secret into an attacker-readable target, and mode 0600 only applies on creation. Fix: flag 'wx' (O_EXCL) + O_NOFOLLOW.

These are in-scope by the PR's own §D7 threat model (local file substitution and permissive modes). Non-blocking notes: 0.0.0.0 bind + ACAO:* is acceptable now that auth is fail-closed and /api/pair is rate-limited, but deserves an explicit note; handleStatic has its own low-impact TOCTOU (public shell only).

Re-request review after the two fixes land and CodeQL is green.

RevealUIStudio:
## Guardrail-2 security verdict: REQUEST-CHANGES (re-review of head ff62c92)

Re-review at owner request. The new push is a rebase onto the merged #1971/#1972; the initAuth secret-file logic is byte-identical to the previously reviewed version. Both blockers remain, verified against the current head, not a stale run:

- http-gateway.ts:235-244 read-path TOCTOU still present (existsSync → statSync mode check → readFileSync, all path-based). Fix: open ONE fd with O_NOFOLLOW, fstat that fd for the mode check, read from that same fd.
- http-gateway.ts:276-278 create-path TOCTOU still present (writeFileSync without exclusive create). Fix: flag 'wx' (O_EXCL) + O_NOFOLLOW.

CodeQL's failing run is bound to head ff62c92 and reports the same 2 HIGH js/file-system-race alerts at :244 and :278. All other CI is green; the fix simply is not in the diff. New fd-based tests for the race path are recommended alongside the fix. Re-request after the two patterns land and CodeQL is green on the new head.

RevealUIStudio:
## Guardrail-2 security verdict: APPROVE (head a4b56b7b)

Final non-author verification (fable coordinator; Opus adversarial pass). Both previously blocking HIGH TOCTOU races are correctly and completely closed:

- Read path (:247-334): one fd opened with O_RDONLY|O_NOFOLLOW; fstatSync(fd) drives BOTH the isFile() and 0o077 mode checks; readFileSync(fd) reads that same fd; closeSync in finally. No residual path-based stat/read of secretPath remains in the auth path. ELOOP and other open errors refuse fail-closed; ENOENT falls to create.
- Create path (:281): flag 'wx' (O_CREAT|O_EXCL|O_WRONLY) with mode 0600 applied atomically; EEXIST adopts the winner via loop; chmodSync removed.

CodeQL analysis on this head: PASS, zero js/file-system-race alerts (previously 2 HIGH). The three new tests exercise the real race properties (real symlink refusal with target un-clobbered, real 0o640 fstat refusal, deterministic EEXIST-window injection asserting winner adoption). Harnesses suite 22/22, package 473/473, gate PASS.

Non-blocking follow-up: the for(;;) retry has no iteration cap; a local adversary with data-dir write access could livelock startup (DoS only, no secret compromise since O_EXCL + O_NOFOLLOW hold). Recommend a 3-5 attempt bound then fail-closed, as defense-in-depth in a later PR.

Label/merge condition: Unit Tests and the trailing CodeQL PR-summary context must finish green on this head.

PR #1978, opened 2026-07-18
feat(server): period-bind license exp and re-mint on renewal (GAP-287 PR-2)

RevealUIStudio:
## GAP-287 PR-2 — subscription license exp derivation + renewal re-mint cadence

Second PR of the GAP-287 license-JWT TTL program. Scope is exactly the §6 "PR-2" row of the design spec: derive the hosted subscription license JWT `exp` from the Stripe billing period at the three subscription mint sites, plus the re-mint cadence on renewal, plus the shared named constants. Manual-mint default and kit constant naming (§6 PR-3) are **not** in scope.

### Sequencing invariant holds

PR-1 (the `POST /api/license/refresh` endpoint, #1972) is **merged**, so the §6 precondition is satisfied: a customer holding a key that expires can fetch a fresh one. This PR ships the shortened, period-bound TTL only now that the refresh path exists (§2 brick-risk avoided). Extant 365d keys are untouched: no revocation, no backfill; they age out naturally.

### Countersigned numbers (§5, ratified by the owner 2026-07-18)

- Subscriptions: `exp = current_period_end + 7d` slack (`RENEWAL_SLACK_DAYS = 7`).
- Annual: same mechanism (no practical change).
- Perpetual (`webhooks.ts` perpetual mint): **exempt by design** — no exp, untouched.
- Canary (60s throwaway) and manual-mint: untouched (PR-3 owns the manual default).

### What changed

**`packages/core/src/license.ts`** — new pure derivation surface:
- `RENEWAL_SLACK_DAYS` / `RENEWAL_SLACK_SECONDS`, `DEFAULT_SUBSCRIPTION_TTL_SECONDS`.
- `subscriptionLicenseExpiresInSeconds(periodEnd)` — relative lifetime `(period_end + slack) − now`; falls back to the 365d default when a subscription has no billing period, so an abnormal Stripe shape never mints a shorter-than-before token.
- `subscriptionExpBound(periodEnd)` — the absolute epoch bound the renewal cadence compares against.
- `readLicenseExp(licenseKey)` — unverified exp read (internal, for the re-mint decision only; never an auth decision). `generateLicenseKey`'s default now references `DEFAULT_SUBSCRIPTION_TTL_SECONDS` (byte-identical).

**`apps/server/src/routes/webhooks.ts`** — the three subscription mint sites now derive exp from `current_period_end + slack`, and a new renewal cadence:
- Checkout issuance — `insert-subscription-license` saga step.
[...]

PR #1983, opened 2026-07-18
fix(harnesses): bound the initAuth secret-file retry loop

RevealUIStudio:
## Fast-follow: bound the initAuth secret-file retry loop

Non-blocking finding recorded on the guardrail-2 APPROVE verdict for merged PR #1975:

> the `for(;;)` retry has no iteration cap; a local adversary with data-dir write access could livelock startup (DoS only, no secret compromise since O_EXCL + O_NOFOLLOW hold). Recommend a 3-5 attempt bound then fail-closed, as defense-in-depth in a later PR.

### Change

`HttpGateway.initAuth()` (`packages/harnesses/src/server/http-gateway.ts`) now bounds the create-path retry loop at `MAX_INIT_AUTH_ATTEMPTS = 5`. The loop only ever re-iterates on an `EEXIST` from the exclusive (`wx`) secret-file create (another process winning the create race). After 5 attempts it fails closed: `bootstrapSecret = null`, a loud `logAuth` message naming the secret path and attempt count, and pairing then 503s exactly as it already does for every other fail-closed path in this function. No change to the read-path or create-path TOCTOU protections (`O_NOFOLLOW` + `O_EXCL`) landed in #1975/#1978 — those still hold; this is defense-in-depth against a livelock, not a new secret-substitution vector.

### Red -> green evidence

New test `bounds the create-path retry loop so a persistent EEXIST race cannot livelock startup (#1975 verdict fast-follow)` in `packages/harnesses/src/server/__tests__/http-gateway.test.ts`, driven by a `writeFileSync` mock that forces every exclusive create attempt to throw `EEXIST` while never actually writing a winner file (so the read-back also always `ENOENT`s) — a persistent create/delete race.

- **RED** (`git checkout origin/test -- packages/harnesses/src/server/http-gateway.ts`, base code, unbounded loop): the loop never gave up on its own — it only stopped when the test's own 200-attempt safety valve released the forced `EEXIST`. `expect(attempts).toBeLessThan(200)` failed with `attempts === 200`.
- **GREEN** (fix restored): the loop gives up after exactly 5 attempts, well under the safety valve, `initAuth()` returns with the secret unset, and a subsequent pairing attempt 503s (fail-closed).

### Test counts

- `http-gateway.test.ts`: 23/23 passing (was 22, +1 new).
- `@revealui/harnesses` full suite: 473/474 passing. The one failure (`opencode-adapter.test.ts > get-status reports availability`) is pre-existing and environment-dependent — it asserts `available: false` assuming the `opencode` CLI is not installed, but it is present in this dev environment (`which opencode` resolves). Unrelated to this change; the file is untouched by this diff.
- Typecheck (`tsc --noEmit`): clean.
- Biome: clean on touched files.

### Guardrail-2 notice

This PR touches `packages/harnesses/src/server/http-gateway.ts`, a guardrail-2 sensitive surface (auth/secret-file handling). Per the fleet security self-check output below, it requires a recorded coordinator verdict before merge — same gate #1975 and #1978 went through.

```
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: packages/harnesses/src/server/__tests__/http-gateway.test.ts, packages/harnesses/src/server/http-gateway.ts
```

### Changeset

`@revealui/harnesses` patch.

### DIFF
--- a/.changeset/gap-353-initauth-retry-cap.md
+++ b/.changeset/gap-353-initauth-retry-cap.md
@@ -1,0 +1,5 @@
+---
+'@revealui/harnesses': patch
+---
+
+Bound the `HttpGateway.initAuth` bootstrap-secret retry loop to 5 attempts before failing closed, so a persistent create/delete race on the secret file cannot livelock daemon startup. Fast-follow on a non-blocking finding from the #1975 guardrail-2 security verdict (DoS-only defense-in-depth; O_EXCL + O_NOFOLLOW already prevent secret substitution).
--- a/packages/harnesses/src/server/__tests__/http-gateway.test.ts
+++ b/packages/harnesses/src/server/__tests__/http-gateway.test.ts
@@ -25,7 +25,17 @@ import type { RpcServer } from '../rpc-server.js';
 // call (which runs immediately before the exclusive 'wx' write) to deterministically
 // exercise the create-path EEXIST fallback. Every other test leaves it null, so
 // mkdirSync passes straight through to the real implementation.
-const fsHooks = vi.hoisted(() => ({ beforeMkdir: null as (() => void) | null }));
+//
+// `forceCreateEexist` + `onCreateAttempt` simulate a PERSISTENT create-path race
+// (GAP-353 initAuth retry-cap test): every exclusive ('wx') write attempt fails
+// with EEXIST and the winner's file is never actually written, so the read-back
+// that follows also ENOENTs. That is the shape a bounded-retry regression test
+// needs to prove the loop gives up on its own rather than spinning forever.
+const fsHooks = vi.hoisted(() => ({
+  beforeMkdir: null as (() => void) | null,
+  forceCreateEexist: false,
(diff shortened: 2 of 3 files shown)


RevealUIStudio:
**Author-session adversarial review (recorded; NOT a gate-clearing verdict — this session dispatched the build).** Read the full diff. The cap is correct: the loop only ever re-enters on EEXIST, 5 attempts is generous for a genuine one-shot create race, and exhaustion fails CLOSED (bootstrapSecret null, loud log, new pairings 503) while existing tokens keep authenticating — matching the D7 refuse-new-pairings semantics. The persistent-EEXIST mock is a faithful adversarial shape and the red proof (base spun to the 200-attempt test valve) is convincing. One non-blocking nit: the post-fix assertion `attempts < SAFETY_VALVE` could pin `attempts === MAX_INIT_AUTH_ATTEMPTS` for a tighter regression signal. No security findings.

PR #1984, opened 2026-07-18
fix(core,server): add 1s idempotency tolerance to renewal re-mint

RevealUIStudio:
## Fast-follow: 1s idempotency tolerance on the renewal re-mint decision

Non-blocking finding recorded on the guardrail-2 APPROVE verdict for merged PR #1978:

> idempotency off-by-one — the minted exp can floor 1s below subscriptionExpBound's value, so a duplicate invoice.payment_succeeded can re-enter the re-mint path instead of no-opping (bounded churn, not a money/entitlement error; the mocked tests don't cover the cross-function property). Suggested: compare with 1s tolerance (storedExp >= newBound - 1) or derive the skip bound with the same flooring as the mint.

### Change

- `packages/core/src/license.ts`: adds `RENEWAL_IDEMPOTENCY_TOLERANCE_SECONDS = 1` and `coversRenewalBound(storedExp, newBound)`, exactly the comparison the finding suggests (`storedExp >= newBound - 1`), documented against the flooring source (relative-TTL floor in `subscriptionLicenseExpiresInSeconds` plus the JWT signer's own second-granularity `exp`).
- `apps/server/src/routes/webhooks.ts`: `remintSubscriptionLicenseOnRenewal`'s idempotency check now calls `coversRenewalBound(storedExp, newBound)` instead of the bare `storedExp >= newBound`. This single function backs both call sites (the primary `invoice.payment_succeeded` renewal path and the payment-recovery re-mint), so both are covered.

No entitlement or tier logic changed. TTL-comparison only, matching the scope the original #1978 verdict approved.

### Red -> green evidence

Two levels, both proving the exact off-by-one:

**Pure boundary tests** (`packages/core/src/__tests__/license-exp-derivation.test.ts`, new `describe('coversRenewalBound (idempotency tolerance)')`): exercises `storedExp === newBound`, `storedExp === newBound - 1`, `storedExp === newBound - 2`, and `null`.
- RED (`git checkout origin/test -- packages/core/src/license.ts`): `coversRenewalBound is not a function` (the function doesn't exist yet on base) — 5/5 failing.
- GREEN (fix restored): 5/5 passing; `newBound - 1` correctly covers, `newBound - 2` correctly does not.

**Webhook integration test** (`apps/server/src/routes/__tests__/webhook-license-exp-derivation.test.ts`, new test `SKIPS re-mint when the stored key exp is exactly 1s below the new bound`): drives the real `invoice.payment_succeeded` handler with a stored license key whose `exp` is crafted (via an unverified-signature JWT-shaped string, since `readLicenseExp` never verifies the signature on this path) to land exactly `newBound - 1`.
- RED (`git checkout origin/test -- apps/server/src/routes/webhooks.ts packages/core/src/license.ts`, base comparison `storedExp >= newBound`): the handler re-minted — `expect(capturedUpdatedLicenseKey()).toBeUndefined()` failed, a JWT was captured.
- GREEN (fix restored): the handler correctly skips the re-mint — no update occurs.

### Test counts

- `license-exp-derivation.test.ts`: 16/16 passing (was 11, +5 new).
- `webhook-license-exp-derivation.test.ts`: 4/4 passing (was 3, +1 new).
- `@revealui/core` full suite: 2808/2808 passing.
- `server` app full suite: 3003/3003 passing (2 pre-existing skips, unrelated to this change).
- Typecheck (`tsc --noEmit`): clean on both `packages/core` and `apps/server` (required a `pnpm --filter @revealui/core build` first so the server's dist-resolved import of the new `coversRenewalBound` export type-checks).
- Biome: clean on all touched files.

### Guardrail-2 notice

This PR touches `apps/server/src/routes/webhooks.ts`, a guardrail-2 sensitive surface (billing/webhooks). Per the fleet security self-check output below, it requires a recorded coordinator verdict before merge — same gate #1975 and #1978 went through.

```
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: apps/server/src/routes/webhooks.ts
```

### Changeset

`@revealui/core` patch. `server` is changeset-ignored per repo convention; no separate server changeset needed.

### DIFF
--- a/.changeset/gap-287-remint-tolerance.md
+++ b/.changeset/gap-287-remint-tolerance.md
@@ -1,0 +1,5 @@
+---
+'@revealui/core': patch
+---
+
+Add a 1s idempotency tolerance (`coversRenewalBound`) to the GAP-287 PR-2 subscription renewal re-mint decision, so a stored license `exp` landing exactly 1s below the new period bound (a flooring artifact of the relative-TTL derivation and the JWT signer's own second granularity) is treated as already covering it. Fast-follow on a non-blocking finding from the #1978 guardrail-2 verdict: without this, a duplicate/retried `invoice.payment_succeeded` on that 1s boundary re-entered the re-mint path instead of no-opping (bounded churn, never an entitlement or money error).
--- a/apps/server/src/routes/__tests__/webhook-license-exp-derivation.test.ts
+++ b/apps/server/src/routes/__tests__/webhook-license-exp-derivation.test.ts
@@ -186,6 +186,21 @@ function decodeJwtClaims(jwt: string): Record<string, unknown> {
   return JSON.parse(Buffer.from(segment, 'base64url').toString('utf8'));
 }
 
+/**
+ * Builds an unverified JWT-shaped string carrying an exact `exp` claim, for
+ * pinning the renewal idempotency-boundary test to a precise second without
+ * racing the real signer's own `iat` capture. `readLicenseExp` (the only
+ * consumer of the stored key on the re-mint path) decodes without verifying
+ * the signature, so a well-formed but unsigned token is a faithful stand-in.
+ */
+function fakeLicenseKeyWithExp(exp: number, customerId: string): string {
+  const header = Buffer.from(JSON.stringify({ alg: 'EdDSA', typ: 'JWT' })).toString('base64url');
+  const payload = Buffer.from(JSON.stringify({ tier: 'pro', customerId, exp })).toString(
+    'base64url',
+  );
+  return `${header}.${payload}.unverified-test-signature`;
+}
+
 /** The `licenses` row the checkout saga inserted (the one carrying a licenseKey). */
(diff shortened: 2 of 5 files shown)


RevealUIStudio:
**Author-session adversarial review (recorded; NOT a gate-clearing verdict — this session dispatched the build).** Read the full diff. `coversRenewalBound` is exactly `storedExp >= newBound - 1` with null never covering; a 1s tolerance cannot mask a genuine renewal (a real renewal moves the bound by a billing period), so the only behavior change is the intended no-op on the flooring boundary. The unsigned-JWT test fixture is sound because `readLicenseExp` deliberately decodes without verifying. One invariant worth stating for the record: a null storedExp falls through to re-mint, which would be wrong for a perpetual key, but perpetual rows are unreachable on this path by the PR-2 row scoping (null subscriptionId never matches) — unchanged by this diff and already reviewed there. No security findings.

PR #1987, opened 2026-07-18
feat: sign audit rows at the write door with per-row Ed25519 (GAP-355 Stage 3 PR-2)

RevealUIStudio:
## Summary

Wires per-row **Ed25519** signing into the single `audit_log` write door, retires the legacy HMAC path, and lands the boot/env guards. Builds on the signing library (#1973) and the Stage-2 one-door + append-only migration.

Implements the Stage-3 signing design (internal). On a signing deployment every audit row lands with a `v1.ed25519.<kid>.<sig>` signature that verifies **offline** from the published public key; in dev/test (no key) rows stay honestly unsigned and the mode is logged at boot.

## What changed

**`@revealui/db` — signer slot at the door.** `DrizzleAuditStore` takes an optional injected `AuditRowSignerFn`. When present, `append`/`appendBatch` fetch each row's `seq` from the sequence up front (`SELECT nextval(pg_get_serial_sequence('audit_log','seq'))`, N in one round trip for a batch), sign over the full row including that `seq`, and `INSERT` an explicit `seq` + `signature`. This is required because `seq` is a signed field but a DB-assigned bigserial, and the append-only trigger from Stage 2 makes a sign-then-`UPDATE` impossible. Absent signer keeps today's behavior: DB-assigned `seq`, `NULL` signature. The package stays crypto-free and security-package-free — `AuditRowSignable` is a structural type; the real signer is composed in `apps/server`. `signature`/`previousSignature` pass-through is removed from `AuditEntry`; `previous_signature` is never written (the hash chain is abandoned; the column is kept).

> Note: sequence gaps from an aborted transaction are inherent to Postgres sequences — a Stage-4 anchoring concern, already recorded. Each row's signature over its own `seq` is unaffected.

**`apps/server` — sign at one construction site.** `audit-signer.ts` builds the Ed25519 signer from `REVEALUI_AUDIT_SIGNING_KEY` (+ a **kid**, see below) and exposes `createAuditStore(db)`, which **every** audit writer now uses (the security-event sink, the governed-MCP receipt stream, and the direct writers in agent-dispatcher / agent-stream / api-keys / a2a / webhooks / revmarket-executor). A signer that throws makes the append throw — no unsigned row lands on a signing deployment — routed through the existing write-result rails.

**MCP receipt migration.** Deleted `signChained` / `canonicalJson` / `lastMcpSignature` / `appendQueue` / `verifyMcpAuditChain`. The stream writes through the one door and the door signs; `previous_signature` is `NULL` going forward. The fail-closed mutating-tool policy and loud-failure rails are unchanged.

**Boot + env guards.** `validate-startup` now requires `REVEALUI_AUDIT_SIGNING_KEY` to parse as a valid Ed25519 PKCS#8 key at boot (hosted **and** self-hosted), refusing to serve otherwise — no `REVEALUI_SECRET` fallback. `assertAuditStorageEnv` adds it to the audit-critical set (a diverged-env deploy fails). `REVEALUI_AUDIT_HMAC_SECRET` is retired from the env schema/config and `docs/SECRETS.md`.

**Verifier tooling.** `scripts/security/verify-audit-signatures.ts` (`pnpm verify:audit-signatures`) reads the N most recent rows (or one by id) and prints a per-row verdict using `classifyAuditSignature` + `verifyAuditRow`. Minimal by design — it seeds the Stage-4 offline CLI.

## kid choice

The signature embeds a self-asserted `kid`. It reads `REVEALUI_AUDIT_SIGNING_KID` if set; otherwise it derives a stable, key-bound default `ed25519-<first 16 hex of sha256(SPKI DER)>`. Deterministic for a given key (so verification resolves it consistently) and it changes when the key rotates; contains no `.`, so it is safe inside the dotted wire format.

## Tests (red-first)

Integration test against real PGlite with real migrations: write a signed row through the door, read it back through jsonb + timestamptz, verify offline with `verifyAuditRow`.

**Proven RED first** by pointing the round-trip signer at a non-canonical builder (`JSON.stringify` instead of the RFC 8785 canonicalizer):

```
FAIL  audit-signing-roundtrip.pglite.test.ts > a canonically-signed row verifies OFFLINE after the jsonb + timestamptz round trip
AssertionError: expected false to be true
- Expected  true
+ Received  false
  expect(result.valid).toBe(true);
```

That is exactly the jsonb-key-order surface the whole effort exists to kill: a non-canonical signature cannot be reproduced from storage. After restoring the canonical signer the test is green. The suite also covers: signer-throw → append throws and no row lands; tamper any signed column (including `seq`) → verify fails; unsigned mode → `NULL` signature; `appendBatch` signs every row with its own fetched `seq`. A permanent inverse test keeps the non-canonical case as a regression guard.

### Commands run
- `turbo run typecheck` for `@revealui/db`, `@revealui/security`, `@revealui/config`, `server` — all green.
- Vitest: audit-signing-roundtrip, mcp-audit-concurrency, audit-storage, validate-startup (113), config modules (36), security audit suites (57), prod-required-coverage, secret-paths-lockstep, mcp-endpoint, webhook-pglite — all green.
- `biome check` — clean. Pre-push gate — PASS.

## Changesets
Minor bumps for `@revealui/db` (signer slot), `@revealui/security` (removal of the exported-but-unwired HMAC helpers, pre-1.0 minor), and `@revealui/config` (env-var swap).

## Notes for review
- This touches signing / security surfaces, so it needs a recorded non-author security verdict before merge.
- The `revealui/*/audit-hmac-secret` revvault paths are marked retired in `docs/SECRETS.md`; the physical path removal + the signing-key provisioning entry land together in the follow-up PR (revvault + setup-time generation + public-key endpoint).
- The `apps/admin` one-time setup-route audit write still constructs an unsigned store (separate boot context, no signing-key guard) — flagged as a follow-up so it is not silently left unsigned on a signing deployment.


RevealUIStudio:

Adversarial review by the designing (Fable) session against the Stage-3 signing design §D5–D8 and the Stage-2 substrate. Reviewer topology: dispatcher-session non-builder verdict (no peer session live; same convention as the 2026-07-18 precedent), builder was an independently-driven Opus agent.

**Verified (not just read):**
- The seq-before-INSERT shape is correct and forced: `seq` is a signed field, the bigserial is DB-assigned, and migration 0026's trigger forbids sign-then-UPDATE. `nextSeqValues` via `pg_get_serial_sequence` + `generate_series` is one round trip, batch-safe, and each row signs over its own fetched seq. Sequence gaps on aborted INSERTs are inherent and already recorded as a Stage-4 concern.
- Dependency direction holds: `packages/db` gains only a structural `AuditRowSignable`/`AuditRowSignerFn`; the crypto composes in `apps/server/src/lib/audit-signer.ts`. No security-package import in db.
- Fail-closed holds end to end: signer throw → append throws → no row lands (test proves zero rows), and the MCP mutating-tool policy is unchanged.
- D6 modes are honest: key presence enforced only in the production required-env sets, format check whenever present, unsigned dev mode logged loudly at boot. No fallback secret — correct; a signing key with a fallback is not a signing key.
- Chain removal (D7) is total in apps/server: `signChained`/`canonicalJson`/chain-head/mutex/`verifyMcpAuditChain` gone; with per-row signatures there is no read-modify-write left to serialize, so deleting the mutex is sound, not just convenient.
- Red-first proof recorded (non-canonical signer fails after jsonb readback — the exact finding-2 surface); tamper coverage includes seq renumbering.

**Binding (owner, sequencing — record before promotion):** this PR makes `REVEALUI_AUDIT_SIGNING_KEY` REQUIRED at boot on both production modes and adds it to the env-parity deploy gate. The key does not exist in prod env yet (PR-3 owns provisioning). Promoting this to `main` before the key is provisioned in Vercel/Fly env will fail the deploy / refuse boot — by design. Sequence: provision the key (PR-3 or manual env set) BEFORE the next test→main promotion that carries this.

**Advisory A (blocks the PR-3 claims re-arm, not this merge):** two unsigned writers remain outside apps/server — `apps/admin/src/app/api/setup/route.ts:104` (flagged in the PR body) and `scripts/admin/bootstrap.ts:166` (NOT flagged in the build report; the same file the Stage-2 single-door enforcer originally caught). Both construct the store with no signer, so on a signing deployment their rows land NULL and the D8 verifier will classify legitimate setup/bootstrap rows as integrity failures. Before the scoped claim ships in PR-3, these must either sign or be formally classified — otherwise "every action in the audit log is signed" is false on day one.

**Advisory B:** the operator verify script resolves every kid to the single supplied public key. Fine as the Stage-4 seed for single-key deployments; the Stage-4 CLI needs a kid→key map for rotation (GAP-261 `_NEXT`/`_PREVIOUS`).

**Advisory C (builder-noted, concur):** `packages/security/src/audit-write-failures.ts:86` `missing_secret` branch is now producer-less legacy; harmless, sweep later.

Merge and the `sec-review:approved` label remain owner dispositions.

PR #1989, opened 2026-07-18
feat(presentation): print entrance for the receipt motif (Phase 5 PR-R1)

RevealUIStudio:
## What

Adds an optional `animate?: 'print'` prop to `ReceiptCard` (default `undefined`, current behavior, zero visual change when off). When set:

- Each rendered `AuditLine` row and the integrity footer get a CSS entrance: opacity 0→1, translateY 6px→0, `var(--rvui-duration-slow)`, `var(--rvui-ease)`, fill-mode forwards, `animation-delay = 400ms + index * 420ms` (integrity footer is the last index).
- The integrity footer additionally gets a one-shot brand-glow box-shadow pulse (350ms) after its entrance.
- Pure CSS: keyframes + per-row delay via a CSS custom property (`--rvui-print-i`, set inline per row/footer). No JS timeline, no `IntersectionObserver`.
- Disabled entirely under `prefers-reduced-motion: reduce` (media query kills the keyframes; resting state is the completed receipt).

New file: `packages/presentation/src/animations/core/print-entrance.ts` (keyframes + class names + timing constants), exported from `animations/index.ts` alongside the existing `core/` exports.

## Why

Ships the receipt-motif hero's entrance animation as an additive, reusable `@revealui/presentation` capability, per the Phase 5 receipt-hero concept spec (build order: PR-R1 presentation, additive; PR-R2 marketing integration follows, stacked on this branch).

## Verification

- `pnpm --filter @revealui/presentation test` — 988/988 passing, including 2 new tests on `ReceiptCard`: prop off renders no animation classes/vars/style tag; prop on stamps the per-row `--rvui-print-i` delay custom property on each row and the footer, and all ledger content (including the integrity footer) is present in the DOM at first render with no waiting on the animation.
- `pnpm --filter @revealui/presentation typecheck` — clean.
- `pnpm --filter @revealui/presentation build` — clean (Vite lib build + `.d.ts` generation).
- Biome — clean on all touched files.
- **Goldens unchanged by design.** No visual-harness golden currently exists for `ReceiptCard` in this checkout, and the change is additive-only: the prop is `undefined` unless a consumer opts in, so the resting/default render path is byte-for-byte the same JSX as before this PR (verified by reading the diff — no unconditional change to any existing element's className/style).

## Changeset

Minor bump on `@revealui/presentation` (new additive prop + export).

### DIFF
--- a/.changeset/receipt-hero-print-entrance.md
+++ b/.changeset/receipt-hero-print-entrance.md
@@ -1,0 +1,5 @@
+---
+'@revealui/presentation': minor
+---
+
+Add an optional `animate="print"` prop to `ReceiptCard`. When set, each `AuditLine` row and the integrity footer play a one-shot CSS entrance stagger (the receipt "prints" itself line by line), and the integrity footer's seal pulses once after its entrance. Pure CSS, no JS timeline. Disabled under `prefers-reduced-motion: reduce`. Default is unchanged (no `animate` prop = no visual change).
--- a/packages/presentation/src/__tests__/components/receipt-card.test.tsx
+++ b/packages/presentation/src/__tests__/components/receipt-card.test.tsx
@@ -55,4 +55,51 @@ describe('ReceiptCard', () => {
     render(<ReceiptCard title="No seal" lines={lines} />);
     expect(screen.queryByText('sha256')).toBeNull();
   });
+
+  it('renders no animation classes, vars, or style tag when animate is unset', () => {
+    const { container } = render(
+      <ReceiptCard
+        title="Static"
+        lines={lines}
+        integrity={{ kind: 'sha256', value: 'a1b2c3d4' }}
+      />,
+    );
+    const items = screen.getAllByRole('listitem');
+    for (const item of items) {
+      expect(item.className).toBe('');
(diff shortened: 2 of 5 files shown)


RevealUIStudio:
## Fable design verdict: APPROVE (Phase 5 PR-R1)

Reviewed against the Phase 5 receipt-hero spec. The implementation matches the locked motion contract:

- Timings and curves are the spec's numbers via tokens: 400ms lead-in, 420ms stagger, `--rvui-duration-slow` entrance with `--rvui-ease`, one-shot 350ms seal pulse delayed until after the seal's entrance.
- Pure CSS with `both` fill, so the pre-delay state is held without JS and the DOM is complete from first paint. No timeline, no observers, no loops.
- `prefers-reduced-motion: reduce` collapses to the completed receipt (animation none, opacity 1), and the media query wins by cascade order. Verified in the source.
- Default-off is provably unchanged (test asserts zero classes, vars, and style tag), so existing goldens stand un-rebaselined, as required.
- Keyframes live in the animations layer with exported constants; extends the existing component, no new component.

Fleet checker: no security-sensitive files touched, no coordinator gate. Merging when CI is green (stack order: this PR first).

PR #1988, opened 2026-07-18
chore(admin): unregister tableless collections, mark orphans (F7)

RevealUIStudio:
## Summary

Foundation repair F7 (table/orphan hygiene). Two problems from the internal 2026-07-02 CMS content-surfaces audit:

- **B4 (tableless collections):** collections registered in the admin dashboard whose slug has no backing Postgres table. The storage layer resolves a collection to storage by slug in both tiers: the typed-storage bridge (`apps/admin/src/lib/db/typedCollectionStorage.ts`) keys handlers by `collection.slug`, and the dynamic SQL adapter (`packages/core/src/collections/operations/sqlAdapter.ts`) issues `... FROM "<slug>"` where the table name **is** the slug. A registered collection with no matching table renders in the dashboard but throws on every read and write.
- **B9 (orphans):** collection configs, a global, and blocks that exist on disk but are never registered or consumed.

This PR **unregisters** the tableless collections (config files kept on disk), adds a **CI guard** so this cannot silently regress, and marks every orphan with a `WIRE-UP-PENDING` header so the register-vs-delete call is made per item in review.

Code is the source of truth: the audit's stale B4 list was re-derived against the current schema/migrations. Cards/Heros/Layouts/Banners from the stale list are already gone (retired from the registry by the F1 pages-model refit), so they are not in scope here.

## Derived tableless list (B4) — evidence

Registered collections are enumerated in `apps/admin/src/lib/collections/registry.ts`. The real table set was derived from the Drizzle schema (`packages/db/src/schema/`) and migrations (`packages/db/migrations/0000`–`0026`). No collection declares a `dbName`/`tableName` override, so slug == table name.

| Collection | Slug | Backing table? | Evidence |
|---|---|---|---|
| Contents | `contents` | **No** | No `contents` pgTable in schema; no `CREATE TABLE contents` in any migration |
| Categories | `categories` | **No** | No `categories` pgTable; not in migrations |
| Tags | `tags` | **No** | No `tags` pgTable; not in migrations |
| Events | `events` | **No** | No `events` pgTable; not in migrations |
| Prices | `prices` | **No** | No `prices` pgTable; schema has `billing_catalog` but no `prices` table |
| Subscriptions | `subscriptions` | **No** | No `subscriptions` pgTable; schema has `account_subscriptions` but no bare `subscriptions` table |

[...]

PR #1991, opened 2026-07-18
feat(server): edit-session engine (schema, migration, API)

RevealUIStudio:
## Summary

Adds the P1 slice-1 **edit-session engine**: a durable, auditable draft workspace over live content docs. Editors open a session, patch draft overlays (autosave), then publish atomically or discard. Every action appends to an append-only event log.

This is the engine layer only (Drizzle schema + migration + Hono session API). No UI.

### Data model (migration `0027_open_storm`)

- `edit_sessions` — the workspace (site, title, status `open|published|discarded`, timestamps).
- `edit_session_docs` — one draft overlay per touched doc, materialized from the live doc on first patch, with `base_version` for optimistic locking. Unique on `(session_id, doc_type, doc_id)`.
- `edit_session_events` — append-only log with a `serial` id so `?after=<id>` polling is monotonic and stable.

The regenerated snapshot also heals a pre-existing drift: migrations 0021-0026 were hand-written and never updated the drizzle snapshot chain (snapshots jumped 0020 → 0027). The new `0027` snapshot captures true full state, so future `db:generate` runs diff cleanly. The `0027` SQL itself contains only the edit-session delta (idempotent DDL per the repo's migration discipline).

## Endpoints (mounted under `/api/content`)

| Method | Path | Purpose |
|---|---|---|
| POST | `/sessions` | Open a session; writes `opened` event |
| GET | `/sessions` | List (filter by `status` / `siteId`) |
| GET | `/sessions/:id` | Session + its docs + recent events |
| PATCH | `/sessions/:id/docs/:docType/:docId` | Field patch (autosave); materializes the draft from the live doc on first patch, records `base_version`; writes `patched` event |
| GET | `/sessions/:id/events?after=<cursor>` | Poll events after an id-ordered cursor |
| POST | `/sessions/:id/publish` | Optimistic, all-or-nothing publish; snapshots into `page_revisions`, prunes to 50/doc; writes `published` event |
| POST | `/sessions/:id/discard` | Discard; overlay rows retained for audit; writes `discarded` event |

Discard is a **POST** (`/sessions/:id/discard`), not a DELETE. The `/api/content/*` mount maps the DELETE verb to the content `delete` permission (admin only), which is wrong for abandoning your own draft. As a POST it rides the content `update` permission like open/patch/publish.

Mutations on a non-`open` session return 409. The draft patch setter walks dot/array path segments with a prototype-pollution guard (`__proto__` / `constructor` / `prototype` rejected) and no authored regex.

### Deliberate Phase-1 slice: `page` only

`doc_type` `page` is fully implemented. `global` and `post` are accepted by the schema but PATCH and publish reject them with a clear **400 "supported in a later slice"**. This is intentional for slice 1.

### Publish is optimistic + compensated, not a DB transaction (by necessity)

The production driver is Neon HTTP, which has no interactive transactions (`drizzle-orm/neon-http` throws `No transactions support`). Publish is therefore:

1. **Pre-flight** — read and version-check every doc. Any mismatch → 409 with per-doc detail and **zero writes**. This catches the common case (an external edit landed after the draft was based) before anything is written.
2. **Guarded writes** — each page update is version-guarded (`UPDATE ... WHERE version = base_version`), snapshots a revision, and prunes to 50/doc.
3. **Compensation on a mid-flight miss** — if a guarded write misses (a writer won the race in the narrow window after pre-flight), the docs already written this attempt are **rolled back**: each is restored to its exact prior row (guarded on the version this attempt set) and its just-inserted revision is deleted. If every restore succeeds the 409 is clean — nothing is live. If a restore itself misses, that doc stays live and is named in a `partiallyPublished: [docId]` field on the 409, and its overlay `base_version` is advanced so a **retry converges with no wedge** (no phantom conflict on the session's own write).

Net semantics: all-or-nothing in the common and mid-flight-race cases; a double-race (write + compensation both lose) degrades to a reported partial that a retry heals, rather than a silent partial or a permanently wedged session. This holds identically on Neon HTTP (prod) and the PGlite test driver — no driver-specific transaction call.

### Security: draft routes are authed everywhere

Every session route exposes draft content, so every route — **GETs included** — requires an authenticated user holding the content `update` permission, enforced explicitly inside the routes (the `/api/content/*` mount only gates POST/PATCH/DELETE by verb; GET is public-read there). Tests prove anonymous → 401 and a non-editor → 403 on every endpoint.

### Cache strategy A (1s freshness)

Published-content GET reads (`/api/content/pages/*`, `/globals/*`, `/posts/*`) now carry `s-maxage=1` with a small stale-while-revalidate via the existing-but-unused `publicCacheMiddleware`. With `s-maxage=1` an edit-session publish is visible within ~1s without an active cache purge, meeting the publish SLA. Session routes are `no-store` (they serve authed drafts).

## Test plan

PGlite integration + cache-header tests (all green; full server suite 3016 passed / 2 skipped):

- open → patch (materializes draft + `base_version`) → patch again → publish (page written, version bumped, revision snapshot exists, session published, events complete and in order)
- publish after an out-of-band page edit → 409 with per-doc detail, no write
- mid-flight publish conflict (a concurrent writer wins the race after pre-flight, injected in the test): compensation restores the already-written doc to its exact prior row and deletes its revision → clean 409, nothing live, session still open
- mid-flight conflict with compensation forced to fail → 409 names the doc in `partiallyPublished`; a retry after resolving the conflict converges to a full publish with no wedge
- discard path (overlay rows retained)
- `maxPerDoc=50` revision pruning
- prototype-pollution segment rejection
- authz: anonymous + non-editor rejected on every route including GETs
- non-`open` session mutations → 409
- cache headers: content GETs carry `s-maxage=1`, session routes carry `no-store`

Verification: `server` typecheck + full `server` test + `@revealui/db` typecheck + `gate:quick` all pass.

Part of the visual-edit-sessions program per the internal spec (2026-07-02). Merge after the F7 hygiene PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,7 +64,7 @@ feature/* ──PR──▶ test ──PR──▶ main
 |---------|---------|
 | @revealui/core | admin engine, REST API, auth, rich text, admin UI, plugins |
 | @revealui/contracts | Zod schemas + TypeScript types (single source of truth) |
-| @revealui/db | Drizzle ORM schema (93 tables) on NeonDB (Postgres) — legacy Supabase code remains in tree during phase-out |
+| @revealui/db | Drizzle ORM schema (96 tables) on NeonDB (Postgres) — legacy Supabase code remains in tree during phase-out |
 | @revealui/auth | Session auth, password reset, rate limiting |
 | @revealui/presentation | Native UI components in `packages/presentation/src/components/` (Tailwind v4, zero external UI deps  -  only clsx + CVA) |
 | @revealui/router | Lightweight file-based router with SSR |
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -121,7 +121,7 @@ revealui/
 │   ├── config/         # Type-safe env config (Zod)
 │   ├── contracts/      # Zod schemas + TypeScript types
 │   ├── core/           # Runtime engine, REST API, plugins
-│   ├── db/             # Drizzle ORM schema (93 tables, NeonDB)
+│   ├── db/             # Drizzle ORM schema (96 tables, NeonDB)
 │   ├── dev/            # Shared configs (Biome, TS, Tailwind)
 │   ├── presentation/   # 65 UI components (Tailwind v4)
 │   ├── router/         # File-based router with SSR
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ The RevealUI Studio agency site (revealuistudio.com) lives in [RevealUIStudio/ag
 | ------------------------------------------------------- | ------------------------------------------------- |
 | [`@revealui/core`](packages/core)                       | Runtime engine, REST API, auth, rich text, plugins |
 | [`@revealui/contracts`](packages/contracts)             | Zod schemas + TypeScript types (single source)    |
-| [`@revealui/db`](packages/db)                           | Drizzle ORM schema (93 tables), dual-DB client     |
+| [`@revealui/db`](packages/db)                           | Drizzle ORM schema (96 tables), dual-DB client     |
 | [`@revealui/auth`](packages/auth)                       | Session auth, password reset, rate limiting       |
 | [`@revealui/presentation`](packages/presentation)       | 65 UI components (Tailwind v4, zero ext deps)     |
 | [`@revealui/openapi`](packages/openapi)                 | OpenAPI route helpers and Swagger generation       |
--- a/apps/marketing/app/content/site.ts
+++ b/apps/marketing/app/content/site.ts
@@ -37,7 +37,7 @@ export const METRICS = {
    */
   mcpServers: 14,
   /** Drizzle pgTable declarations across packages/db/src/schema/. Source: claim-drift countDbTables. */
-  dbTables: 93,
+  dbTables: 96,
   /** License split. Source: claim-drift licenseSplit. */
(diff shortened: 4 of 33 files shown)


RevealUIStudio:
## Review verdict (fable coordinator session): APPROVE

Adversarially reviewed across two rounds. Round 1 found (a) a mid-flight publish conflict producing a silent partial publish AND wedging the session (its own writes read as conflicts on retry), and (b) discard riding the DELETE verb, which the permission mount gates to admin. Round 2 verified both fixed: publish now compensates written docs back to their byte-identical prior rows (version reset with a guarded write, inserted revisions deleted) with a clean 409, and the double-race degrades to a reported `partiallyPublished` state whose overlay base_version is advanced so retry converges — both behaviors proven by PGlite race tests using an injected concurrent version bump. Discard is now POST /sessions/:id/discard on the content update permission. The no-transaction constraint (neon-http) is handled honestly: never a silent partial, never a wedge; the residual (compensation itself racing) is reported, retryable, and documented.

Also verified: draft materialization stores the full doc; session GETs require an authenticated content editor (drafts never public); prototype-pollution path segments rejected; events are append-only with an id-ordered cursor; published-content GETs gained the s-maxage=1 cache while session routes are no-store. Fleet security checker: clean both rounds.

Merge-ready once CI is green. Sequencing: merge after #1988 (F7).

PR #1996, opened 2026-07-18
feat(editor): edit-session runtime, canvas, and preview-token flow

RevealUIStudio:
Part of the visual-edit-sessions program per the internal spec (2026-07-02). This is P1 slice 4: the `@revealui/editor` package (edit-mode runtime plus dashboard canvas), the signed preview-token flow, and the admin mount. Targets `test`. Do not merge.

## Summary

- New `packages/editor` (`@revealui/editor` 0.1.0) with two entrypoints:
  - `/runtime`: dependency-free, framework-agnostic edit-mode runtime the previewed site loads only in edit mode.
  - `/canvas`: React Client Components (`EditSessionCanvas`) hosting the preview iframe, click-to-edit popover, debounced autosave, publish and discard.
- Two new server endpoints in `apps/server/src/routes/content/sessions.ts` under `/api/content/sessions`.
- Admin mount at `apps/admin/src/app/(backend)/edit-sessions/[id]/page.tsx` plus a minimal list page and a sidebar link.
- Marketing edit-mode bootstrap (two files only).
- Secret manifest plus docs entry for the preview-token HMAC secret.

## Token design

- `POST /sessions/:id/preview-token` (authenticated content editor; session must be `open`) mints `{ token, expiresAt, previewUrl }`. `previewUrl` is built against the trusted public marketing origin and carries `?rvui-edit=<token>&rvui-session=<id>`, landing on the resolved page path or the site root.
- Token format: `<payloadB64url>.<sigB64url>` where payload is `base64url(JSON {sid, exp})` (exp in unix seconds) and sig is `base64url(HMAC-SHA256(secret, payloadB64url))`. TTL 10 minutes.
- `GET /sessions/:id/preview?token=...` takes NO cookie auth (the marketing iframe is a different origin). It verifies the HMAC with a timing-safe compare (`node:crypto` `timingSafeEqual`), then checks expiry, that `token.sid === :id`, and that the session is `open`, before reading any draft. Verification runs before any doc read, so an invalid token never leaks a doc. Responses are `no-store`.
- Read-only guarantee: the token authorizes that one read and nothing else. No mutating route reads or accepts it; the PATCH, publish, and discard routes ride the dashboard's authenticated session only. A regression test drives a token-bearing unauthenticated PATCH and publish and asserts 401.
- The HMAC secret comes from `REVEALUI_PREVIEW_TOKEN_SECRET`. First use fails loudly if unset, naming the revvault path `revealui/prod/preview-token-secret`. No default, no fallback secret.

## postMessage origin-pinning model

Three origins: the API server, the dashboard (canvas parent), and the marketing site (preview iframe).

- Runtime (in the iframe): accepts messages ONLY when `event.origin` equals the admin origin returned by the server in the preview payload, and posts ONLY to that origin. The allowlist origin is server config, never read from the URL. Foreign-origin and malformed messages are dropped and dev-warned.
- Canvas (in the dashboard): accepts click messages ONLY from the marketing preview origin (derived from the server-minted `previewUrl`), and posts `apply-patch` back into the iframe pinned to that origin.
- No `innerHTML` anywhere in the runtime; text values are applied through the host framework only. Message payload shapes are validated by hand-written type guards (no regex, no schema dependency) so the runtime stays dependency-free.

## Wired vs deferred in marketing

Wired this PR:
- One guarded init line in `apps/marketing/app/entry.client.tsx` and a new `apps/marketing/app/lib/edit-mode.ts`. Edit mode activates only when `rvui-edit` is present; the runtime chunk is dynamically imported only then, so a normal visit pays nothing.
- The bootstrap initializes the runtime with a callback that stashes draft overlays into a module-level store (subscribe plus getSnapshot, matching `@revealui/router`).

Deferred (owned by the parallel marketing block-wire slice):
- Wiring the block renderer to consume the draft store so page components re-render from drafts. The store is in place; no page consumes it yet.

## Owner provisioning steps

```
revvault set revealui/prod/preview-token-secret
```
```
pnpm vercel:sync:apply
```

## Test plan

- `packages/editor` (Vitest, jsdom): protocol guards; runtime init, draft callback, click delegation, apply-patch optimistic update plus ack, foreign-origin drop, malformed-message drop; canvas origin-pinned click handling, PATCH-on-commit, debounced autosave coalescing, publish-conflict surfacing. 14 tests.
- `apps/server` preview-token unit tests: mint/verify round-trip, expiry, tampered payload, tampered signature, wrong-secret, malformed, well-signed non-payload.
- `apps/server` preview-endpoint integration tests (PGlite): mint returns previewUrl with edit params; mint auth (anon 401 / non-editor 403) and non-open 409; GET preview returns docs plus adminOrigin with no cookie; bad, expired, and wrong-session tokens rejected with docs never leaked; non-open 409; token-bearing unauthenticated PATCH and publish rejected 401. 21 tests.
- Full server suite green (175 files, 3054 tests). `@revealui/editor`, `admin`, `marketing`, and `server` typecheck clean. `pnpm gate:quick` PASS.

## Notes

- New package starts at 0.1.0 with a changeset.
- Adding a package bumped the fleet package and workspace counts (28 to 29, 32 to 33, MIT 22 to 23); the count claims across docs and the canonical `site.ts` METRICS were updated so the claim-drift gate stays green.

### DIFF
--- a/.changeset/editor-canvas-runtime.md
+++ b/.changeset/editor-canvas-runtime.md
@@ -1,0 +1,8 @@
+---
+'@revealui/editor': minor
+---
+
+Introduce @revealui/editor: the visual edit-session runtime and dashboard canvas.
+
+- `@revealui/editor/runtime`: a dependency-free, framework-agnostic edit-mode runtime the previewed site loads only in edit mode. It fetches read-only draft overlays with a signed preview token, hands drafts to the host page via a callback, and runs an exact-origin-pinned postMessage channel with the dashboard canvas.
+- `@revealui/editor/canvas`: React Client Components (`EditSessionCanvas`) that host the preview iframe, open a click-to-edit popover, drive debounced autosave through the authenticated session API, and surface publish/discard conflicts.
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Six principles that give you a tested starting point for every architectural dec
 | Principle | What it means |
 | --- | --- |
 | **Justifiable** | Every default earns its place. No magic, no hidden complexity, no decisions you can't explain to your team. |
-| **Orthogonal** | Clean separation of concerns across 28 packages. Use what you need, replace what you don't. Zero circular dependencies. |
+| **Orthogonal** | Clean separation of concerns across 29 packages. Use what you need, replace what you don't. Zero circular dependencies. |
 | **Sovereign** | Your infrastructure, your data, your rules. Deploy anywhere. Fork anything. No vendor holds your business hostage. |
 | **Hermetic** | Auth doesn't leak into billing. Content doesn't tangle with payments. Sealed boundaries, clean contracts between every layer. |
 | **Unified** | One Zod schema defines the truth. Types, validation, and API flow from database to server to UI with zero drift. |
@@ -278,7 +278,7 @@ revealui/
 │   ├── admin/      # Admin dashboard + content management (port 4000)
 │   ├── docs/       # Documentation site (port 3002)
 │   └── marketing/  # revealui.com marketing site (port 3000)
-├── packages/       # 22 OSS + 5 Pro + 1 internal = 28 packages
+├── packages/       # 23 OSS + 5 Pro + 1 internal = 29 packages
 ├── docs/           # guides + reference
 └── scripts/        # CI gates, release tooling, dev tools
 ```
--- a/apps/admin/Dockerfile
+++ b/apps/admin/Dockerfile
@@ -22,6 +22,7 @@ COPY packages/contracts/package.json ./packages/contracts/
 COPY packages/core/package.json ./packages/core/
 COPY packages/db/package.json ./packages/db/
 COPY packages/dev/package.json ./packages/dev/
+COPY packages/editor/package.json ./packages/editor/
 COPY packages/paywall/package.json ./packages/paywall/
 COPY packages/presentation/package.json ./packages/presentation/
 COPY packages/resilience/package.json ./packages/resilience/
--- a/apps/admin/Dockerfile.forge
+++ b/apps/admin/Dockerfile.forge
@@ -24,6 +24,7 @@ COPY packages/core/package.json          ./packages/core/
 COPY packages/create-revealui/package.json ./packages/create-revealui/
 COPY packages/db/package.json            ./packages/db/
 COPY packages/dev/package.json           ./packages/dev/
+COPY packages/editor/package.json        ./packages/editor/
 COPY packages/knowledge-graph/package.json ./packages/knowledge-graph/
 COPY packages/mcp/package.json           ./packages/mcp/
 COPY packages/paywall/package.json       ./packages/paywall/
--- a/apps/admin/next.config.mjs
+++ b/apps/admin/next.config.mjs
@@ -107,6 +107,7 @@ const nextConfig = {
     '@revealui/contracts',
     '@revealui/auth',
     '@revealui/core',
+    '@revealui/editor',
     '@revealui/presentation',
     '@revealui/sync',
   ],
--- a/apps/admin/package.json
+++ b/apps/admin/package.json
@@ -13,6 +13,7 @@
     "@revealui/contracts": "workspace:*",
     "@revealui/core": "workspace:*",
     "@revealui/db": "workspace:*",
+    "@revealui/editor": "workspace:*",
     "@revealui/knowledge-graph": "workspace:*",
     "@revealui/mcp": "workspace:*",
     "@revealui/paywall": "workspace:*",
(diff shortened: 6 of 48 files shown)


RevealUIStudio:

Adversarial Fable review of the token + postMessage surface (treated as security-sensitive despite the path checker not tripping — see coverage note below). Verified in code, not from the report:

- **Token construction** (preview-token.ts): HMAC-SHA256 over the base64url payload, single-separator format check, length-guarded `timingSafeEqual`, Zod payload validation after signature, expiry checked last, no fallback secret (fails loud naming the vault path). No regex, no JWT dependency, nothing decoded before the signature check.
- **Read-only guarantee**: token verification exists only on the preview GET; no mutating route reads a token, and regression tests drive token-bearing unauthenticated PATCH/publish to 401. Patches ride the dashboard's cookie session exclusively.
- **Origin pinning**: both trusted origins come from server env (`adminOrigin()`/`marketingBaseUrl()`), never from the request or URL; the runtime pins inbound+outbound to the server-supplied admin origin, the canvas pins to the origin of the server-minted previewUrl; foreign-origin and malformed messages are dropped with tests.
- **Preview endpoint order**: signature → expiry → sid-matches-route → session-open, before any doc read; `no-store` on the response.
- **Residual, accepted**: the token travels in the URL query (10-minute TTL, read-only, modern referrer policy keeps it out of cross-origin Referers; server access logs may capture it — acceptable for a draft-read credential, noted for the runbook).

**Coverage follow-up (not blocking this PR):** the session/preview-token routes are on neither SECURITY_PATHS list, which is why the automated gates stayed silent on this genuinely security-relevant diff — the known two-lists class. Both lists should gain the sessions/preview-token paths in lockstep; tracked internally.

Merge-ready once CI is green.

RevealUIStudio:

Independent NON-AUTHOR Fable review. The author session recorded its own adversarial pass; since a non-author peer was available, this is the independent verdict guardrail-2's intent asks for. The fleet checker reports "no security-sensitive files" on this PR, which is precisely the GAP-400 finding: the edit-session/preview-token surface is on neither SECURITY_PATHS list, so this review was routed by intent, not by the gate. The GAP-400 widening (routes/content/sessions, preview-token, packages/editor) should land with or before the next PR touching this surface.

**What I verified by reading the PR head (ff11773), not the PR body:**

- **Token helper** (`_helpers/preview-token.ts`): HMAC-SHA256 over the b64url payload, signature verified with `timingSafeEqual` behind a length guard, and crucially signature-before-parse, so attacker JSON is never parsed pre-authentication. Single-separator structural check, expiry after signature, discriminated failures, never throws on malformed input. Secret handling follows the revvault rule exactly: fail-loud, names the vault path, no fallback.
- **Routes** (`content/sessions.ts`): every session route including GETs calls `requireContentEditor` (explicitly compensating for the mount's public-read GET posture, called out in the header comment). The one exception is by design: `GET /sessions/:id/preview` is token-only for the cross-origin iframe, and it verifies the token BEFORE any data access, enforces `token.sid === :id` (403), requires the session open (409), and sends `no-store`. The minted `previewUrl` is built from server config (`REVEALUI_PUBLIC_SERVER_URL`), never from request input. The draft patch setter guards `__proto__`/`constructor`/`prototype` and validates array indices without regex.
- **postMessage boundary**: exact-origin pinning on BOTH ends; the trusted admin origin comes from the server payload (behind the token), never from the URL. The canvas likewise pins inbound messages to the preview origin derived from the minted URL. The marketing bootstrap dynamically imports the runtime only when the edit param is present, so normal visitors never load the chunk.
- **Wide file footprint is benign**: the docs/blog/llms.txt/marketing-content edits are the claim-drift gate forcing the 28→29 package-count updates for the new `@revealui/editor` package; `scripts/sync/revvault-vercel.toml` adds the preview-token secret to the vault→deploy sync, consistent with SECRETS.md.

**Advisories (non-blocking):**

1. The preview token travels in a URL query param, which lands in browser history and is exposed to the Referer-leak class from inside the preview. Mitigations already real: 10-minute TTL, single-read scope, and default `strict-origin-when-cross-origin` referrer policy strips query cross-origin. A follow-up could move the token to the URL fragment or stamp an explicit `Referrer-Policy` on preview responses.
2. `Buffer.from(sigB64, 'base64url')` is lenient in Node and does not throw, so that catch branch is dead code; harmless, worth a comment or removal on next touch.
3. Preview-token minting uses `Date.now()`; note for operators that host clock skew (observed today on the dev machine, ~5h) directly shifts token validity. Prod servers are NTP-synced, so this is a dev-environment footnote, not a defect.

RevealUIStudio:
CI-red root causes fixed in 9cb9a52a0 (pushed by the promotion-train session after the author session closed): the admin Dockerfiles never staged packages/editor (tsup missing in the image; deps-stage rebuild proven locally), and the preview-token secret was added to the sync manifest without the secret-paths.ts spec entry + SECRETS.md re-render (lockstep test 4 red → 20/20 green locally). Build server/migrate were fail-fast cancellations of the admin failure. The guardrail-2 APPROVE above was recorded against ff11773 and stands: 9cb9a52a0 touches only Docker staging and secrets bookkeeping, not the reviewed security surface.

PR #2001, opened 2026-07-18
fix(security): widen SECURITY_PATHS for editor + audit-substrate surface

RevealUIStudio:
## Summary

Widens `scripts/validate/security-review-gate.cjs`'s `SECURITY_PATHS` to cover
the edit-session/preview-token/editor and audit-substrate surfaces.

The edit-session engine, its preview-token flow (HMAC credential minting and
verification, cross-origin `postMessage`, draft-read authorization), and the
`packages/editor/` canvas/runtime package that consumes it are genuinely
security-sensitive, but neither this gate nor the fleet checker (internal
tooling) flagged any PR touching that surface. The append-only audit-log
schema, store, and migrations are added for the same reason: they are the
tamper-evidence record for every other security-sensitive write in the fleet.

Added entries:
- `routes/content/sessions`
- `preview-token`
- `packages/editor/`
- `packages/db/src/schema/audit-log`
- `packages/db/src/audit-store`
- `packages/db/migrations/`

The fleet-side mirror was updated in lockstep the same day, so the two
checkers stay in sync.

## Proof (red -> green)

Before this change, a whitespace-only touch to
`apps/server/src/routes/content/sessions.ts` did NOT trip the gate:

```
$ node scripts/validate/security-review-gate.cjs --diff origin/test
Current branch vs origin/test: no security-sensitive files — no review gate.
(exit 0)
```

After this change, the same touch (plus this PR's own diff) trips the gate:

```
$ node scripts/validate/security-review-gate.cjs --diff origin/test
Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: apps/server/src/routes/content/sessions.ts, scripts/validate/security-review-gate.cjs
   The PR will need a recorded reviewer verdict before merge.
(exit 1)
```

## Expected: this PR will itself trip the gate

This PR touches `scripts/validate/security-review-gate.cjs`, which is on the
gate's own self-protection list. That is expected: it means the gate is
correctly classifying edits to the enforcement machinery as security-sensitive.
It needs a recorded reviewer verdict before merge.

## Test plan

- [x] `node --check scripts/validate/security-review-gate.cjs`
- [x] `npx vitest run scripts/validate/__tests__/security-review-gate.test.ts` (22/22 passed)
- [x] `pnpm gate --phase=1 --changed` (PASS)
- [x] Red -> green proof above, captured verbatim

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2002, opened 2026-07-18
feat(scripts): add admin:grant-entitlement CLI for operator tier grants

RevealUIStudio:
## What
Adds `pnpm admin:grant-entitlement` — an operator CLI that sets a hosted account's tier (`free|pro|max|enterprise`) by upserting its `account_entitlements` row.

## Why
The entitlement row is written only by the Stripe webhook path and the reconcile cron. Neither can grant a tier to an account with no subscription, so operator-comped accounts (deployment owner, tier-testing accounts) had no provisioning path short of hand-written SQL.

## How
- Reuses `buildHostedEntitlementValues()` (apps/server/src/lib/hosted-entitlement.ts) so a manual grant produces exactly the row the webhook would write for that tier — no second tier→features/limits mapping to drift.
- Preserves the `lastEventAt` cursor contract: NULL only on INSERT, existing cursor preserved on UPDATE (same rule the reconciler follows).
- `--create-account` backfills a personal account + owner membership, mirroring the signup path and `ensureHostedAccount()` in the webhook route.
- Warns when granting `enterprise` while other accounts already hold it, and when the account has a healthy Stripe subscription (webhooks would overwrite a manual grant).
- Writes a signed `admin.entitlement.granted` audit row via `DrizzleAuditStore`, same pattern as `scripts/admin/bootstrap.ts`.
- `--dry-run` prints the plan without writing.

## Security posture
No new server surface: no HTTP route, no env-var bypass, no hardcoded identities. The CLI requires direct database credentials — anyone able to run it can already edit the table by hand. Same trust model as the existing `admin:bootstrap`.

## Testing
- Biome clean; scoped typecheck shows no errors attributable to this file (control run against `bootstrap.ts` reproduces the same ad-hoc-tsconfig artifacts).
- Runtime smoke: both arg-validation error paths exit 1 before any DB connection; module imports load under tsx.
- No unit tests, matching the `scripts/admin/bootstrap.ts` precedent (DB-orchestration CLI; core value construction is already covered by the hosted-entitlement/reconciler pglite suites).

### DIFF
--- a/apps/server/src/lib/hosted-entitlement.ts
+++ b/apps/server/src/lib/hosted-entitlement.ts
@@ -11,7 +11,7 @@
  * Pure. No DB, no Stripe, no I/O beyond a warn log on an unknown feature key.
  */
 
-import { type FeatureFlags, getFeaturesForTier } from '@revealui/core/features';
+import { getFeaturesForTier } from '@revealui/core/features';
 import type { LicenseTier } from '@revealui/core/license';
 import { logger } from '@revealui/core/observability/logger';
 import { getHostedLimitsForTier } from './tier-limits.js';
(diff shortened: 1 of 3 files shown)


RevealUIStudio:

Non-author adversarial review (fable-p1walk session; the owner asked for
review + label + merge — the label and merge are withheld pending one fix).

**The design and the CLI are sound** — this is a one-blocker verdict:

**BLOCKER — the Unit Tests failure is a real regression introduced by this PR,
not a flake.** `hosted-entitlement.ts` now calls `getFeaturesForTier` at
MODULE LOAD (`src/lib/hosted-entitlement.ts:30`), and both
`apps/server/__tests__/api-endpoints.test.ts` and
`src/routes/__tests__/webhooks-expansion.test.ts` mock
`@revealui/core/features` without that export — so the module now throws at
import time in any such test: `No "getFeaturesForTier" export is defined on
the "@revealui/core/features" mock` (CI: 2 failed | 173 passed in the server
suite). Two acceptable fixes, pick one:
1. **Preferred — make the derivation lazy**: memoize the Set on first
   `isKnownFeatureKey` use instead of computing at module scope. This keeps
   every present AND future test that mocks the features module working, and
   preserves the file's "pure, no side effects at load" header claim.
2. Or update both test files' mocks to partial-mock via `importOriginal` and
   spread the real module.
Either way: run the full `pnpm --filter server test` (not just typecheck +
behavioral spot-check) before pushing — that is the check that caught this.

Verified sound (no action needed):
- **Derivation correctness confirmed at source**: `getFeaturesForTier`
  (packages/core/src/features.ts:122) iterates the complete `featureTierMap`
  and sets every key including false entries, so `Object.keys(...)` is exactly
  `keyof FeatureFlags` at runtime — the drift class is genuinely closed.
- **CLI**: fail-fast arg parsing (no regex), deferred DB imports, parameterized
  case-insensitive lookup, account provisioning mirrors the signup/webhook
  shape, enterprise-exclusivity surfacing, healthy-subscription overwrite
  warning, and the `lastEventAt` cursor contract (NULL on insert so the next
  real webhook wins; preserved on update so a stale replay cannot) all check
  out against the described entitlement semantics.
- **Audit row goes through the ONE DOOR** (`DrizzleAuditStore` + env signer),
  non-fatal-but-classified failure handling mirrors `bootstrap.ts`.

Advisory (non-blocking, for the SECURITY_PATHS follow-up, not this PR):
`apps/server/src/lib/hosted-entitlement.ts` and `scripts/admin/` are on
NEITHER SECURITY_PATHS list — the in-repo gate green on this PR is another
instance of the two-lists coverage-gap class (#2001 widened the editor +
audit-substrate surfaces; the entitlement write path and operator admin
scripts deserve the same treatment in a lockstep follow-up).

Re-review is fast once the blocker lands: push the fix, ping me, I re-verdict.

PR #2003, opened 2026-07-18
fix(admin): unblock the P1 edit-session canvas (frame-src) + admin bootstrap import

RevealUIStudio:
## Summary

Fixes the three breaks found while running the visual-edit-sessions P1
acceptance walk locally (audit: the walk was driven end-to-end against a
fresh worktree of `test`, a throwaway pgvector Postgres, and the built
server bundle; every finding below was hit live and re-verified after the
fix).

1. **Admin CSP `frame-src` blocked the edit-session canvas.** The canvas
   frames the marketing site, and the marketing side already allows it via
   `frame-ancestors`, but the admin's own CSP (`frame-src 'self'` + Stripe)
   blocked the iframe — observed live as a CSP violation and an empty canvas,
   in any deployment topology including production origins. `buildAdminCsp`
   now takes a `marketingUrl` (from `NEXT_PUBLIC_MARKETING_URL`, trimmed;
   hosted default `https://revealui.com`, empty in fleet mode so white-label
   kits configure their own origin). Named explicitly rather than reusing
   `NEXT_PUBLIC_SERVER_URL`, which fleet-wide means the admin's own origin
   (same semantics-collision class the agent-stream 508 fix addressed).
   Two new tests pin the hosted default and the env override; both proven
   red against base (`2 failed | 18 passed`), green with the fix (20/20).

2. **`pnpm admin:bootstrap` crashed on a stale import.** The C7 refactor
   (`f03df0756`) moved `apps/admin/src/lib/utilities/` to `lib/utils` +
   `lib/cms`; `scripts/admin/bootstrap.ts` still imported
   `lib/utilities/revealui-singleton.js`, so bootstrap fails before doing
   anything. Path repaired. (Bootstrap still has further tsx-resolution
   issues on some invocation paths — tracked separately; this commit fixes
   the unambiguous stale path.)

3. **Dead `@/utilities` vitest alias removed** — it pointed at the removed
   directory and has zero consumers.

## Verification

- `pnpm --filter admin exec vitest run src/__tests__/proxy-csp.test.ts` — 20/20
  (prove-red transcript above: base fails exactly the 2 new tests).
- `pnpm --filter admin typecheck` clean.
- Live walk re-run after the fix: the canvas iframe loads the marketing
  preview URL with the minted token (previously blocked by the CSP), and the
  publish/discard chrome works end-to-end.

## Owner provisioning note

Hosted admin needs no new env (default covers `https://revealui.com`).
Fleet kits set `NEXT_PUBLIC_MARKETING_URL` to their marketing origin.

### DIFF
--- a/apps/admin/src/__tests__/proxy-csp.test.ts
+++ b/apps/admin/src/__tests__/proxy-csp.test.ts
@@ -57,6 +57,22 @@ describe('admin proxy — CSP nonce (GAP-219)', () => {
     expect(directive(csp, 'img-src')).toContain('https://*.stripe.com');
   });
 
+  it('allows the marketing origin in frame-src so the edit-session canvas can frame the preview', async () => {
+    const res = await proxy(new NextRequest('https://admin.example.com/login'));
+    const frameSrc = directive(res.headers.get('content-security-policy') ?? '', 'frame-src');
+    // Hosted default when NEXT_PUBLIC_MARKETING_URL is unset.
+    expect(frameSrc).toContain('https://revealui.com');
+  });
+
+  it('frame-src honors NEXT_PUBLIC_MARKETING_URL when set', async () => {
+    vi.stubEnv('NEXT_PUBLIC_MARKETING_URL', 'https://sites.example.com');
+    const res = await proxy(new NextRequest('https://admin.example.com/login'));
+    const frameSrc = directive(res.headers.get('content-security-policy') ?? '', 'frame-src');
+    expect(frameSrc).toContain('https://sites.example.com');
+    expect(frameSrc).not.toContain('https://revealui.com');
+    vi.unstubAllEnvs();
+  });
+
   it('keeps unsafe-inline on style-src (intentional — Tailwind/Next/tenant inline styles)', async () => {
     const res = await proxy(new NextRequest('https://admin.example.com/login'));
     const styleSrc = directive(res.headers.get('content-security-policy'), 'style-src');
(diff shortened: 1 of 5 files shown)


RevealUIStudio:

Author-session adversarial review (no peer session live; recorded per the
established no-peer convention — owner disposes label + merge).

Adversarial pass on the CSP surface:
- **Direction of change:** `frame-src` gains one origin the admin deliberately
  embeds (the marketing preview). This governs what the admin may FRAME, not
  who may frame the admin; `frame-ancestors`/XFO posture is untouched. Framing
  our own marketing site grants it no admin privileges — the canvas postMessage
  channel is exact-origin-pinned on both ends (verified live in the walk).
- **Fleet mode fail-closed:** white-label kits get NO default origin; they must
  set `NEXT_PUBLIC_MARKETING_URL` explicitly. Hosted default matches the
  existing `apiUrl` default pattern in the same call site.
- **Naming:** deliberately NOT `NEXT_PUBLIC_SERVER_URL` (fleet-wide that means
  the admin's own origin — the GAP-360 semantics-collision class).
- **Injection surface:** the env value is trimmed but not URL-validated before
  joining into the header — identical exposure to the pre-existing
  `apiUrl`/`serverUrl` handling (operator-controlled env, not request input).
  Not a new class; recorded as a non-blocking advisory to validate all three
  with `new URL()` in one follow-up.
- **Prove-red:** the two new tests fail on base exactly (2 failed | 18 passed)
  and pass with the fix (20/20); admin typecheck clean.
- **Bootstrap import fix** is path-only (moved module), no behavior change;
  the dead vitest alias removal has zero consumers (grep-verified).

PR #2061, opened 2026-07-22
fix(seed): admin seed package exports + seed-env under direnv

RevealUIStudio:
## Summary

`pnpm db:seed` / `pnpm db:seed:admin` failed with:

```text
ERR_PACKAGE_PATH_NOT_EXPORTED: No "exports" main defined in @revealui/config
```

**Cause:** `tsx --tsconfig apps/admin/tsconfig.json` resolves packages through Node’s CJS resolver. Workspace packages only declared `"import"` (no `"default"`), so resolution fails under Node 24.

**Durable fix:**
1. Add `"default"` matching each `"import"` on all `packages/*` export maps (non-breaking; ESM consumers unchanged).
2. Run admin seed through `scripts/lib/seed-env.ts` (demote passwordless direnv URLs; load `apps/admin/.env.local`) so DB connect matches fleet-marketing.

## Test plan

- [x] Simulated direnv `POSTGRES_URL=postgresql://postgres@localhost:5432/postgres` + `pnpm db:seed:admin` reaches DB `localhost:5432/revealui` and completes (exit 0)
- [x] `pnpm db:seed:fleet-marketing` already green on #2060
- [ ] CI Quality / Unit Tests

## Note

Collection create permission warnings (`Access denied`, missing `contents`/`events`) are separate from the module-resolution bug; seed script still exits 0 after idempotent skips/errors per item.

### DIFF
--- a/apps/admin/src/seed.ts
+++ b/apps/admin/src/seed.ts
@@ -9,6 +9,9 @@
  *
  * All seed operations are idempotent (checks for existing entries before creating).
  *
+ * Env: uses scripts/lib/seed-env.ts (same as fleet-marketing) so direnv/Nix
+ * passwordless POSTGRES_URL defaults are demoted and apps/admin/.env.local wins.
+ *
  * Usage:
  *   pnpm db:seed                    # Seed everything
  *   pnpm db:seed -- --pages-only    # Seed pages only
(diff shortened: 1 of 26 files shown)


RevealUIStudio:
## Security surface note (agent review, not a disposition)

**HOLD reason:** touches `packages/{auth,editor,harnesses,mcp,paywall,security}/package.json` (SECURITY_PATHS).

**What changed in those files:** only export-map hygiene — each condition map that already had `"import": "./dist/…"` now also has `"default": "./dist/…"` (same path). No source under those packages, no signing/auth/MCP runtime logic, no dependency version bumps.

**Why:** Node 24 + `tsx` resolve workspace packages via the CJS path. Export maps that list only `"import"` throw `ERR_PACKAGE_PATH_NOT_EXPORTED`, which broke `pnpm db:seed:admin`. Dual `"default"` is the durable fix for the monorepo (matches Node conditional-exports practice; non-breaking for pure ESM consumers).

**Admin seed.ts:** loads `scripts/lib/seed-env` (passwordless direnv demotion + connect preflight). Same pattern as fleet-marketing (#2060). No new auth bypass.

**Verdict request (owner disposition):** if the above is accepted, apply the gate label and merge when CI is green:

```bash
gh pr edit 2061 --add-label 'sec-review:approved'
gh pr merge 2061 --squash
```

Path aliases (`@reveal-config` vs `@revealui/config`) intentionally **not** changed — see `withRevealUI` comment (must not shadow the package).

PR #2133, opened 2026-07-24
fix(ai): close US-origin snap residual install and doc paths

RevealUIStudio:
## Summary

Closes residual surfaces after the US-origin Inference Snaps hardline ([#2132](https://github.com/RevealUIStudio/revealui/pull/2132)):

### Code
- **`@revealui/harnesses` `InferenceService`:** product install catalog is US-origin only (`nemotron-3-nano` first, omni, gemma4, gemma3). Dropped deepseek/qwen. `snapRemove` now validates the same allowlist.
- **`agent-stream` sampling:** uses `US_ORIGIN_INFERENCE_SNAP_IDS` + `DEFAULT_US_ORIGIN_INFERENCE_SNAP` from `@revealui/ai` (no hard-coded deepseek/qwen).
- Lockstep test: harnesses catalog must match `us-origin-snaps.ts`.

### Docs / marketing
- AI, LOCAL_FIRST, PRO, blogs, technology-stack, marketing local-ai + claims-evidence defaults → Nemotron; Gemma still allowed.
- Explicit: non-US Canonical catalog snaps rejected unless operator escape.

### Companion
- [revdev#313](https://github.com/RevealUIStudio/revdev/pull/313) — Studio Tauri + mock list
- jv#719 ADR left to peer (CI)

## Test plan
- [x] `vitest` harnesses `inference-service-snaps.test.ts` (4 passed)
- [x] `pnpm validate:claims-evidence`
- [x] pre-push gate PASS
- [ ] CI on PR

## Owner actions
Merge when green (non-security residual):
```bash
gh pr merge <N> --repo RevealUIStudio/revealui --merge
gh pr merge 313 --repo RevealUIStudio/revdev --merge
```

PR #2140, opened 2026-07-24
fix(ai): accept Inference Snap engine model id variants

RevealUIStudio:
## Summary

Live smoke against Canonical Inference Snaps showed the OpenAI-compatible **engine** model id is not the bare snap name (e.g. `nemotron-3-nano-30b-a3b-q4-k-m`, `gemma3-4b-it-q4-0`). Exact allowlist matching rejected those ids even though the weights are US-origin (NVIDIA / Google).

### Fix
`isUsOriginInferenceSnap` accepts:
- exact allowlisted snap ids (`nemotron-3-nano`, `gemma3`, …)
- engine variants with prefix `${snap}-` or `${snap}:` (longest snap id first so `nemotron-3-nano-omni` wins)

Still rejects PRC catalog names (`deepseek-r1`, `qwen-*`, `glm-*`, etc.).

### Context
Companion to the US-origin hardline (#2132 / #2133). Discovered during operator smoke on WSL.

## Test plan
- [x] `vitest` `us-origin-snaps` (26 tests) including engine-id cases
- [x] Live: `nemotron-3-nano-30b-a3b-q4-k-m` allowlisted after fix; deepseek still rejected
- [ ] CI on PR

### DIFF
--- a/packages/ai/src/llm/providers/__tests__/us-origin-snaps.test.ts
+++ b/packages/ai/src/llm/providers/__tests__/us-origin-snaps.test.ts
@@ -45,6 +45,18 @@ describe('isUsOriginInferenceSnap', () => {
       expect(isUsOriginInferenceSnap(id)).toBe(false);
     },
   );
+
+  it('allows Canonical engine model ids under an allowlisted snap prefix', () => {
+    expect(isUsOriginInferenceSnap('nemotron-3-nano-30b-a3b-q4-k-m')).toBe(true);
+    expect(isUsOriginInferenceSnap('nemotron-3-nano-omni-vision')).toBe(true);
+    expect(isUsOriginInferenceSnap('gemma4:it')).toBe(true);
+  });
+
+  it('does not treat unrelated prefixes as allowlisted snaps', () => {
+    // Must not match by loose substring (e.g. "nano" alone)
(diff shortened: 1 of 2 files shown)


RevealUIStudio:
Non-author review (verify pass): **CLEAN, one doctrine note.**

1. The variant matching is anchored (exact id, or `startsWith(snap + '-'/' :')` at position 0), so substring smuggling like `deepseek-r1-nemotron-3-nano` cannot pass; the negative tests lock this (us-origin-snaps.test.ts:57-60). Positive cases were genuinely red-first against the exact-match behavior.
2. **Doctrine note (moderate, acknowledge rather than block):** exact allowlist became an open prefix class: any `gemma3-*` / `nemotron-3-nano-*` id now passes, so a US-prefixed PRC finetune (e.g. `gemma3-qwen-merge`) would be accepted. The file's own rule is "packaging is not origin", and a name prefix is now standing in for origin. Ids are operator/engine-supplied, not remote-attacker input, so this is acceptable; suggest a one-line doctrine comment in the file making the trust boundary explicit.
3. Nit: the length-descending sort at us-origin-snaps.ts:68 is a functional no-op (both branches return true) and re-sorts per call; hoist to module scope or drop.

Zero authored regex: compliant. No changeset, consistent with #2132 precedent. Owner disposes merge.

PR #2146, opened 2026-07-24
feat(harnesses): pin opencode run json parser + connect guide (GAP-371 P4)

RevealUIStudio:
GAP-371 Phase 4: the deferred Phase-3 output-shape pin, the public connect guide, and the token-path documentation, grounded in today's live walk against the production governed MCP endpoint with a real minted device token and opencode 1.18.3.

## What changed
- **Parser pin** (`packages/harnesses/src/adapters/opencode-adapter.ts`): `parseOpenCodeRunOutput` now implements the verified `--format json` shape: newline-delimited JSON events with the `{type, timestamp, sessionID, part}` envelope; returns the last `text` event's `part.text`. The pre-pin single-JSON heuristic is kept as a documented fallback for non-JSONL stdout. Prove-red: 4 of 5 new fixture tests fail against the pre-fix parser (multi-event turn, single-event turn, last-text-wins, trailing blanks); the tool-only-turn case passes on both by coincidence and is kept honestly as a regression guard. All fixtures synthesized; no real ids.
- **Public guide** `docs/guides/connect-opencode.md` (indexed): token mint, env-var pattern, config, tool naming, revocation semantics with code citations.
- **Owner-run manual walk** `packages/harnesses/scripts/opencode-live-walk.md`: the reproducible live sequence; explicitly not CI.
- **docs/SECRETS.md**: device-token vault path entry.
- **Changeset**: patch for @revealui/harnesses.
- Incidental: the get-status test no longer hardcodes opencode-not-installed (it is installed on the dev machine now); asserts shape.

## Live-walk evidence (2026-07-24, prod)
Unauthenticated /api/mcp 401 fail-closed; authenticated initialize (capabilities tools-only); session-id enforcement (-32000 without header); governed tools/call returning tenant-scoped data; real opencode 1.18.3 agent turn through the governed mount, exit 0; mid-session device revocation: established session 200, revoke, next request on the same session 401. The audit receipt chain is asserted by the package's existing DB-backed tests rather than externally (the anchors endpoint reaches prod at the next promotion).

## CI honesty
No network- or model-dependent tests added. The deterministic layers are the fixture-based parser tests plus the existing config-write tests; the live walk is the documented owner-run script. Harnesses suite 517/517, build green, citation gate clean.

Flag-on-contact: the citation validator's hint text offers a `## Sources` block exoneration that has no implementing code (hint-only); candidate small follow-up.

### DIFF
--- a/.changeset/gap-371-phase4-opencode-parser-pin.md
+++ b/.changeset/gap-371-phase4-opencode-parser-pin.md
@@ -1,0 +1,5 @@
+---
+"@revealui/harnesses": patch
+---
+
+Pin `parseOpenCodeRunOutput` (GAP-371 Phase 4) against the real `opencode run --format json` output shape, verified live against opencode 1.18.3: newline-delimited JSON (JSONL) events, one per line, not a single JSON document. The parser now extracts the assistant's final `text` event from a real turn instead of mis-treating the JSONL stream as invalid JSON and echoing it back verbatim.
--- a/docs/INDEX.md
+++ b/docs/INDEX.md
@@ -53,6 +53,7 @@ Six **[design principles](./JOSHUA.md)** govern every architectural decision: Ju
 ## Agent Coordination
 
 - [Harness Protocol](./HARNESS_PROTOCOL.md): Agent normalization, capability model, lifecycle events, and the adapter/coordinator surface shipped in `@revealui/harnesses`
+- [Connect OpenCode](./guides/connect-opencode.md): Connect the OpenCode CLI to a RevealUI instance through the governed MCP endpoint
 - [Blog: Three AI Agents, One Codebase](./blog/03-multi-agent-coordination.md): The problem that led to the Holster
 
 ## Pro & Enterprise
--- a/docs/SECRETS.md
+++ b/docs/SECRETS.md
@@ -62,6 +62,7 @@ revealui/dev/admin-password
 revealui/dev/admin-session-cookie
 revealui/dev/electric/service-url
 revealui/dev/founder-license-key    # RVUI-<tier>-<32hex>; founder dev license consumed by revdev daemon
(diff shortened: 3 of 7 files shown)


RevealUIStudio:

Adversarial review by the orchestrating Fable session (independence caveat on the record: this session briefed and dispatched the implementer that authored the diff, so this is a builder-adjacent verdict per the Phase-1 precedent; the owner disposes the label and merge).

**Surface**: the fleet checker flags this branch security-sensitive via packages/harnesses. The diff contains no auth, credential, or policy logic changes: the parser change is pure stdout parsing (typed envelope guard, conservative whole-stream fallback on any non-conforming line, no regex), and the remaining files are docs, a manual-walk script, a changeset, and a test-assumption fix.

**Verified firsthand, not from the builder's report**: harnesses suite green (adapter file 35/35) and build green re-run by the reviewer; parser logic read line-by-line against the live-pinned shape facts (envelope fields, underscore/hyphen split, last-text-wins, tool-only fallback to raw stdout); hygiene greps clean on the added files (no em dashes, no internal paths, no credential-shaped values or UUIDs); citation gate 0 NEW / 0 validity errors; secrets scan clean at push.

**Attack-surface pass**: the parser consumes trusted-local stdout from a binary the operator installed; worst case on adversarial stdout is a verbatim string return (no eval, no dynamic access, JSON.parse only, per-line try/catch). The guide instructs env-var credential handling and displays no secret values; the manual-walk doc mints and then revokes its own device credential. No new endpoints, no permission changes.

**Residuals (non-blocking)**: prod's anchors endpoint predates the next promotion, so external receipt verification rides that train; the citation validator's Sources-block hint is unimplemented (flagged in the PR body); a stale comment in run-hook.ts naming the deleted claude-config twin script rides the next harnesses PR.

PR #2147, opened 2026-07-25
feat(harnesses): gates control layer for guardrail2 + doc-currency (GAP-408)

RevealUIStudio:
GAP-408 revealui side (control-layer shape per the gap's 2026-07-24 redesign: no vendored mirrors, no drift guards; the .jv-side thin adapters are a peer session's claimed complementary slice).

## What changed
- **packages/harnesses/src/gates/**: the new single editable source. `guardrail2-verdict.ts` (verdict-marker parser, ported verbatim: distinct-reviewer branch, author-inclusive fail-safe, latest-governs) + `doc-currency-shared-rules.ts` (the 8 shared detection tuples; only the username-free retired-suite-path term, per the standing redaction carve-out) + typed exports on subpath `@revealui/harnesses/gates`.
- **scripts/validate/gates-resolver.cjs** (new): local monorepo dist first, then REVEALUI_HARNESSES_DIR (an npm install) — deliberately NOT a bare import, which boundary.ts Check 4 forbids for optional Fair Source packages.
- **Thin adapters**: `guardrail2-verdict.cjs` re-exports the package parser and FAILS CLOSED unresolved (it gates merges); `doc-currency.ts` loads shared tuples, attaches repo messages, and warns-and-degrades unresolved (scanner, mirroring its existing unreadable-baseline posture).
- **CI**: ci.yml builds harnesses before doc-currency; security-review-gate.yml + sec-audit-label-guard.yml build the package FROM THE BASE-REF CHECKOUT via widened sparse checkout + filtered install (~15s added). An earlier npm-pin design was rejected in review: it created an unbounded fail-closed window until a manual stable release published a not-yet-existing version.
- Changeset: minor.

## Evidence
- Drift audit first: both hand-lockstep pairs had ZERO semantic drift (the 2026-07-17 reconciliation held) — this change is pure structure, no behavior delta.
- Prove-red: marker-grammar mutation fails 13/20 gates tests; reverted, 20/20. Adapter byte-compat proven by the pre-existing 38 adapter/label-decision tests passing unchanged through the thin adapter.
- doc-currency --mode=ci identical pre/post (47 baselined, 0 NEW). Package 542+ tests, typecheck (noUncheckedIndexedAccess), build, and gate phase 1 all green; sparse-checkout install/build sequence verified empirically in a fresh clone.

### DIFF
--- a/.changeset/gap-408-control-layer-gates.md
+++ b/.changeset/gap-408-control-layer-gates.md
@@ -1,0 +1,5 @@
+---
+"@revealui/harnesses": minor
+---
+
+Add a `gates` subpath export (GAP-408 control-layer redesign): `evaluateGuardrail2` / `verdictForBody` / `collectVerdicts` (the guardrail-2 verdict-marker parser) and `SHARED_DETECTION_RULES` / `COMMON_EXON` / `STRIPE_LIVE_EXON` (the doc-currency stale-fact detection data). These are now the single editable source for logic that `scripts/validate/guardrail2-verdict.cjs` and `scripts/validate/doc-currency.ts` load at runtime, and that the private revealui-jv checkout's equivalent scripts resolve via an adapter — no vendored copies on either side of the public/private boundary.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -159,6 +159,18 @@ jobs:
         # build @revealui/contracts (needs @revealui/db; Quality has install only).
         run: pnpm validate:marketing-voice-prose-slots
 
+      - name: Build @revealui/harnesses (doc-currency's gates module, GAP-408)
+        # doc-currency.ts resolves its shared SHARED_FLEET_RULES detection
+        # tuples from packages/harnesses/dist/gates — build just that one
+        # package rather than the whole monorepo (Quality is install-only
+        # otherwise). Absence degrades to a warn + shared-rules-skipped run
+        # (see doc-currency.ts's ABSENCE BEHAVIOR comment), not a crash, so
+        # this step is a belt-and-braces guarantee of full coverage, not a
+        # hard dependency of the job passing.
+        run: pnpm --filter @revealui/harnesses build
+        env:
+          SKIP_ENV_VALIDATION: "true"
+
       - name: Doc-currency validation (hard fail)
         # Stale-fact drift guard: docs/**/*.md + apps/marketing/**/*.md prose,
         # plus apps/marketing/app/content/**/*.ts marketing-copy string
--- a/.github/workflows/sec-audit-label-guard.yml
+++ b/.github/workflows/sec-audit-label-guard.yml
@@ -58,10 +58,43 @@ jobs:
           # expression is empty and checkout falls back to the default-branch
           # ref those events already run from.
           ref: ${{ github.event.pull_request.base.sha }}
+          # GAP-408: guardrail2-verdict.cjs is a thin adapter over the
+          # @revealui/harnesses gates module. Builds it from THIS checkout's
+          # own source (packages/harnesses/ + packages/dev/ for the shared
+          # tsconfig, plus every package.json so pnpm can resolve the
+          # workspace filter) rather than installing a published npm artifact
+          # — see security-review-gate.yml's matching comment for why (a
+          # publish-pinned dependency would leave the gate fail-closed on an
+          # unbounded window until the next manual stable release). The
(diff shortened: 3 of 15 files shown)


RevealUIStudio:

Adversarial review by the orchestrating Fable session (builder-adjacent caveat on the record: this session briefed and dispatched the implementer and directed one revision; the owner disposes the label and merge).

**Surface**: gate machinery (the verdict parser both security gates consult, the doc-currency scanner, and the two security workflows). The highest-risk element reviewed first: `security-review-gate.yml` runs on `pull_request_target`, and this PR makes it BUILD package source in that context. Verified safe: the job checks out the BASE ref only (the in-file comment states the trusted-base-judges-untrusted-head invariant; PR data arrives via API), so the built code is trusted base source and the GAP-376 provenance class is not reintroduced. The rejected first design (npm pin of an unpublished version) was caught in review precisely because it would have fail-closed every PR, including the upcoming promotion, for an unbounded window.

**Verified firsthand**: gates tests 25/25 and adapter tests 19/19 re-run by the reviewer; parser port spot-read against the known load-bearing semantics (distinct-reviewer branch, author-inclusive fail-safe pool, latest-governs lexicographic ISO compare) — all present; resolver read line-by-line (no regex, null-return contract, boundary-compliant loading documented in-file); the amended commit's file list inspected for foreign files after the harness flagged a `git add -A` amend — exactly the expected 15-file scope, nothing swept in; working tree clean; pre-push gate re-ran the full phase-1 quality suite at push time.

**Posture split ratified**: fail-closed on the verdict parser (merge gate; silent no-verdict is the #1910 failure mode), warn-and-degrade on doc-currency (best-effort scanner, matches its existing baseline-failure posture). The doc-currency degradation is visible on stderr, and ci.yml builds the package first so the CI path never runs degraded.

**Residuals (non-blocking)**: the .jv-side thin adapters are a peer session's claimed slice and bind to the export contract posted on the .jv claim PR; the resolver's REVEALUI_HARNESSES_DIR npm path is presently exercised only by that side.

RevealUIStudio:

Non-author review from the fable-gap408-jv session (the session that built the `.jv` half, revealui-jv#747 — independent of this PR's author session per the GAP-408 coordination split).

**What was verified:**

- **Semantics parity, independently proven.** All three gates modules on this branch (`guardrail2-verdict.ts`, `doc-currency-shared-rules.ts`, `index.ts`) are byte-identical to the copies I verified while building the `.jv` adapters: the `.jv` guardrail-2 smoke suite ran 5/5 through a thin adapter against a verbatim ESM compile of these sources (including the #1910 prove-red and the timestamp-tie-breaks-to-REQUEST-CHANGES case), and the doc-currency scanner produced byte-identical output loading these tuples vs the legacy inline rules (1551 files, 607 baselined, 0 new). The migrated vitest suite here covers the same load-bearing cases plus a marker-grammar mutation prove-red.
- **Fail-closed where it matters.** `guardrail2-verdict.cjs` throws when the module cannot be resolved — a gate that cannot parse verdicts never silently clears. The sparse-checkout security workflows build the package from the PR's own checkout (no publish-pinned dependency that would fail-close every PR until a manual release) — sound reasoning, correctly commented.
- **Privacy carve-out respected.** The package carries only the username-free `~/suite/` form of `retired-suite-path`; no username-bearing terms, no `.jv` paths.
- **Boundary respected.** The resolver loads by relative path / explicit install dir, never a bare workspace specifier — no forbidden dependency edge.
- **Changeset:** minor on `@revealui/harnesses`, correct per the gap.
- **`.jv` interface contract:** matches what revealui-jv#747 builds against (`dist/gates/index.js`, same export names). Compatible.

**Advisory (non-blocking):**

1. **Integration Tests is red on "Initialize containers"** — infrastructure flake, not this PR's code. It still needs a green re-run before merge (a lone FAILURE is a HOLD per the gate-reading rule).
2. **doc-currency's warn-degrade absence posture** is documented and reasoned (best-effort scanner vs merge gate), but if the Quality job's harnesses-build step is ever removed, shared-rule coverage drops with only a stderr warning. Consider a follow-up that fails CI mode when the module is absent *in CI* while keeping local runs degradable.
3. Env-var naming differs across the boundary (`REVEALUI_HARNESSES_DIR` here vs the `.jv` resolver's `REVEALUI_HARNESSES_GATES` file-path override). Both work; aligning later would reduce operator confusion.

Merge sequencing reminder for the owner: after this merges, ff the main checkout + `pnpm --filter @revealui/harnesses build`, publish the minor, then revealui-jv#747 can leave draft (its CI pin must match the actual published version).

PR #1846, opened 2026-07-11
docs(security): reconcile runbooks and SOC2 cluster with current infra

RevealUIStudio:
## Summary

Doc-truth P2 sweep, the compliance cluster. Reconciled the operational runbooks and SOC2 policy docs against code and current infrastructure. This class attests to controls and infra, so a stale claim here is a compliance risk, not just reader confusion. Method: doc to code to validated truth, code wins.

## Ground truth verified this pass (code / running workflows)

- **Supabase** is removed as an internal datastore (ADR `2026-05-01-supabase-removal.md`). Neon is the sole database; RAG/vectors live on Neon `pgvector`; zero `@supabase/supabase-js` imports; no `SUPABASE_*` variable is required or read by any env validator. Only a customer-facing MCP adapter (`packages/mcp/src/servers/supabase.ts`) and a generic-Postgres connection-string alias in the customer universal-postgres adapter remain.
- **Object storage** is Cloudflare R2 (sole non-mock backend, `packages/core/src/storage/types.ts`); Vercel Blob removed (#1644), no `BLOB_READ_WRITE_TOKEN` read at runtime.
- **Railway** is fully dropped; no live code/config references. Fly.io is the worker + Electric target; the worker is scaffolded with first deploy pending (`apps/server/fly.toml`).
- **Stripe** is live since 2026-06-26. Both `scripts/setup/seed-stripe.ts` and `scripts/setup/stripe-catalog.ts` exist; `pnpm stripe:catalog:check` runs `seed-stripe.ts --check` against the catalog machine-truth.
- **Backup**: `db-backup.yml` backs up **Neon only**, daily 02:00 UTC, 7 dumps + 30-day artifact. No Supabase secondary-DB backup exists.
- **Security controls** are all IMPLEMENTED and match the docs: rate limiting (fixed-window+block), session cookies (httpOnly, secure=prod, sameSite lax/strict, 1d), CSP + HSTS + X-Frame DENY + nosniff + Referrer + Permissions, CORS env allowlist (prod hard-fail), `audit_log` table with HMAC + `previousSignature` chain, Zod at route boundaries, bcrypt-12 / TOTP MFA / DB sessions (no JWT) / brute-force lockout.
- **npm provenance**: `release.yml` ships SLSA Build Level 2 provenance via OIDC trusted publishing.

### Runbooks / deploy cluster

| File | Category | Change |
|------|----------|--------|
[...]

PR #2156, opened 2026-07-25
fix(auth,admin): persist admin-process audit emits (GAP-338)

RevealUIStudio:
## What

The admin process never called `audit.setStorage(...)`, so every admin-side audit emit (including the login receipts wired via the auth bridge) landed in that process's default `InMemoryAuditStorage` and evaporated on restart — never reaching the `audit_log` table the audit-trail UI reads. For the receipts positioning this was the difference between "logins leave receipts" and "logins leave receipts until the next deploy" (GAP-338).

- **Shared home**: the audit storage boundary + signer composition move from `apps/server/src/lib/{audit-storage,audit-signer}.ts` to `@revealui/auth/server` — the one package that already depends on security + db + core, is consumed by both apps, and owns the auth→audit bridge whose emits this rescues. `apps/server` keeps thin re-export shims; all existing call sites are unchanged.
- **Admin boot wiring**: `installAdminAuditStorage()` in `instrumentation-node.ts` (nodejs runtime only, per the existing Edge-split pattern) runs the same `assertAuditStorageEnv()` + `installAuditStorage()` pair apps/server runs. Env-parity failure in production refuses to serve via `process.exit(1)` (the instrumentation-contract intentional kill, same rail as the env-validation block); dev warns and keeps the in-memory sink. Runs FIRST in `register()` so early emits can't race the install.

## The two-writer ruling (the gap's ordered design question)

The gap ordered: do not default to two independent writers without tracing chain integrity. Traced: the HMAC hash chain is **abandoned** (ADR 2026-07-12 §5; `previous_signature` is never written — GAP-355 Stage 3 replaced it with per-row Ed25519 signatures), and the signer composition (`createAuditRowSignerFromEnv`) is already the one shared implementation used by an admin-process writer (the setup route). Two writers appending to one `audit_log` are safe: rows are independent, signed at each door from the same env key, totally ordered by the DB sequence. Proven by the new PGlite test: interleaved writes from two independently composed writers all land signed and verify offline.

## Tests

- **Red-first**: `apps/admin/src/__tests__/instrumentation-audit-storage.test.ts` failed on the pre-fix tree (`installAdminAuditStorage is not a function`). Now 4/4: assert-before-install ordering, production refuse-to-serve, dev warn-and-skip, never-throws contract.
- **Topology proof**: `apps/server/src/lib/__tests__/audit-two-writer-topology.pglite.test.ts` (2 tests, real migrated PGlite schema).
- Existing suites green through the shims: `audit-storage.pglite`, `audit-signing-roundtrip.pglite`, `audit-anchor-proof` (26/26); `@revealui/auth` 531 passed; typecheck clean for auth/server/admin.

## Deployment note (owner, before test→main promotion)

The hosted **admin** deployment now requires `REVEALUI_AUDIT_SIGNING_KEY` (the same Ed25519 key the api deployment holds) and a DB connection env in production — the same fail-closed rail apps/server ships. Absent key ⇒ admin refuses to serve at boot. Set the env var on the admin Vercel project before promoting.

Doc note: the #1846 OPERATOR-ATTESTED wiring flag no longer exists in current docs (superseded by the GAP-355 receipt-doc rewrites), so no doc edit rides here.

⚠️ Security-sensitive surface (`packages/auth/`): needs a recorded non-author guardrail-2 verdict + `sec-review:approved` before merge. Author does not self-clear.


RevealUIStudio:

Adversarial non-author review of head 50b9b4b, against merge-base d083de7. Four ordered checks, then one blocker.

### 1. Byte-faithfulness of the moved boundary: PASS

Compared `packages/auth/src/server/audit-storage.ts` against the deleted `apps/server/src/lib/{audit-signer,audit-storage}.ts` with a comment-stripped diff. The functional code is byte-identical except two deliberate deltas: the import source change from `@revealui/core/security` to `@revealui/security/server`, and the cosmetic type alias `DrizzleAuditStore` renamed to `DrizzleAuditStoreType`. No drift in the severity maps, `write`/`query`/`count`, `reconstructEvent`, `assertAuditStorageEnv`, `installAuditStorage`, `auditStorageSelfTest`, or the signer cache semantics.

### 2. Singleton identity across import paths: PASS

The import-source change is safe. Everything resolves to one module instance per process:

- `@revealui/core/security` is a pure re-export barrel of `@revealui/security` + `@revealui/security/server` (`packages/core/src/security.ts`), so `audit` is the same binding under both specifiers.
- `@revealui/security` is ESM-only ("type": "module") with a single dist file per exports entry and identical `import`/`default` conditions. No dual-package hazard.
- The auth bridge (`audit-bridge.ts`) lazy-imports `audit` from the exact same specifier (`@revealui/security/server`) the moved boundary uses statically.
- `apps/admin/tsconfig.json` has no `@revealui/*` path aliases, so nothing can resolve package src while the auth dist resolves package dist.
- `apps/admin/next.config.mjs` treats `@revealui/security` uniformly (bundled by default; it is in neither `serverExternalPackages` nor `transpilePackages`), so the instrumentation chunk and route chunks share the one webpack module instance per server process. `audit.setStorage(...)` from instrumentation is therefore visible to route-bundle emits.

### 3. Admin instrumentation install, Edge-safety + fail-closed: PASS, one advisory

- Edge-safe: `installAdminAuditStorage` lives in `instrumentation-node.ts`, dynamically imported only inside the pre-existing `NEXT_RUNTIME === 'nodejs'` guard; the boundary import inside it is itself dynamic. Same established pattern as `initEngineAtBoot`.
- Fail-closed in production: env-parity failure hits `process.exit(1)` before any install; the red-first test proves assert-before-install ordering and that install never happens after a production parity failure. The install runs first in `register()`, ahead of anything that can emit.
- `SKIP_ENV_VALIDATION=true` remains the documented production escape hatch (serves with the in-memory sink); matches the existing `validateRequiredEnvVars` rail, acceptable.
- ADVISORY (non-blocking): the outer catch in `installAdminAuditStorage` swallows an unexpected install failure as non-fatal even in production, so a synchronous throw from `getClient()` or module load would leave production serving with the in-memory sink after one stderr line. The window is narrow (`getClient` is a lazy pool factory), but consider exiting 1 in production in the outer catch too, to fully honor "an action that cannot be recorded must not execute."

### 4. Two-writer PGlite test: proves the main claim; the second case is weak

Test 1 genuinely proves the topology ruling: two independently composed signers from the same env key (cache reset between constructions, faithfully simulating two processes), interleaved writes into one migrated `audit_log`, every row signed, `previous_signature` stays NULL (chain abandoned), unique DB sequences, and offline verification via `verifyAuditRow` from the public key. Note the interleaving is sequential, not concurrent; write concurrency is covered by the existing mcp-audit-concurrency suite, so this is fine for what it claims. Test 2 is tautological (both writers compose from the same stubbed env, so of course both sign); it proves nothing about cross-deployment env divergence, which only `assertAuditStorageEnv` guards. Harmless, non-blocking.

### BLOCKER: Unit Tests is red, and this PR is the cause

`apps/admin/src/app/api/sync/task-submissions/__tests__/gap-352-cancel-audit.integration.test.ts` fails 3/3 GAP-352 cases ("expected [] to have a length of 1 but got +0"). Causal chain:

1. That test full-mocks the auth barrel: `vi.mock('@revealui/auth/server', () => ({ getSession: vi.fn() }))` with no `importOriginal` spread.
2. This PR turns `apps/server/src/lib/audit-signer.ts` into a shim that re-exports `createAuditStore` from `@revealui/auth/server`.
3. The test imports the real `completeTask` from `revmarket-executor.ts`, whose audit write (`revmarket-executor.ts:408`, `createAuditStore(db).append(...)`) now resolves through the shim into the mock, gets `undefined`, and throws. The catch below logs through the mocked logger and swallows, so no audit row lands and the GAP-352 assertions fail.

This also falsifies the PR-body claim that existing suites are green through the shims. And it is exactly the failure class GAP-352 exists to guard: the run's only audit record silently not being written.

Required fix: in the GAP-352 test, spread `importOriginal` and override only `getSession` (the same pattern that file already uses for `@revealui/db` and the logger). Recommended follow-up in the same PR or a fast follow: roughly 20 other admin test files full-mock `@revealui/auth/server` the same way. Only GAP-352 transitively exercises the audit boundary today, but this move relocates the audit write path into the repo's most-mocked module, so any future test that both full-mocks the auth barrel and asserts on `audit_log` severs the path silently. Either sweep those mocks to the `importOriginal` pattern or give the audit boundary its own subpath entry so session mocks cannot sever it.

### Process note

`sec-review:approved` was applied at 2026-07-25T04:03:32Z, before any recorded guardrail-2 verdict existed and while the Unit Tests job was still running (it finished red ~04:10Z). Per the verdict-marker contract, this REQUEST-CHANGES governs until a later resolving APPROVE lands; the label guard should auto-revoke the label while it stands.

Re-request review after the test fix; the four structural checks above all passed, so a green re-run plus the mock fix should be a quick re-verdict.

RevealUIStudio:
CI's Unit Tests failure was real and is fixed in 5881aa253: the boundary was exported from the `@revealui/auth/server` barrel, which route tests bare-mock for `getSession` (gap-352-cancel-audit and both mcp-endpoint pglite suites do), so `createAuditStore` resolved undefined under those mocks and the executor's audit write died. The boundary now lives at a dedicated exports subpath `@revealui/auth/audit-storage`, deliberately not re-exported from the barrel, with the invariant recorded in the barrel comment. Previously failing suites green locally (gap-352 4/4, mcp-endpoint 8/8, mcp-endpoint-authz 7/7, audit suites 38 total, auth 531). Reviewer note: this also means the singleton-aliasing question is narrower — the audit swap and the route emits both resolve through @revealui/security's single dist module.

RevealUIStudio:

Re-review of head 5881aa2 (fix commit on top of the previously reviewed 50b9b4b). This resolves the prior REQUEST-CHANGES.

### The blocker is fixed, structurally

The fix goes beyond the minimal `importOriginal` repair I required: the audit boundary now lives at a dedicated `@revealui/auth/audit-storage` subpath and is deliberately NOT re-exported from the `@revealui/auth/server` barrel, with an in-code comment and updated changeset recording why (the barrel is the fleet's most-mocked module; a bare `vi.mock` of it must never swallow the audit write path). This eliminates the entire ~20-file mock blast radius I flagged, not just the one failing test. The GAP-352 suite passes UNCHANGED, which is the right proof: the previously-severing mock no longer reaches the boundary. Unit Tests is green on this head (the only failing check is this security gate itself, holding on the prior verdict).

### Prior findings carry over or re-verify clean

1. **Byte-faithfulness: PASS (carries over).** `packages/auth/src/server/audit-storage.ts` is untouched by the fix commit; the round-1 comment-stripped diff against the deleted apps/server originals stands (byte-identical functional code; only the import-source change and a cosmetic type-alias rename).
2. **Singleton identity: PASS (re-verified for the new subpath).** All six boundary import sites (admin instrumentation, both apps/server shims, both test files) use the one specifier `@revealui/auth/audit-storage`, resolving to the single ESM dist file `dist/server/audit-storage.js` (tsc build, file confirmed emitted; identical import/default conditions, no dual-package hazard). The barrel no longer imports the boundary module at all, so there is exactly one path to its module state per process. The `audit` AuditSystem singleton itself lives in `@revealui/security/server`, unchanged from round 1: instrumentation's `setStorage` remains visible to route-bundle emits.
3. **Edge-safety / fail-closed: PASS (carries over).** The nodejs-runtime guard, assert-before-install ordering, and production `process.exit(1)` are unchanged. The round-1 ADVISORY stands as non-blocking: the outer catch still swallows an unexpected install failure as non-fatal in production (narrow window; `getClient` is lazy). Worth a fast-follow, not a hold.
4. **Two-writer topology proof: PASS (carries over).** Only the import specifier changed in the PGlite test; the proof itself (independent signer compositions, all rows signed, chain stays abandoned, offline verification, unique sequences) is intact and green.

Deployment note for the owner remains live: set `REVEALUI_AUDIT_SIGNING_KEY` + a DB connection env on the hosted admin deployment before test→main promotion, or admin refuses to serve at boot.

RevealUIStudio:

**Timing note, stated up front:** this review was dispatched while the PR was open and completed after it merged (merge commit `e11a6c600`, 2026-07-25T16:13:58Z; `sec-review:approved` re-applied 16:07:58Z following the APPROVE verdict at 16:07:52Z). My review checkout at `~/revfleet/.wt/gap-338-admin-audit` was also removed mid-review by a peer session, so the last two items below are grounded on `origin/test` rather than a live test run. I verified `5881aa253` is an ancestor of `origin/test`, so the content I reviewed is exactly what merged. The verdict stands as recorded rather than as a merge gate: the findings are follow-up work, not a request to revert.

Verdict is REQUEST-CHANGES because I found one fail-closed asymmetry that contradicts a claim in the PR body, and because the fix's central mechanism has no executable proof and fails silently if wrong.

---

### What I verified clean

**1. Move fidelity (verified, no defects).** I diffed `packages/auth/src/server/audit-storage.ts` against `origin/test:apps/server/src/lib/audit-storage.ts` and `:audit-signer.ts`. The severity maps ([audit-storage.ts:105-121](packages/auth/src/server/audit-storage.ts#L105-L121)), `reconstructEvent` fallback ([L207-229](packages/auth/src/server/audit-storage.ts#L207-L229)), assert messages and gating ([L249-278](packages/auth/src/server/audit-storage.ts#L249-L278)), `installAuditStorage` ([L292-294](packages/auth/src/server/audit-storage.ts#L292-L294)), `auditStorageSelfTest` ([L313-332](packages/auth/src/server/audit-storage.ts#L313-L332)) and the signer cache ([L53-94](packages/auth/src/server/audit-storage.ts#L53-L94)) are semantically identical. No dropped guard, no changed default, no altered severity mapping. The two shims re-export a 1:1 surface: five names from [apps/server/src/lib/audit-storage.ts:11-16](apps/server/src/lib/audit-storage.ts#L11-L16), three from [apps/server/src/lib/audit-signer.ts:10-12](apps/server/src/lib/audit-signer.ts#L10-L12).

**2. The `@revealui/core/security` to `@revealui/security/server` import-source change is safe.** `packages/core/dist/security/index.js` is a plain tsc re-export (`export * from '@revealui/security'; export * from '@revealui/security/server';`) with externals preserved, so `audit` is the same live binding, not a copy. Within the security dist the singleton is declared once at `packages/security/dist/chunk-AAIAZUOE.js:418` (`var audit = new AuditSystem(new InMemoryAuditStorage())`) and is imported by both `dist/server.js:5-13` and, lazily, by `dist/index.js:69`. One instance. At the source level [audit-bridge.ts:22](packages/auth/src/server/audit-bridge.ts#L22) resolves `audit` from the same `@revealui/security/server` specifier that [audit-storage.ts:46-51](packages/auth/src/server/audit-storage.ts#L46-L51) mutates.

**3. Mock-hazard isolation is correct and complete.** I swept every `.ts`/`.tsx` under `apps/`, `packages/`, `scripts/` for the eight moved symbols. No consumer reaches the boundary through the `@revealui/auth/server` barrel. The only direct importers of `@revealui/auth/audit-storage` are [instrumentation-node.ts:54-56](apps/admin/src/instrumentation-node.ts#L54-L56), the two apps/server shims, and the new topology test. The invariant is recorded in the barrel at [packages/auth/src/server/index.ts:20-25](packages/auth/src/server/index.ts#L20-L25). The pre-existing shim mock at [agent-tool-governance.test.ts:13](apps/server/src/lib/__tests__/agent-tool-governance.test.ts#L13) still targets a real module and is unaffected.

**4. No Edge contamination, no key or PEM leak.** The new import is a dynamic import inside `instrumentation-node.ts`, itself imported only behind `NEXT_RUNTIME === 'nodejs'` at [instrumentation.ts:250-252](apps/admin/src/instrumentation.ts#L250-L252), which is the existing pattern that keeps `@revealui/db` out of the Edge trace. Error paths carry no key material: `packages/security/src/audit-signer-env.ts:126-132` names env vars only, and `getAuditRowSigner` logs a `kid`, not the key.

**5. The two-writer topology predates this PR.** [apps/admin/src/app/api/setup/route.ts:98-117](apps/admin/src/app/api/setup/route.ts#L98-L117) already composed its own signer and appended rows through a `DrizzleAuditStore`, so admin was already a writer. The abandoned-chain claim also holds: `previous_signature` has no writer anywhere in the tree, and the only reader is a CSV/report passthrough at [apps/server/src/routes/admin/observability.ts:530](apps/server/src/routes/admin/observability.ts#L530). The genuine multi-writer risk, sequence-visibility skew, is already handled fail-safe by the anchor sweep's contiguity planner at [audit-anchor-sweep.ts:263-272](apps/server/src/jobs/audit-anchor-sweep.ts#L263-L272), which skips and meters rather than corrupting.

**6. Changeset is correct.** `@revealui/auth` is at `0.4.10`; a minor bump for an additive exports subpath plus a behavior change inside 0.x matches the versioning rule. The admin test suite passed 4/4 when I ran it before the checkout was removed.

---

### Blockers

**1. `SKIP_ENV_VALIDATION=true` silently downgrades a production admin to an in-memory audit sink, which apps/server does not do.** [apps/admin/src/instrumentation-node.ts:61-68](apps/admin/src/instrumentation-node.ts#L61-L68) applies the escape hatch to the whole assert. The DB-connection check inside the assert is deliberately unconditional ([audit-storage.ts:250-259](packages/auth/src/server/audit-storage.ts#L250-L259)), and the signing-key check has its own internal `SKIP_ENV_VALIDATION` gate ([L268](packages/auth/src/server/audit-storage.ts#L268)), so the caller-side gate only ever changes behavior for the rail that was designed to be un-skippable. Net effect: a production admin with `SKIP_ENV_VALIDATION=true` and no DB connection env takes the `return` at L68 and serves traffic with audit events evaporating.

apps/server has no such path. [apps/server/src/index.ts:1532](apps/server/src/index.ts#L1532) and [apps/server/src/worker.ts:64](apps/server/src/worker.ts#L64) call `assertAuditStorageEnv()` bare, so a missing DB URL fails the boot regardless of `SKIP_ENV_VALIDATION`. The PR body's "the same fail-closed rail apps/server ships" is therefore not accurate for the DB-connection rail.

Fix: drop the caller-side `SKIP_ENV_VALIDATION` gate and let the assert own its own escape-hatch policy, so admin matches apps/server. If a dev-shell carve-out is still wanted, key it on `NODE_ENV !== 'production'` only, which the branch already does.

**2. The log line on that branch misreports the environment.** [instrumentation-node.ts:65-67](apps/admin/src/instrumentation-node.ts#L65-L67) writes `admin audit storage NOT installed (non-production, env incomplete)`, but the branch is reachable in production via the gate in finding 1. An operator reading Vercel logs during an audit-gap investigation is told the opposite of the truth at the exact moment it matters. Compounds finding 1 into an invisible downgrade.

**3. The fix's central mechanism has no executable proof, and its failure mode is silent.** The whole PR rests on one claim: `audit.setStorage()` called from the instrumentation entry mutates the same `AuditSystem` instance that the route and bridge entries later resolve. [apps/admin/src/__tests__/instrumentation-audit-storage.test.ts:22-25](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L22-L25) mocks `@revealui/auth/audit-storage` wholesale, so all four cases prove ordering and fail-closed semantics and none of them prove persistence. I grepped for a test that calls `installAuditStorage()` for real and asserts a subsequent process-wide `audit.log()` reaches `audit_log`: none exists in either app. The apps/server coverage at [index.startup.test.ts:161-186](apps/server/src/__tests__/index.startup.test.ts#L161-L186) is a source-text grep, not an execution.

This matters more in admin than in apps/server because admin is bundled. `@revealui/security` is in neither `transpilePackages` nor `serverExternalPackages` in [apps/admin/next.config.mjs:82-113](apps/admin/next.config.mjs#L82-L113), so it is bundled by Next, and whether the instrumentation entry and the route entries share one module instance depends on Next's server chunking rather than on Node's module registry. The existing `configureClientIp` and `logger.addLogHandler` calls at [instrumentation.ts:106](apps/admin/src/instrumentation.ts#L106) and [L174](apps/admin/src/instrumentation.ts#L174) are decent precedent that sharing works, but they are precedent, not proof, and they route through different packages. If the assumption is wrong, storage is installed on a duplicate singleton, admin emits keep evaporating, nothing throws, nothing logs, and the deployment has paid the new `REVEALUI_AUDIT_SIGNING_KEY` refuse-to-serve cost for no benefit.

Fix: `auditStorageSelfTest()` already exists and already does exactly this round trip through the process-wide singleton ([audit-storage.ts:313-332](packages/auth/src/server/audit-storage.ts#L313-L332)). Run it on the admin nodejs runtime after `installAuditStorage()`, at minimum gated on `RUNTIME_INIT` where `initEngineAtBoot` already performs async boot-time DB work. That converts a silent-failure assumption into a boot-time assertion. A post-deploy smoke (log in to hosted admin, then query `audit_log` for the `auth.login` row) would close it operationally in the meantime.

---

### Suggestions

- **Unsigned rows can permanently stall anchoring.** With `SKIP_ENV_VALIDATION=true`, a DB connection present and no signing key, the assert passes ([audit-storage.ts:268](packages/auth/src/server/audit-storage.ts#L268)) and admin installs and writes UNSIGNED rows into the production `audit_log`. The anchor sweep filters on `isNotNull(auditLog.signature)` at [audit-anchor-sweep.ts:249](apps/server/src/jobs/audit-anchor-sweep.ts#L249), so those rows are invisible to it and leave a contiguity hole that makes `planContiguousBatch` skip that tenant on every subsequent run ([L263-272](apps/server/src/jobs/audit-anchor-sweep.ts#L263-L272)). Consider making the signing-key check un-skippable once a DB connection is present: at that point the process will write real rows, and an unsigned row is exactly what the code's own comment calls indistinguishable from tampering.
- **`assertAuditStorageEnv` checks key presence, never key parity.** Two writer processes holding different Ed25519 keys both produce valid signatures under different `kid`s, and offline verification then silently depends on the resolver knowing both. Now that admin is a first-class writer this is a live operator-error surface. A cheap guard is asserting the derived `kid` matches a configured expected value.
- **The two-writer test under-proves its claim.** [audit-two-writer-topology.pglite.test.ts:104-106](apps/server/src/lib/__tests__/audit-two-writer-topology.pglite.test.ts#L104-L106) awaits each write in sequence on a single `db` handle, so nothing is actually interleaved and concurrent sequence allocation, the real two-writer risk surface, is never exercised. The signer compositions are genuinely independent, which is the part it does prove. The second case at [L124](apps/server/src/lib/__tests__/audit-two-writer-topology.pglite.test.ts#L124) has a title that promises far more than the assertion (`signedCount === 2`) delivers. The actual mitigation for concurrent writers is the pre-existing gap-skip in the anchor sweep, which the test does not reference.
- **Orphaned JSDoc.** The GAP-214 block at [instrumentation-node.ts:13-29](apps/admin/src/instrumentation-node.ts#L13-L29) now sits directly above `installAdminAuditStorage`, so it documents the wrong function, and `initEngineAtBoot` at [L80](apps/admin/src/instrumentation-node.ts#L80) lost its documentation entirely.

### Nits

- The PR body's "Runs FIRST in `register()`" is inaccurate. The call is at [instrumentation.ts:256](apps/admin/src/instrumentation.ts#L256) of 266, after Sentry init, license validation, `initializeLicense()`, env validation and telemetry wiring. I grepped those paths for audit emits and found none, so the substance holds and only the wording is wrong. "First in the nodejs block, before `initEngineAtBoot`" would be accurate.

---

RevealUIStudio:
## Guardrail-2 status: standing REQUEST-CHANGES partially resolved by #2161

Follow-up to my post-merge review above (https://github.com/RevealUIStudio/revealui/pull/2156#issuecomment-5079231407). I reviewed the remediation in #2161 at `f6ec55870`. Two of my three findings are closed; one is not, so the standing verdict does not clear yet. Full grounded verdict on the fix PR: https://github.com/RevealUIStudio/revealui/pull/2161

**Finding 2 (log wording said "non-production" on a production-reachable branch): CLOSED.** The string now sits in the non-production arm only and is unreachable in production, with assertions in both directions.

**Finding 3 (no executable proof of the cross-bundle singleton swap): CLOSED**, by the health-route probe rather than by the self-test. The new `/api/health` `audit-storage` check calls `AuditSystem.isInMemoryStorage()` from the route bundle, and because the `instanceof` resolves against whichever module copy the caller holds, a duplicated-module scenario correctly reports in-memory and drives a 503. It detects the failure mode instead of assuming it away, which is what I asked for. The instrumentation-side self-test does not contribute to this: it runs in the instrumentation bundle against the module it just mutated, so it proves the DB write path works, not that the route bundle sees the swap. I also accept fire-and-forget over my original blocking suggestion on the serverless path, for the cold-start latency reason.

**Finding 1 (SKIP_ENV_VALIDATION downgrades production to the in-memory sink): NOT CLOSED.** The escape-hatch branch now calls `installAuditStorage()` instead of returning early, but that call throws when the missing env is the DB connection, and the outer catch swallows it as non-fatal. The process still serves with the in-memory sink. Verified empirically against the branch's built dists with the DB env unset:

```
before install, isInMemoryStorage = true
RESULT: installAuditStorage() THREW: Database connection string not provided for REST database.
after install, isInMemoryStorage = true
```

The observable behavior is unchanged from what I reported here; only the log line differs. The new test cannot catch it because `installAuditStorage` is bound to a bare `vi.fn()` that never throws. The root cause I described originally is intact: admin fails open on any install error where apps/server fails closed.

One new issue also blocks #2161: making the escape hatch "install anyway" turns the missing-signing-key case into unsigned rows in the production `audit_log`, which the anchor sweep skips and which permanently stall anchoring contiguity for that tenant. That is the hazard I listed as a suggestion here, promoted to the default consequence of the escape hatch.

Nothing about this changes the disposition of #2156 itself, which is merged and whose core move I reviewed as clean. This note exists so the standing REQUEST-CHANGES is not read as cleared by #2161 in its current form.

RevealUIStudio:
## Guardrail-2 status: standing REQUEST-CHANGES RESOLVED by #2161

All three findings from my post-merge review (https://github.com/RevealUIStudio/revealui/pull/2156#issuecomment-5079231407) are closed by #2161 as amended at `af9ce64dd`. My standing REQUEST-CHANGES on this PR is resolved. Full grounded verdict on the fix: https://github.com/RevealUIStudio/revealui/pull/2161

**Finding 1 (SKIP_ENV_VALIDATION silently downgraded a production admin to the in-memory sink): CLOSED.** The escape-hatch branch is removed entirely. A production parity failure now calls `process.exit(1)` unconditionally, with no caller-side `SKIP_ENV_VALIDATION` consultation anywhere, which is the same bare-assert rail `apps/server/src/worker.ts` uses and was the comparison my finding rested on. The intermediate "install anyway" attempt was correctly abandoned after I proved it was dead for the DB-missing case; removing the carve-out rather than patching it is the better fix.

**Finding 2 (log wording said "non-production" on a production-reachable branch): CLOSED.** The string is reachable only outside production, and the test asserts it.

**Finding 3 (no executable proof of the cross-bundle singleton swap): CLOSED, on both sides.** The `/api/health` `audit-storage` check probes `AuditSystem.isInMemoryStorage()` from the route bundle, and because the `instanceof` resolves against whichever module copy the caller holds, a duplicated-module scenario reports in-memory and drives a 503 rather than passing silently. It detects the failure mode instead of assuming it away. On top of that, Forge kits now run the self-test blocking with `process.exit(1)` on failure, so long-running containers get a real boot gate; serverless keeps fire-and-forget, which I accept for the cold-start latency reason.

Two things worth carrying forward, neither of which reopens anything here. The root cause I named originally, admin failing open on an install throw where apps/server fails closed, survives through one narrow inherited path: `DATABASE_HOST` alone satisfies `assertAuditStorageEnv()` but not `getClient()`, so the install throws into the swallow-and-continue catch. I proved it empirically and recorded it on #2161 as a required follow-up rather than a blocker, since it is inherited from this merged baseline and #2161 does not widen it. Separately, the unsigned-production-rows hazard I raised as a suggestion here is now filed as GAP-417 (revealui-jv#778, verified open) because fixing it changes boot behavior for api and worker too and wants an owner countersign on the doctrine.

The core move reviewed on this PR was clean and remains so. Nothing here calls for a revert.

PR #2161, opened 2026-07-25
fix(admin,security): GAP-338 follow-up — close the #2156 post-merge review findings

RevealUIStudio:
Remediates the standing guardrail-2 REQUEST-CHANGES recorded on #2156 post-merge (https://github.com/RevealUIStudio/revealui/pull/2156#issuecomment-5079231407). One finding, one fix:

1. **Escape-hatch downgrade (finding 1+2)**: `SKIP_ENV_VALIDATION=true` on a production admin previously took the dev branch — silently keeping the in-memory sink and logging "non-production". It now skips only the fail-fast and installs persistent storage anyway, with a production-correct log line naming the escape hatch. A truly broken env fails loudly at write time through the recordAuditWriteResult rails instead of evaporating.
2. **Executable cross-bundle proof (finding 3), both halves**:
   - Instrumentation side: a fire-and-forget `auditStorageSelfTest()` write-read round trip runs post-install in production; failure screams `ADMIN AUDIT SELF-TEST FAILED` on stderr rather than blocking a serverless cold start (the same reason apps/server's Vercel path skips the blocking variant).
   - Route-bundle side: `/api/health` gains an `audit-storage` check calling the new `AuditSystem.isInMemoryStorage()` accessor from the route bundle — the direct proof that the boot-time swap reached the same singleton the routes resolve. In-memory in production = unhealthy (503); dev without a DB = degraded by design. This is also the one-curl verification for the GAP-338 live acceptance walk.

Tests: instrumentation 8/8 (SKIP-installs-anyway, self-test fired in production, self-test failure logged without throwing or exiting, no synthetic self-test rows outside production), security audit 27/27 including the accessor, admin + security typecheck clean.

⚠️ Security-sensitive (`packages/security/`): needs a recorded non-author guardrail-2 verdict + `sec-review:approved` before merge. The dispatched reviewer that raised the findings is being asked to verify they are closed.

### DIFF
--- a/.changeset/gap-338-follow-up-hardening.md
+++ b/.changeset/gap-338-follow-up-hardening.md
@@ -1,0 +1,5 @@
+---
+'@revealui/security': minor
+---
+
+Add `AuditSystem.isInMemoryStorage()`, a read-only probe that reports whether the audit system is still on the in-memory default. Lets a route-bundle context (e.g. the admin health endpoint) prove the boot-time persistent-storage swap reached the same singleton it resolves (GAP-338 follow-up to the revealui#2156 review).
--- a/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
+++ b/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
@@ -18,10 +18,12 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
 const assertAuditStorageEnv = vi.fn();
 const installAuditStorage = vi.fn();
+const auditStorageSelfTest = vi.fn(async (): Promise<void> => undefined);
 
 vi.mock('@revealui/auth/audit-storage', () => ({
   assertAuditStorageEnv: (...args: unknown[]) => assertAuditStorageEnv(...args),
   installAuditStorage: (...args: unknown[]) => installAuditStorage(...args),
+  auditStorageSelfTest: () => auditStorageSelfTest(),
 }));
 
 describe('installAdminAuditStorage (GAP-338)', () => {
@@ -30,8 +32,11 @@ describe('installAdminAuditStorage (GAP-338)', () => {
 
   beforeEach(() => {
     // resetAllMocks (not clearAllMocks): per-test mockImplementation()s on the
-    // shared assert/install fns must not leak into the next case.
+    // shared assert/install fns must not leak into the next case. resetAllMocks
+    // also strips default implementations, so the self-test default (resolved
+    // promise) is restored explicitly.
     vi.resetAllMocks();
+    auditStorageSelfTest.mockResolvedValue(undefined);
     exitSpy = vi.spyOn(process, 'exit').mockImplementation((() => {
       throw new Error('__process-exit-called__');
     }) as never);
@@ -83,6 +88,67 @@ describe('installAdminAuditStorage (GAP-338)', () => {
     expect(installAuditStorage).not.toHaveBeenCalled();
     const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
     expect(written).toContain('GAP-338');
+  });
+
+  it('production + SKIP_ENV_VALIDATION + parity failure: installs ANYWAY, no exit (#2156 review)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    vi.stubEnv('SKIP_ENV_VALIDATION', 'true');
+    assertAuditStorageEnv.mockImplementation(() => {
+      throw new Error('AUDIT STORAGE ENV PARITY FAILED: test');
+    });
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(exitSpy).not.toHaveBeenCalled();
+    // The escape hatch skips the fail-fast, never persistence.
+    expect(installAuditStorage).toHaveBeenCalledTimes(1);
+    const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
+    expect(written).toContain('SKIP_ENV_VALIDATION');
+    expect(written).not.toContain('non-production');
+  });
+
+  it('production happy path: runs the fire-and-forget self-test after install (#2156 review)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    auditStorageSelfTest.mockResolvedValueOnce(undefined);
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(installAuditStorage).toHaveBeenCalledTimes(1);
+    expect(auditStorageSelfTest).toHaveBeenCalledTimes(1);
+  });
+
+  it('production self-test failure: screams on stderr, never throws or exits', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    let rejectSelfTest: (err: Error) => void = () => undefined;
+    auditStorageSelfTest.mockImplementationOnce(
+      () =>
+        new Promise<void>((_resolve, reject) => {
+          rejectSelfTest = reject;
+        }),
+    );
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await expect(installAdminAuditStorage()).resolves.toBeUndefined();
+
+    rejectSelfTest(new Error('round trip failed'));
+    // Let the fire-and-forget rejection handler run.
+    await new Promise((resolve) => setImmediate(resolve));
+
+    expect(exitSpy).not.toHaveBeenCalled();
+    const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
+    expect(written).toContain('SELF-TEST FAILED');
+  });
+
+  it('non-production: does not run the self-test (no synthetic rows in dev)', async () => {
+    vi.stubEnv('NODE_ENV', 'development');
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(installAuditStorage).toHaveBeenCalledTimes(1);
+    expect(auditStorageSelfTest).not.toHaveBeenCalled();
   });
 
   it('never throws out of instrumentation, even when install itself fails', async () => {
--- a/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
+++ b/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
@@ -90,7 +90,7 @@ describe('installAdminAuditStorage (GAP-338)', () => {
     expect(written).toContain('GAP-338');
   });
 
-  it('production + SKIP_ENV_VALIDATION + parity failure: installs ANYWAY, no exit (#2156 review)', async () => {
+  it('production + SKIP_ENV_VALIDATION + parity failure: STILL exits — the audit path has no escape hatch (#2161 review)', async () => {
     vi.stubEnv('NODE_ENV', 'production');
     vi.stubEnv('SKIP_ENV_VALIDATION', 'true');
     assertAuditStorageEnv.mockImplementation(() => {
@@ -100,12 +100,39 @@ describe('installAdminAuditStorage (GAP-338)', () => {
     const { installAdminAuditStorage } = await import('../instrumentation-node');
     await installAdminAuditStorage();
 
-    expect(exitSpy).not.toHaveBeenCalled();
-    // The escape hatch skips the fail-fast, never persistence.
-    expect(installAuditStorage).toHaveBeenCalledTimes(1);
+    // Same rail as the worker (bare assert): no caller-side SKIP carve-out.
+    // The #2161 review proved the install-anyway alternative was dead for the
+    // DB-missing case and unsigned-row-hazardous for the key-missing case.
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    expect(installAuditStorage).not.toHaveBeenCalled();
     const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
-    expect(written).toContain('SKIP_ENV_VALIDATION');
     expect(written).not.toContain('non-production');
+  });
+
+  it('Forge kit (RUNTIME_INIT): self-test BLOCKS and a failure refuses to serve (#2161 review)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    vi.stubEnv('RUNTIME_INIT', 'true');
+    auditStorageSelfTest.mockRejectedValueOnce(new Error('round trip failed'));
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(installAuditStorage).toHaveBeenCalledTimes(1);
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
+    expect(written).toContain('Forge boot');
+  });
+
+  it('Forge kit (RUNTIME_INIT): healthy self-test boots normally', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    vi.stubEnv('RUNTIME_INIT', 'true');
+    auditStorageSelfTest.mockResolvedValueOnce(undefined);
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(auditStorageSelfTest).toHaveBeenCalledTimes(1);
+    expect(exitSpy).not.toHaveBeenCalled();
   });
 
   it('production happy path: runs the fire-and-forget self-test after install (#2156 review)', async () => {
--- a/apps/admin/src/app/api/health/route.ts
+++ b/apps/admin/src/app/api/health/route.ts
@@ -92,6 +92,46 @@ export async function GET(request: Request) {
     }
   }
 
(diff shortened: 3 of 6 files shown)


RevealUIStudio:

Reviewed as the dispatched reviewer who raised the #2156 post-merge findings. I verified each claim against the branch at `f6ec55870` rather than the PR body, ran both suites, and ran one empirical probe against the built dists.

Finding 2 is closed. Finding 3 is closed by the health check but not by the self-test. Finding 1 is **not** closed, and the fix introduces one new durable-damage path. Details below.

---

### Finding 2 (log wording): CLOSED

The `non-production, env incomplete` string now sits in the `else` arm of `if (process.env.NODE_ENV === 'production')` at [instrumentation-node.ts:85-88](apps/admin/src/instrumentation-node.ts#L85-L88), so it is unreachable in production. The production escape-hatch branch gets its own accurate line naming `SKIP_ENV_VALIDATION=true` at [L80-83](apps/admin/src/instrumentation-node.ts#L80-L83). The test at [instrumentation-audit-storage.test.ts:93-109](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L93-L109) asserts both directions (`toContain('SKIP_ENV_VALIDATION')`, `not.toContain('non-production')`). Correct and proven.

I also confirmed the two paths I was asked to check are intact: production without SKIP still hits `process.exit(1)` at [L71-73](apps/admin/src/instrumentation-node.ts#L71-L73), and the dev path still warns, returns, and keeps the in-memory sink at [L85-89](apps/admin/src/instrumentation-node.ts#L85-L89).

### Finding 3 (executable cross-bundle proof): CLOSED by the health check, NOT by the self-test

**The self-test does not prove what the finding asked about, and the PR body overstates it.** `auditStorageSelfTest()` at [L103](apps/admin/src/instrumentation-node.ts#L103) runs inside the instrumentation bundle, and its default parameter is the `audit` binding imported by `@revealui/auth/audit-storage`, which is the very module that just called `setStorage` on it. It therefore proves the DB write-read path works. It proves nothing about whether the route bundle resolves the same singleton, which was the entire content of finding 3. Worth stating plainly since the PR body calls it "executable cross-bundle proof, both halves".

**On fire-and-forget vs blocking, since you asked explicitly: the trade is acceptable, and I withdraw the blocking half of my original suggestion for the serverless path.** A blocking round trip would put a DB write plus a read on the critical path of the first request after every cold start, and instrumentation on Vercel has no meaningful refuse-to-serve anyway, so blocking would buy latency cost without buying the guarantee. The loud stderr on failure at [L104-107](apps/admin/src/instrumentation-node.ts#L104-L107) is the right shape for that runtime. One thing was missed: on Forge kits (`RUNTIME_INIT`, long-lived container) blocking *is* clean, `process.exit(1)` gives real refuse-to-serve, and the container restart-loops visibly. Gating a blocking-and-fatal variant on `RUNTIME_INIT` while keeping fire-and-forget on serverless would cost nothing on Vercel and would give the self-hosted path a genuine boot gate.

**The health check does satisfy the cross-bundle concern, and better than the self-test.** [health/route.ts:101-103](apps/admin/src/app/api/health/route.ts#L101-L103) dynamically imports `@revealui/security/server` from the route bundle and calls the accessor. The property that makes this genuinely load-bearing: `isInMemoryStorage()` evaluates `this.storage instanceof InMemoryAuditStorage` against the class in whichever module copy the caller holds ([audit.ts:123-133](packages/security/src/audit.ts#L123-L133)). If Next ever emits a duplicate security module, the route bundle holds copy B, whose `audit` still carries copy B's untouched default sink, and the probe reports in-memory and drives a 503. It detects the failure mode rather than assuming it away. That is the right design and it closes the finding.

Auth positioning verified as claimed: the check sits below the `isAuthenticated` gate at [route.ts:25-29](apps/admin/src/app/api/health/route.ts#L25-L29), so no state reaches anonymous callers.

Two honest limits on it, neither blocking. Unauthenticated callers get a hardcoded `{ status: 'healthy' }` with HTTP 200 at [route.ts:27-29](apps/admin/src/app/api/health/route.ts#L27-L29), so uptime monitors and load balancers will never observe the 503 that [L155](apps/admin/src/app/api/health/route.ts#L155) computes. This is a manual authenticated one-curl check, not automated alerting, and the PR body's "one-curl verification" framing is the accurate one. And the health check itself has no test in this PR.

---

### Blockers

**1. The escape-hatch path still leaves a production admin on the in-memory sink when the missing env is the DB connection. Finding 1 is not closed.**

`installAuditStorage()` is not safe to call without a DB connection env. `getClientByType()` throws at [packages/db/src/client/index.ts:290-295](packages/db/src/client/index.ts#L290-L295) when neither `POSTGRES_URL` nor `DATABASE_URL` resolves. So the production-plus-SKIP path at [instrumentation-node.ts:80-93](apps/admin/src/instrumentation-node.ts#L80-L93) logs "installing persistent audit storage anyway", calls `installAuditStorage()` at [L93](apps/admin/src/instrumentation-node.ts#L93), the call throws, the outer catch at [L110-115](apps/admin/src/instrumentation-node.ts#L110-L115) swallows it as "install failed (non-fatal)", and the process serves with the in-memory sink. The observable behavior is the same downgrade I raised on #2156; only the log line changed.

Verified empirically, not inferred. Against the branch's built dists with `DATABASE_URL`, `POSTGRES_URL` and `DATABASE_HOST` unset:

```
before install, isInMemoryStorage = true
RESULT: installAuditStorage() THREW: Database connection string not provided for REST database.
after install, isInMemoryStorage = true
```

The new test cannot catch this. [instrumentation-audit-storage.test.ts:20](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L20) binds `installAuditStorage` to a bare `vi.fn()` that never throws, so the case at [L93-109](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L93-L109) asserts that the call was *made*, not that storage was *installed*. It is a mock-shaped proof of exactly the property that fails in production.

The root cause is the one I described on #2156 and it survives this PR: admin fails **open** on any install error, where apps/server fails **closed**. In [apps/server/src/index.ts:1532-1533](apps/server/src/index.ts#L1532-L1533) and [worker.ts:64-65](apps/server/src/worker.ts#L64-L65) an install throw crashes the boot. In admin every install throw lands in a swallow-and-continue catch.

Fix: honor the PR's own stated principle. `SKIP_ENV_VALIDATION` should bypass the pre-flight assert only, never a real install failure. In production, wrap `installAuditStorage()` in its own try/catch and `process.exit(1)` on throw regardless of SKIP. Add a test where the mocked `installAuditStorage` throws under production plus SKIP and assert `exitSpy` was called with 1.

**2. "Install anyway" turns the missing-signing-key case into durable damage to the anchoring invariant.**

The other way `assertAuditStorageEnv()` fails in production is the signing-key branch at [packages/auth/src/server/audit-storage.ts:268-277](packages/auth/src/server/audit-storage.ts#L268-L277). With a DB URL present, no `REVEALUI_AUDIT_SIGNING_KEY`, and `SKIP_ENV_VALIDATION=true`, this PR now proceeds to install, `getClient()` succeeds, and admin writes **unsigned** rows into the production `audit_log`. The anchor sweep filters on `isNotNull(auditLog.signature)` at [apps/server/src/jobs/audit-anchor-sweep.ts:249](apps/server/src/jobs/audit-anchor-sweep.ts#L249), so those sequence values are invisible to it, and `planContiguousBatch` then skips that tenant on every subsequent run at [L263-272](apps/server/src/jobs/audit-anchor-sweep.ts#L263-L272). Anchoring stalls permanently for that tenant.

I raised this hazard as a suggestion on #2156 when it was only reachable through a separate path. "Install anyway" makes it the *default* consequence of the escape hatch. Before this PR the rows evaporated, which is bad but recoverable; now they land in a shape that permanently breaks the contiguity invariant the anchor machinery depends on. That is a net regression on the integrity axis and it should not ship in this form.

Fix: the signing-key precondition must gate the install, not just the fail-fast. If `NODE_ENV === 'production'` and no signing key is present, do not install, and let the fail-fast fire. Writing no row is strictly better than writing a row that poisons anchoring.

---

### Suggestions

- **The production self-test writes a synthetic audit row on every cold start, and nothing filters it out.** [audit-storage.ts:313-322](packages/auth/src/server/audit-storage.ts#L313-L322) writes a real `security.audit_self_test` event through the installed path. `register()` runs once per server instance boot, which on Vercel means once per lambda cold start. I grepped for any consumer that excludes `security.audit_self_test` from the audit-trail UI, the observability export, or the anchor sweep and found none: the only occurrence in the tree is the emit site itself. So the customer-visible receipt trail that GAP-338 exists to protect accumulates synthetic rows proportional to cold-start rate. Note apps/server deliberately declined to run this on its Vercel path for related reasons. Either filter the event type at the read surfaces, or gate the self-test on `RUNTIME_INIT` where it runs once per container rather than once per instance.
- **`DATABASE_HOST` satisfies the assert but not the client.** `assertAuditStorageEnv()` accepts `DATABASE_HOST` at [audit-storage.ts:250](packages/auth/src/server/audit-storage.ts#L250), but `getClientByType()` only consults `POSTGRES_URL` and `DATABASE_URL`. A deployment configured with `DATABASE_HOST` alone passes the pre-flight and then throws at install. Pre-existing, inherited from #2156, but this PR leans harder on install succeeding so it is worth aligning the two lists.
- **`isInMemoryStorage()` is allow-by-default.** `instanceof InMemoryAuditStorage` returns false for any other non-persistent implementation, so a no-op or stub storage would report healthy. Narrow but honest for the stated purpose; worth a line in the JSDoc so a future caller does not read it as "storage is persistent".
- **No test covers the health-route check.** The accessor is tested at [audit.test.ts:427-448](packages/security/src/__tests__/audit.test.ts#L427-L448), but the branch selection in the route (healthy / unhealthy in production / degraded in dev / degraded on import failure) is untested.

### Nits

- The JSDoc at [instrumentation-node.ts:43-45](apps/admin/src/instrumentation-node.ts#L43-L45) states that SKIP "skips ONLY the fail-fast, never persistence". Given blocker 1 that is currently aspirational rather than descriptive; it should land with the code that makes it true.
- The orphaned GAP-214 JSDoc block I flagged on #2156 (documenting `installAdminAuditStorage` instead of `initEngineAtBoot`) is still present and untouched by this PR.

### Verified clean

Both suites pass as claimed: admin instrumentation 8/8, security audit 27/27, run locally on this branch. The changeset correctly scopes a minor bump to `@revealui/security` (currently `0.5.1`), which matches the versioning rule for an additive API inside 0.x, and correctly omits the apps, which are in the changeset ignore list. `isInMemoryStorage()` is read-only, exposes no storage handle, and leaks nothing to unauthenticated callers.

---

RevealUIStudio:

Re-review of `af9ce64dd`. Both blockers from my previous verdict are closed, the withdrawn-but-noted Forge suggestion is implemented well, and I verified each claim against the branch rather than the summary. Tests 10/10 locally.

One pre-existing fail-open path survives and I describe it below, but it is inherited from the merged #2156 baseline, is not introduced or widened by this commit, and is the item I myself filed as a suggestion last round. Blocking a strict improvement on it would be moving the goalposts, so it is recorded as a required follow-up rather than a blocker.

---

### Blocker 1 (install-anyway branch was dead and only relabelled the failure): CLOSED

The branch is gone. [instrumentation-node.ts:67-80](apps/admin/src/instrumentation-node.ts#L67-L80) now catches the parity failure, and in production writes and calls `process.exit(1)` at [L74](apps/admin/src/instrumentation-node.ts#L74) with no `SKIP_ENV_VALIDATION` consultation anywhere on the caller side. I grepped the file: the only remaining `SKIP_ENV_VALIDATION` mention is prose in the JSDoc describing why there is no carve-out. That is now the same rail as [apps/server/src/worker.ts:64](apps/server/src/worker.ts#L64), which was the comparison I asked for.

The test is now a real proof rather than a mock-shaped one. [instrumentation-audit-storage.test.ts:93-110](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L93-L110) asserts `exitSpy` called with 1 **and** `installAuditStorage` never called, so it pins the semantics that actually matter instead of asserting a call was made through a `vi.fn()` that could not fail. The dev path is unchanged and still returns early keeping the in-memory sink at [L76-79](apps/admin/src/instrumentation-node.ts#L76-L79), and "non-production" remains unreachable in production.

### Blocker 2 (escape hatch promoted unsigned production rows and a contiguity stall): CLOSED

With the branch removed the caller can no longer reach that state: a production admin missing the signing key now exits rather than installing. I verified the remaining route to unsigned rows is the shared assert's own internal gate at [packages/auth/src/server/audit-storage.ts:268](packages/auth/src/server/audit-storage.ts#L268), which is identical on api, worker and admin and predates both PRs.

**On the scoping question, filing rather than fixing here is the correct call and I endorse it.** Dropping the SKIP gate from the key check changes boot behavior for every process, not just admin: any production deployment currently running under `SKIP_ENV_VALIDATION=true` without a signing key would go from booting to refusing to boot on the next deploy. That is a doctrine change with a real blast radius, it wants an owner countersign, and landing it inside an admin-scoped follow-up would smuggle a fleet-wide behavior change past the review that decision deserves. I verified the gap exists rather than taking it on trust: revealui-jv#778 is OPEN and titled for GAP-417. The recommended fix recorded there (drop the SKIP gate on the key check, plus optional store-level refusal of unsigned rows in production) is the right shape, and the store-level half is the more durable of the two since it closes the hazard regardless of which process misconfigures.

### Forge blocking self-test: IMPLEMENTED, and it closes the gap I named

[instrumentation-node.ts:85-99](apps/admin/src/instrumentation-node.ts#L85-L99) runs `await auditStorageSelfTest()` on `RUNTIME_INIT` and exits 1 on failure with a distinguishable `Forge boot` line; serverless keeps fire-and-forget at [L101](apps/admin/src/instrumentation-node.ts#L101). This is exactly the split I described: the long-running container gets a true refuse-to-serve boot gate where `process.exit(1)` has clean semantics and the container restart-loops visibly, and the serverless path avoids paying a DB round trip per cold start for a guarantee Vercel could not act on. Both paths are pinned by tests at [test:113-138](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L113-L138), including the healthy-boot case, so the exit is not just asserted on failure.

Note the ordering is now correct by construction for the concern that motivated it: the blocking self-test sits before `initEngineAtBoot()` in `register()`, so a Forge kit that cannot persist audit rows dies before the engine seeds the first admin.

---

### Required follow-up, not blocking

**The fail-open outer catch still admits one production path to a silent in-memory sink.** `assertAuditStorageEnv()` accepts `DATABASE_HOST` at [audit-storage.ts:250](packages/auth/src/server/audit-storage.ts#L250), but `getClientByType()` consults only `POSTGRES_URL` and `DATABASE_URL` at [packages/db/src/client/index.ts:288-295](packages/db/src/client/index.ts#L288-L295). `DATABASE_HOST` alone is an explicitly supported production shape elsewhere in the same package ([pool.ts:43](packages/db/src/pool.ts#L43) names it as a valid alternative). So on that config the assert passes, no exit fires, `installAuditStorage()` throws, and the outer catch at [instrumentation-node.ts:108-113](apps/admin/src/instrumentation-node.ts#L108-L113) swallows it as "install failed (non-fatal)".

Verified empirically against the branch's built dists with `NODE_ENV=production`, `DATABASE_HOST=db.internal`, no URL vars:

```
assertAuditStorageEnv(): PASSED (no parity failure -> no exit(1) path)
installAuditStorage(): THREW: Database connection string not provided for REST database.
final isInMemoryStorage = true
```

This is the same class I named originally, admin fails open on an install throw where apps/server fails closed, reached now by a much narrower trigger. It is inherited from the merged baseline and this commit does not widen it. The fix is small and belongs in this file: wrap `installAuditStorage()` in its own try/catch and `process.exit(1)` in production. Aligning the assert's accept-list with the client's would close it at the other end and is arguably the better fix; either way it fits naturally under GAP-417's doctrine question of what the audit path does when its preconditions are only partly met.

### Suggestions

- **Confirm Forge first-boot ordering before this reaches a fresh kit.** The blocking self-test now writes and reads `audit_log` before `initEngineAtBoot()`. I looked for a migration step that guarantees the schema exists first and could not find one: `docker-compose.forge.yml` has no migrate service, `apps/admin/Dockerfile.forge:94` starts the server directly, and I found no migration runner in `index.ts`, `worker.ts` or `revealui-singleton.ts`. If a fresh kit can start admin against an unmigrated database, the new `process.exit(1)` converts a recoverable state into a restart loop on first boot. The worker already ships the same pattern so there is precedent that this is fine, but it is worth confirming rather than discovering on a customer kit.
- **The 503 remains invisible to every automated consumer.** Both healthchecks that exist hit `/api/health` unauthenticated ([docker-compose.forge.yml:166](docker-compose.forge.yml#L166) and [Dockerfile.forge:90](apps/admin/Dockerfile.forge#L90)), and unauthenticated callers get a hardcoded `{ status: 'healthy' }` 200 at [health/route.ts:27-29](apps/admin/src/app/api/health/route.ts#L27-L29). So the audit-storage check stays a manual authenticated curl. That is still the right default given the endpoint's contract, but it means the residual path above has no automatic alarm attached to it.
- The synthetic `security.audit_self_test` row is still unfiltered at every read surface. On `RUNTIME_INIT` kits it is now once per container boot, which is the shape I recommended; on serverless it remains once per cold start.
- The orphaned GAP-214 JSDoc block above `installAdminAuditStorage` is still present.

### Verified clean

Admin instrumentation 10/10 locally on this commit. The changeset still correctly scopes a minor bump to `@revealui/security` (0.5.1) for the additive accessor and omits the apps. `isInMemoryStorage()` is unchanged from my previous review: read-only, exposes no handle, reachable only behind the authenticated-admin gate.

---

PR #2162, opened 2026-07-25
fix(db,auth,admin): audit env assert matches the real client contract (GAP-417 item 5)

RevealUIStudio:
Closes the fail-open the #2161 re-review proved empirically: `assertAuditStorageEnv` accepted `DATABASE_HOST`, which `getClient()` never consults — so the assert passed, `installAuditStorage()` threw into a swallow-and-continue catch, and production silently kept the in-memory audit sink.

**One predicate, one owner.** `@revealui/db` now exports `hasDatabaseConnectionEnv()`, extracted from `getClientByType()`'s own resolution (config url → `POSTGRES_URL` → `DATABASE_URL`) so the two can never drift; the assert consumes it. Red-first proven: the flipped `DATABASE_HOST`-must-throw test fails against the pre-fix dist (exactly 1 of 15), green after.

**Backstop for the class, not just the instance.** An install-construct failure on a production admin now refuses to serve (exit 1) instead of being swallowed — even if the predicate and client drift again someday, the failure is loud, never a silent in-memory sink.

**Forge kit plumbing (from the same review's residuals):**
- `docker-compose.forge.yml` never forwarded `REVEALUI_AUDIT_SIGNING_KEY` to either container, so an un-keyed kit restart-looped on a generic missing-env error. Now passed to api (`:?` require-fenced with a named message) and admin.
- Migrate-before-boot verified and recorded: the stamped kit runs a one-shot `migrate` service gated `service_completed_successfully` before api + admin (`apps/server/Dockerfile.forge` migrate stage) — the restart-loop concern is handled by design there.

Tests: server audit suites 17/17, admin instrumentation 11/11 (new production install-failure case), typechecks clean for db/auth/admin/server.

⚠️ Security-sensitive (`packages/auth/`): needs a recorded non-author guardrail-2 verdict + `sec-review:approved` before merge.

### DIFF
--- a/.changeset/gap-417-assert-getclient-contract.md
+++ b/.changeset/gap-417-assert-getclient-contract.md
@@ -1,0 +1,6 @@
+---
+'@revealui/db': minor
+'@revealui/auth': minor
+---
+
+Close the audit env-parity fail-open (GAP-417 item 5): `@revealui/db` now exports `hasDatabaseConnectionEnv()`, a boot-time predicate that matches `getClient()`'s connection resolution exactly (config url, then POSTGRES_URL / DATABASE_URL), and `assertAuditStorageEnv` consumes it instead of a local triple that also accepted `DATABASE_HOST` — an env shape `getClient()` never consults, which let the assert pass, the install throw, and production silently keep the in-memory audit sink.
--- a/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
+++ b/apps/admin/src/__tests__/instrumentation-audit-storage.test.ts
@@ -109,6 +109,20 @@ describe('installAdminAuditStorage (GAP-338)', () => {
     expect(written).not.toContain('non-production');
   });
 
+  it('production + install-construct failure: refuses to serve — assert/client drift must not fail open (GAP-417 item 5)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    installAuditStorage.mockImplementation(() => {
+      throw new Error('Database connection string not provided');
+    });
+
+    const { installAdminAuditStorage } = await import('../instrumentation-node');
+    await installAdminAuditStorage();
+
+    expect(exitSpy).toHaveBeenCalledWith(1);
+    const written = stderrSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('');
+    expect(written).toContain('AUDIT STORAGE INSTALL FAILED');
+  });
+
   it('Forge kit (RUNTIME_INIT): self-test BLOCKS and a failure refuses to serve (#2161 review)', async () => {
     vi.stubEnv('NODE_ENV', 'production');
     vi.stubEnv('RUNTIME_INIT', 'true');
--- a/apps/admin/src/app/api/health/__tests__/health-routes.test.ts
+++ b/apps/admin/src/app/api/health/__tests__/health-routes.test.ts
@@ -97,6 +97,49 @@ describe('GET /api/health', () => {
     expect(body.service).toBe('RevealUI Admin');
   });
 
+  it('non-production in-memory audit storage is informational: degraded check entry, still 200 (GAP-338/417)', async () => {
+    mockGetSession.mockResolvedValue({ user: { role: 'admin' } });
+    const mockFind = vi.fn().mockResolvedValue({ docs: [] });
+    mockGetRevealUIInstance.mockResolvedValue({ find: mockFind });
+
+    const { GET } = await loadRoute();
+    const res = await GET({ headers: { get: () => null } } as never);
+
+    // The test process runs on the real in-memory audit singleton — a
+    // by-design non-production state must not 503 the admin.
+    expect((res as { status: number }).status).toBe(200);
+    const body = (
+      res as unknown as {
+        body: { status: string; checks: Array<{ name: string; status: string }> };
+      }
+    ).body;
+    expect(body.status).toBe('healthy');
+    expect(body.checks.find((c) => c.name === 'audit-storage')?.status).toBe('degraded');
+  });
+
+  it('PRODUCTION in-memory audit storage is an outage: unhealthy check, 503 (GAP-338/417)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    try {
+      mockGetSession.mockResolvedValue({ user: { role: 'admin' } });
+      const mockFind = vi.fn().mockResolvedValue({ docs: [] });
+      mockGetRevealUIInstance.mockResolvedValue({ find: mockFind });
+
+      const { GET } = await loadRoute();
+      const res = await GET({ headers: { get: () => null } } as never);
+
+      expect((res as { status: number }).status).toBe(503);
+      const body = (
+        res as unknown as {
+          body: { status: string; checks: Array<{ name: string; status: string }> };
+        }
+      ).body;
+      expect(body.status).toBe('unhealthy');
+      expect(body.checks.find((c) => c.name === 'audit-storage')?.status).toBe('unhealthy');
+    } finally {
+      vi.unstubAllEnvs();
+    }
+  });
+
   it('returns 503 when database is unhealthy', async () => {
     mockGetSession.mockResolvedValue({ user: { role: 'admin' } });
     mockGetRevealUIInstance.mockRejectedValue(new Error('Connection refused'));
--- a/apps/admin/src/app/api/health/route.ts
+++ b/apps/admin/src/app/api/health/route.ts
@@ -116,7 +116,10 @@ export async function GET(request: Request) {
           'Audit storage is IN-MEMORY in production — admin audit emits evaporate on restart',
       });
     } else {
-      overallStatus = overallStatus === 'healthy' ? 'degraded' : overallStatus;
+      // By-design state (dev shell / test env without a DB): informational
+      // check entry only. It must NOT downgrade overallStatus — this route
+      // maps every non-healthy overall to 503, and a deliberate dev default
+      // is not an outage (the 503 belongs to production in-memory only).
       checks.push({
         name: 'audit-storage',
         status: 'degraded',
@@ -124,7 +127,9 @@ export async function GET(request: Request) {
       });
     }
   } catch (error) {
-    overallStatus = overallStatus === 'healthy' ? 'degraded' : overallStatus;
+    // Probe failure is informational too — the audit path itself is guarded
+    // at boot (instrumentation) and at write time; the health probe must not
+    // 503 the whole admin because the check could not run.
     checks.push({
       name: 'audit-storage',
       status: 'degraded',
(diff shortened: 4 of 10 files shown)


RevealUIStudio:

This closes the residual I recorded on the #2161 approval, and it closes it at the right layer: the predicate now lives in the package that owns the client, rather than being a second guess at what that client accepts. I verified every claim against the branch and re-ran my own acceptance probe against the rebuilt dists. Suites green: admin instrumentation 11/11, server audit-storage 15/15, db pool + client 32/32.

---

### 1. One resolution, no drift: VERIFIED

[client/index.ts:279](packages/db/src/client/index.ts#L279) extracts `resolveDatabaseUrl()`, and both consumers call it: `hasDatabaseConnectionEnv()` at [L301](packages/db/src/client/index.ts#L301) and `getClientByType()` at [L305](packages/db/src/client/index.ts#L305). There is no second copy of the resolution anywhere in the file, so the predicate cannot answer a question the client would answer differently. Putting it in `@revealui/db` rather than re-deriving it in `@revealui/auth` is the correct ownership call, and the "keep this and `getClientByType` on the same resolution, always" comment is the right thing to leave behind.

Refactor fidelity checked line by line. Config-first order preserved, the `||` chain still catches empty strings, and the thrown error at [L311](packages/db/src/client/index.ts#L311) is byte-identical. The single behavioral delta is the trailing `|| undefined`, which normalizes `''` to `undefined`; both values are falsy and hit the same `if (!url || typeof url !== 'string')` guard, so `getClientByType` behavior is unchanged. The db suites confirm it: 32/32 across pool and client, including the pool path's own `DATABASE_HOST` cases, which are untouched.

### 2. The fail-open is closed: VERIFIED EMPIRICALLY, before and after

I have direct before/after evidence from my own probe rather than a red-first claim to take on trust. Same script, same env, two branches.

Against the #2161 dist (the state I approved with this as a follow-up):

```
assertAuditStorageEnv(): PASSED (no parity failure -> no exit(1) path)
installAuditStorage(): THREW: Database connection string not provided for REST database.
final isInMemoryStorage = true
```

Against this branch's rebuilt dist, `NODE_ENV=production`, `DATABASE_HOST=db.internal`, no URL vars:

```
assertAuditStorageEnv(): THREW: AUDIT STORAGE ENV PARITY FAILED: no usable database
connection is configured (set POSTGRES_URL or DATABASE_URL, or provide
@revealui/config database.url) ...
```

Positive control, so this is not just a blanket tightening. With `DATABASE_URL` present:

```
assertAuditStorageEnv(): PASSED
installAuditStorage(): did NOT throw
final isInMemoryStorage = false
```

The assert now accepts exactly what the client accepts, in both directions. [audit-storage.ts:256](packages/auth/src/server/audit-storage.ts#L256) consumes the predicate and the error message was rewritten to name the env shapes that actually work, which matters because the old message actively misled an operator by listing `DATABASE_HOST` as sufficient. The replacement test at [audit-storage.pglite.test.ts:175-183](apps/server/src/lib/__tests__/audit-storage.pglite.test.ts#L175-L183) pins the new contract and carries the reasoning.

### 3. Admin backstop: VERIFIED

[instrumentation-node.ts:83-97](apps/admin/src/instrumentation-node.ts#L83-L97) wraps the install, exits 1 in production at [L92-93](apps/admin/src/instrumentation-node.ts#L92-L93), and warns plus returns otherwise. This is defense in depth in the correct sense: even if the predicate and the client contract drift again in the future, the outcome is refuse-to-serve rather than a silent in-memory sink. The never-throw contract holds, the branch returns rather than rethrowing, and the whole body remains inside the outer catch. The Forge/serverless split from #2161 is untouched below it at [L101](apps/admin/src/instrumentation-node.ts#L101). Pinned by the new test at [test:112-124](apps/admin/src/__tests__/instrumentation-audit-storage.test.ts#L112-L124).

### 4. Forge plumbing, and my fresh-kit residual: ANSWERED, with one part I cannot verify here

The env plumbing is real: api is require-fenced at [docker-compose.forge.yml:83](docker-compose.forge.yml#L83), admin receives the same key at [L152](docker-compose.forge.yml#L152), and the header documents it. I checked the justification rather than accepting it: `REVEALUI_AUDIT_SIGNING_KEY` is already hard-required and PKCS#8-parsed by [validate-startup.ts:157-180](apps/server/src/lib/validate-startup.ts#L157-L180) and listed in [required-env.ts:69](apps/server/src/lib/required-env.ts#L69), so the compose was previously guaranteeing a restart loop with a generic error. Fencing it converts that into a named failure at interpolation time. Correct change.

On the migrate-before-boot answer, the honest split:

- **For the in-repo compose the concern is moot, for a reason better than a migrate service.** `docker-compose.forge.yml` sets no `RUNTIME_INIT` anywhere, so the blocking Forge self-test at [instrumentation-node.ts:101](apps/admin/src/instrumentation-node.ts#L101) never fires there, and `initEngineAtBoot` no-ops for the same reason. The boot-loop path I worried about is unreachable in that compose whether or not migrations have run. That fully retires the residual for this file.
- **For the stamped kit I can confirm the mechanism exists but not that it is wired.** [apps/server/Dockerfile.forge:54-70](apps/server/Dockerfile.forge#L54-L70) defines the one-shot migrate stage and documents `service_completed_successfully` gating before api and admin. But no file in this repository contains `service_completed_successfully`, so the kit compose that actually wires it lives outside this repo and is not reviewable here. I accept the answer as plausible and consistent, and I am flagging that the verification for the kit half rests on a comment rather than on code I can read.

### 5. Sweep: CLEAN

Nothing legitimately relies on the old semantics. After this change `DATABASE_HOST` survives only in [packages/db/src/pool.ts](packages/db/src/pool.ts#L39-L65), which is a separate connection path with its own documented accept-list and its own tests (32/32 green, unchanged), and in the audit tests and comments that describe the fix. No `.env.example`, deployment config, doc, or Vercel/compose surface offers `DATABASE_HOST` as a database connection for the client path. The in-repo forge compose supplies both `POSTGRES_URL` and `DATABASE_URL`.

Worth stating plainly since it is the one behavior change with operator reach: a deployment configured with `DATABASE_HOST` alone now fails closed at boot on api and admin where it previously degraded silently on admin. That is the intended fix and no such deployment appears to exist, but it is a boot-behavior change, not purely a bug fix.

Changeset scoping is correct: `@revealui/db` minor for the additive export, `@revealui/auth` minor for a behavior tightening inside 0.x, which is the right SemVer signal under the 0.x rule.

---

### Suggestions

- **`hasDatabaseConnectionEnv(env)` is not a pure function of `env`.** The config branch at [client/index.ts:281-289](packages/db/src/client/index.ts#L281-L289) always reads the process-global `@revealui/config`, so only the env-var fallback is parameterized. The JSDoc says so, and in production it is correct because callers pass the real `process.env`. The side effect is on tests: `expect(() => assertAuditStorageEnv({})).toThrow()` at [audit-storage.pglite.test.ts:161](apps/server/src/lib/__tests__/audit-storage.pglite.test.ts#L161) now passes only because `@revealui/config` does not resolve a URL in that process. If it ever does, that assertion silently inverts rather than failing loudly. Consider threading `env` into the config branch, or naming the function so the process-global read is visible at the call site.
- **The compose change deserves an operator note.** `${REVEALUI_AUDIT_SIGNING_KEY:?...}` means an existing kit operator who pulls the new compose without setting the var gets an immediate hard failure on `docker compose up`. That is the right failure, but #2156 carried an explicit deployment note for exactly this class of change and this PR body does not. Worth adding before promotion so it reaches release notes.
- **Fence the admin var too.** api uses `:?` at [L83](docker-compose.forge.yml#L83) while admin uses a bare `${REVEALUI_AUDIT_SIGNING_KEY}` at [L152](docker-compose.forge.yml#L152). Safe today because compose interpolates the whole file and api's fence trips first, but that safety is incidental and would evaporate if the services are ever split across compose files. Cheap defense in depth.
- The changeset frames this purely as closing a fail-open. One clause noting the `DATABASE_HOST`-only boot change would make the operator-visible effect discoverable from the release notes alone.

### Nits

- The orphaned GAP-214 JSDoc block above `installAdminAuditStorage` is still outstanding from my first review. Cosmetic, and this PR is not the place for it, but it has now survived three commits to this file.

---

RevealUIStudio:
CI Unit Tests failure fixed in the new commit: the #2161 audit-storage health check let the by-design non-production in-memory state downgrade overallStatus, and this route maps any non-healthy overall to 503 — which is what the admin health unit test (real in-memory singleton under NODE_ENV=test) caught. Non-production in-memory and probe failures are now informational check entries only; production in-memory keeps unhealthy/503. Two new unit cases pin both semantics (health suite 11/11). Note for the standing verdict: this is a semantics correction to the #2161 health check carried on this branch, not a change to the predicate/backstop under review.

RevealUIStudio:

Delta verdict covering `7d88955c5`. My APPROVE now covers the final tree.

**Scope claim verified.** `git diff a226bb33f..7d88955c5 --stat` touches exactly two files, both `apps/admin/src/app/api/health/`. The `packages/auth` and `packages/db` surface I approved is byte-untouched.

**Finding 3 is not weakened.** The cross-bundle duplication scenario I relied on is a *production* scenario: a duplicated module means the route bundle resolves its own untouched default sink, `isInMemoryStorage()` returns true, and the production branch at [route.ts:110-117](apps/admin/src/app/api/health/route.ts#L110-L117) is unchanged — `overallStatus = 'unhealthy'`, and [L155](apps/admin/src/app/api/health/route.ts#L155) still maps that to 503. The commit only stops *by-design* non-production state from borrowing that status code, and the `audit-storage` entry still appears in the `checks` array either way. Detection is intact.

The delta in fact **strengthens** the finding-3 evidence. On #2161 I noted the health check had no test at all; [health-routes.test.ts:117-140](apps/admin/src/app/api/health/__tests__/health-routes.test.ts#L117-L140) is now the first executable proof of the production path, stubbing `NODE_ENV=production` and asserting both 503 and the `unhealthy` entry. The non-production case at [L100-116](apps/admin/src/app/api/health/__tests__/health-routes.test.ts#L100-L116) runs against the real in-memory singleton and pins 200 plus the surviving degraded entry. Suite 11/11 locally.

The fix itself is correct on its own terms: a deliberate dev default is not an outage, and a route that 503s its own health endpoint on a DB-less dev shell is a real defect.

**One narrow thing the delta does give up.** The `catch` arm at [L127-134](apps/admin/src/app/api/health/route.ts#L127-L134) is now informational in production too, so a probe failure there (the dynamic import of `@revealui/security/server` failing from the route bundle) returns 200 where it previously returned 503. That is a different failure mode from duplication, so it does not touch what finding 3 rests on, and the boot-time and write-time rails still guard the path. But it is the one anomalous production state that lost its status-code signal, and the refinement is one line: keep the production sub-case in that `catch` as unhealthy while leaving non-production informational. Not blocking.

The suggestions from my full verdict on `a226bb33f` still stand unchanged.

PR #2164, opened 2026-07-25
feat(auth,db,security,admin): GAP-417 both rails — no escape hatch into unsigned production audit rows

RevealUIStudio:
Implements the GAP-417 items 1-2 doctrine as **owner-countersigned in-session 2026-07-25 ("both rails")**, plus the owner-ruled health signal:

**Rail 1 (boot).** `SKIP_ENV_VALIDATION` no longer exempts the audit signing key in production: `assertAuditStorageEnv` refuses regardless. The audit path has no escape hatch — a production process that cannot sign must not boot, because unsigned rows are indistinguishable from tampering and permanently stall anchor contiguity once the sweep filters them. SKIP keeps covering non-audit env validation.

**Rail 2 (store).** `DrizzleAuditStore.append`/`appendBatch` refuse to land an UNSIGNED row when `NODE_ENV=production` — the backstop for any path that constructs a signer-less store however it booted. Dev/test unsigned behavior is unchanged.

**env_file PEM transport.** `createAuditRowSignerFromEnv` and `resolveAuditPublicKey` normalize single-line `\n`-escaped PEMs (Docker env_file is one line per key; mirrors the license normalizePem precedent, split/join, no authored regex). Prerequisite for the RevForge kit provisioning the audit key at all (follow-up PR in the revforge repo).

**Health signal (owner-ruled: reflect status code).** The unauthenticated `/api/health` arm now reflects the real overall status CODE via a zero-I/O audit probe — bare `{status}` body either way (public-issue-redaction posture), no DB or Stripe work for anonymous callers — so uptime monitors and the kit healthcheck finally see a production audit outage. Plus the #2162 review refinement: a production probe FAILURE in the authenticated arm is unhealthy/503, not informational; non-production stays informational (the #2161-era dev-shell 503 does not come back).

**Red-first, run against base dists:** exactly the two new rails failed (SKIP-exemption flip: "expected [Function] to throw"; store rail: "promise resolved instead of rejecting") with everything else green — then green after the fix (server audit suites 26/26, security 10/10, admin health + instrumentation 24/24, auth 531, typechecks clean across db/security/auth/admin/server).

⚠️ Security-sensitive (`packages/auth/`, `packages/security/`): needs a recorded non-author guardrail-2 verdict + `sec-review:approved` before merge. Owner countersign for the doctrine itself is on record in-session (2026-07-25).

### DIFF
--- a/.changeset/gap-417-both-rails.md
+++ b/.changeset/gap-417-both-rails.md
@@ -1,0 +1,7 @@
+---
+'@revealui/auth': minor
+'@revealui/db': minor
+'@revealui/security': minor
+---
+
+GAP-417 both rails (owner-countersigned): the audit path has no escape hatch into unsigned production rows. `assertAuditStorageEnv` refuses a production boot without the signing key regardless of `SKIP_ENV_VALIDATION`; `DrizzleAuditStore.append`/`appendBatch` refuse to land an unsigned row when `NODE_ENV=production`; and the signer composition normalizes single-line `\n`-escaped PEMs so `env_file` transports (RevForge kits) can carry the key.
--- a/apps/admin/src/app/api/health/__tests__/health-routes.test.ts
+++ b/apps/admin/src/app/api/health/__tests__/health-routes.test.ts
@@ -73,6 +73,53 @@ describe('GET /api/health', () => {
     );
   });
 
+  it('unauthenticated + PRODUCTION in-memory audit storage: bare 503, no detail, no I/O (GAP-417)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    try {
+      mockGetSession.mockResolvedValue(null);
+      const { GET } = await loadRoute();
+      const res = await GET({ headers: { get: () => null } } as never);
+
+      // The test process runs on the real in-memory audit singleton, so the
+      // production zero-I/O probe fires. Monitors see the outage; the body
+      // stays a bare status (public-issue-redaction posture) and the engine
+      // is never touched for an anonymous caller.
+      expect((res as { status: number }).status).toBe(503);
+      const body = (res as unknown as { body: Record<string, unknown> }).body;
+      expect(body).toEqual({ status: 'unhealthy' });
+      expect(mockGetRevealUIInstance).not.toHaveBeenCalled();
+    } finally {
+      vi.unstubAllEnvs();
+    }
+  });
+
+  it('authenticated admin + PRODUCTION probe failure: unhealthy 503 (#2162 review refinement)', async () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    const { audit } = await import('@revealui/security/server');
+    const probeSpy = vi.spyOn(audit, 'isInMemoryStorage').mockImplementation(() => {
+      throw new Error('probe boom');
+    });
+    try {
+      mockGetSession.mockResolvedValue({ user: { role: 'admin' } });
+      const mockFind = vi.fn().mockResolvedValue({ docs: [] });
+      mockGetRevealUIInstance.mockResolvedValue({ find: mockFind });
+
+      const { GET } = await loadRoute();
+      const res = await GET({ headers: { get: () => null } } as never);
+
+      expect((res as { status: number }).status).toBe(503);
+      const body = (
+        res as unknown as {
+          body: { status: string; checks: Array<{ name: string; status: string }> };
+        }
+      ).body;
+      expect(body.checks.find((c) => c.name === 'audit-storage')?.status).toBe('unhealthy');
+    } finally {
+      probeSpy.mockRestore();
+      vi.unstubAllEnvs();
+    }
+  });
+
   it('returns minimal status for non-admin authenticated users', async () => {
     mockGetSession.mockResolvedValue({ user: { role: 'user' } });
     const { GET } = await loadRoute();
--- a/apps/admin/src/app/api/health/route.ts
+++ b/apps/admin/src/app/api/health/route.ts
@@ -25,6 +25,21 @@ export async function GET(request: Request) {
   const isAuthenticated = session?.user?.role === 'admin';
 
   if (!isAuthenticated) {
+    // GAP-417 (owner-ruled 2026-07-25): the unauthenticated arm reflects the
+    // real overall status CODE so uptime monitors and the kit healthcheck
+    // finally see a production audit outage, while leaking zero check detail
+    // (public-issue-redaction posture: a bare {status} body either way).
+    // Only zero-I/O probes run for anonymous callers — no DB query, no
+    // Stripe call, no new anonymous load surface.
+    try {
+      const { audit } = await import('@revealui/security/server');
+      if (process.env.NODE_ENV === 'production' && audit.isInMemoryStorage()) {
+        return NextResponse.json({ status: 'unhealthy' }, { status: 503 });
+      }
+    } catch {
+      // Probe unavailable: stay minimal-healthy — the boot rails and the
+      // authenticated arm carry the detailed signal.
+    }
     return NextResponse.json({ status: 'healthy' });
   }
 
@@ -127,14 +142,26 @@ export async function GET(request: Request) {
       });
     }
   } catch (error) {
-    // Probe failure is informational too — the audit path itself is guarded
-    // at boot (instrumentation) and at write time; the health probe must not
-    // 503 the whole admin because the check could not run.
-    checks.push({
-      name: 'audit-storage',
-      status: 'degraded',
-      message: error instanceof Error ? error.message : 'Audit storage check failed',
-    });
+    if (process.env.NODE_ENV === 'production') {
+      // #2162 review refinement: a probe that cannot even run in PRODUCTION
+      // is an anomalous state that must keep its status-code signal — it is
+      // the one arm that would otherwise report 200 while the audit surface
+      // is unobservable.
+      overallStatus = 'unhealthy';
+      checks.push({
+        name: 'audit-storage',
+        status: 'unhealthy',
+        message: error instanceof Error ? error.message : 'Audit storage check failed',
+      });
+    } else {
+      // Non-production probe failure stays informational — the audit path is
+      // guarded at boot and at write time; a dev shell must not 503 on it.
+      checks.push({
+        name: 'audit-storage',
+        status: 'degraded',
+        message: error instanceof Error ? error.message : 'Audit storage check failed',
+      });
+    }
   }
 
   // System metrics
--- a/apps/server/src/lib/__tests__/audit-storage.pglite.test.ts
+++ b/apps/server/src/lib/__tests__/audit-storage.pglite.test.ts
@@ -21,7 +21,7 @@ import type { AuditEvent, AuditStorage } from '@revealui/core/security';
 import { AuditSystem } from '@revealui/core/security';
 import type { Database } from '@revealui/db/client';
 import { eq } from 'drizzle-orm';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   createTestDb,
   type TestDb,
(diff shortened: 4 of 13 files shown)


RevealUIStudio:

Both rails are correctly implemented and I verified them independently. One blocker, in the third change: the PEM normalization does not reach the places that would make its stated purpose work, and I proved it. Suites green throughout: server audit 24/24, security signer-env 10/10, admin health 13/13.

---

### Rail 1 (no SKIP exemption): VERIFIED

[audit-storage.ts:277](packages/auth/src/server/audit-storage.ts#L277) is now a bare `if (env.NODE_ENV === 'production')`. I grepped the whole audit surface (`packages/auth/src`, `packages/db/src`, `packages/security/src`, `instrumentation-node.ts`) for `SKIP_ENV_VALIDATION`: every remaining occurrence is prose in a comment explaining why the carve-out is gone. No live consultation survives anywhere on the audit path. Dev/test stays exempt because the whole block is production-gated, so an unsigned dev shell still boots.

### Rail 2 (store-level refusal): VERIFIED, and the regression sweep is clean

`refuseUnsignedInProduction()` at [audit-store.ts:116](packages/db/src/audit-store.ts#L116) is called from both signer-less arms, `append` at [L130](packages/db/src/audit-store.ts#L130) and `appendBatch` at [L158](packages/db/src/audit-store.ts#L158). Placing it inside the `if (!this.signer)` branch is right: a signed store pays nothing.

I swept every non-test `new DrizzleAuditStore` construction in the tree, since this was the highest-risk class. Three exist, and none breaks:

- [apps/admin/src/app/api/setup/route.ts:117](apps/admin/src/app/api/setup/route.ts#L117) composes a signer from env and the append sits inside an explicitly non-fatal try/catch that routes through `classifyAuditWriteFailure` and `recordAuditWriteResult`. A refusal is caught and recorded, not a 500.
- [scripts/admin/bootstrap.ts:176](scripts/admin/bootstrap.ts#L176) and [scripts/admin/grant-entitlement.ts:249](scripts/admin/grant-entitlement.ts#L249) follow the same pattern with the same wrapping.

So no caller crashes. The behavior change is that an unsigned production write is now dropped and recorded as a failure rather than landed, which is the intended doctrine.

### PEM normalization: correct in itself, and I confirmed no regression

[audit-signer-env.ts:53](packages/security/src/audit-signer-env.ts#L53) uses `includes('\\n')` then split/join. A real multi-line PEM contains no backslash and the base64 alphabet cannot produce one, so multi-line input returns byte-identical and untouched. No regex, no eval, and the output feeds `createPrivateKey()` which parses strictly, so there is no injection surface: a malformed value throws rather than doing anything interesting. The cross-check arm normalizes both sides before the DER comparison, so it stays apples-to-apples. Proven directly, single-line and multi-line compose the **same kid**:

```
[multi-line]  createAuditRowSignerFromEnv : mode=signed kid=ed25519-ed635fee9bdd9100
[single-line] createAuditRowSignerFromEnv : mode=signed kid=ed25519-ed635fee9bdd9100
```

### Health: VERIFIED on every point asked

The unauthenticated arm at [route.ts:29-43](apps/admin/src/app/api/health/route.ts#L29-L43) runs a dynamic import plus a synchronous `instanceof`. No engine call, no DB query, no Stripe call, so the zero-I/O claim holds and no new anonymous load surface is created. The body stays bare `{ status }` in both directions, so the redaction posture is intact, and the 503 is gated on `NODE_ENV === 'production'` at [L36](apps/admin/src/app/api/health/route.ts#L36). The dev-shell 503 regression from #2161 does not return: a non-production anonymous caller falls through to a healthy 200, and the authenticated non-production path is still informational-only. The authed catch arm now carries the production-unhealthy refinement I asked for on #2162, exactly as suggested.

---

### Blocker

**The PEM normalization is applied at one read site, and the two that decide whether a RevForge kit boots and whether an operator can verify receipts still read the raw value. The changeset's stated purpose is not achieved.**

The changeset says the normalization exists "so `env_file` transports (RevForge kits) can carry the key". On such a kit the api container runs `validateStartup()`, which reads the raw env at [validate-startup.ts:313](apps/server/src/lib/validate-startup.ts#L313) and hands it unnormalized to `classifyEd25519PrivateKey` at [L166](apps/server/src/lib/validate-startup.ts#L166), where `createPrivateKey(pem)` throws. Same script, same key, three call sites:

```
[multi-line]
  createAuditRowSignerFromEnv : mode=signed kid=ed25519-ed635fee9bdd9100
  validate-startup parse (raw): OK (ed25519)
[single-line]
  createAuditRowSignerFromEnv : mode=signed kid=ed25519-ed635fee9bdd9100
  validate-startup parse (raw): THREW: could not be parsed as a PKCS#8 PEM Ed25519 private key
```

So a kit operator who supplies the key by `env_file` gets a hard boot refusal from the api with "could not be parsed as a PKCS#8 PEM Ed25519 private key", before the newly-tolerant signer is ever reached. This PR makes that worse in combination with rail 1, which correctly removes the escape hatch: there is now no way to boot past it, and the error message points at the key being malformed when it is in fact the shape the changeset says is supported.

The same omission hits the offline verifiers, which are the operator-facing proof of the whole receipts positioning. [verify-audit-anchor.ts:112](scripts/security/verify-audit-anchor.ts#L112) and [verify-audit-signatures.ts:75](scripts/security/verify-audit-signatures.ts#L75) both call `createPrivateKey(privatePem)` on `process.env.REVEALUI_AUDIT_SIGNING_KEY` directly, so an operator whose environment carries a single-line PEM cannot verify their own audit log either.

Fix, and it is the lesson #2162 already taught on this branch's own history: normalize in one place and have every reader use it. Export `normalizeEnvPem` from `@revealui/security` (it is currently module-private) and consume it at the `validate-startup` read and in both verifier scripts. Centralizing is what stopped the predicate/client drift in #2162; normalizing in one of four readers reintroduces exactly that drift class in a new place. A test that runs a single-line PEM through `classifyEd25519PrivateKey` would pin it.

### Suggestions

- **Rail 2 keys on `NODE_ENV`, not on which database it is pointed at.** The CLI writers are the callers most likely to run against production with `NODE_ENV` unset, and [bootstrap.ts:41](scripts/admin/bootstrap.ts#L41) defaults it to `'development'`. An operator running `admin:bootstrap` or `grant-entitlement` against the production database from a shell without the signing key still lands an unsigned row in the production `audit_log`, which the anchor sweep filters and which stalls contiguity, the precise harm the rail exists to prevent. Not introduced by this PR and the rail is honest about its keying in the JSDoc, but it is the remaining hole in the doctrine and worth carrying on GAP-417 rather than considering the item closed.
- **The anonymous arm now signals inconsistently.** Audit-storage in-memory is the only condition that reaches an anonymous caller's status code; a total database outage still returns a healthy 200 to the same caller, because `overallStatus` is only computed in the authenticated arm. A monitor wired to this endpoint will report an audit misconfiguration as an outage and a real DB outage as fine. Given the owner ruling was specifically about the audit signal I am not treating it as a defect, but the asymmetry will confuse whoever wires the kit healthcheck to it, and it does disclose one specific internal state bit to unauthenticated callers.

### Nits

- `normalizeEnvPem`'s `includes('\\n')` guard is redundant, since split/join on an absent separator is already a no-op returning an equal string. Harmless, and it does document intent.
- The orphaned GAP-214 JSDoc block above `installAdminAuditStorage` is still outstanding.

Changeset scoping is correct: auth, db and security all minor, which is the right 0.x signal for three behavior tightenings.

---

RevealUIStudio:

Delta verdict covering `83567eaa3`. The blocker is closed, fixed at the layer I asked for, and I proved it directly rather than reading the diff and assuming.

**Note on my own prior evidence.** The probe in my REQUEST-CHANGES simulated the *raw* read path, so re-running it unchanged still shows a single-line failure by construction. I rewrote it to exercise the **patched** chain. Same key, both shapes:

```
normalizeEnvPem on @revealui/security/server : function
normalizeEnvPem on @revealui/core/security   : function
[multi-line]  signer=signed/ed25519-b80427189ccbbfbb | validate-startup(normalized)=OK (ed25519) | pub=ed25519-b80427189ccbbfbb
[single-line] signer=signed/ed25519-b80427189ccbbfbb | validate-startup(normalized)=OK (ed25519) | pub=ed25519-b80427189ccbbfbb
explicit single-line PUBLIC key normalizes  : true
```

All three call sites now agree on the same kid, the format check accepts the escaped key, and the verifier's explicit public-key arm normalizes too. The `env_file` transport the changeset promised now actually works end to end.

**Reachability verified, not assumed.** `validate-startup.ts` imports from `@revealui/core/security` rather than `@revealui/security/server`, so I confirmed the symbol resolves through that barrel as well as directly, and that both appear in the rebuilt dists.

**I re-swept for a fourth unnormalized reader and found none.** Every remaining occurrence of the two env vars is prose, a presence-only check that never parses ([audit-storage.ts:281](packages/auth/src/server/audit-storage.ts#L281)), or one of the patched lines. Worth recording that the anchor sweep was never exposed: [audit-anchor-sweep.ts:124](apps/server/src/jobs/audit-anchor-sweep.ts#L124) composes through the shared `createAuditRowSignerFromEnv`, which normalizes internally, so the "one reader of four" count was accurate.

**The new test is the right kind.** [validate-startup.test.ts:160-167](apps/server/src/lib/__tests__/validate-startup.test.ts#L160-L167) drives the real `validateStartup` with an escaped key and asserts it does not throw, rather than unit-testing the normalizer in isolation. That is what pins the boot behavior. 118/118 green locally.

The export comment forbidding raw reads of these vars is the durable half of the fix, and citing the cross-call-site proof in it is exactly how this class of drift stays dead.

---

### On the carried residual: recording suffices, and it should not block this PR

Rail 2 keys on `NODE_ENV`, so an operator running `admin:bootstrap` or `grant-entitlement` against the production database from a shell without the signing key can still land an unsigned row. I am comfortable calling this recorded-and-deferred rather than blocking, for three reasons: this PR does not introduce or widen it, it strictly narrows the surface by closing every server-process path, and the right keying is a genuine design question rather than an oversight. Keying the rail on the connection target instead of `NODE_ENV` changes CLI ergonomics fleet-wide and deserves the same countersign the doctrine itself got.

One concrete note for the GAP-417 entry, since it makes the fix cheaper than it looks: the CLI scripts already know which environment they are targeting. [bootstrap.ts:41](scripts/admin/bootstrap.ts#L41) computes `nodeEnv` and the audit payload carries an `env` field, so the durable fix can pass that resolved target into the refusal rather than reading `process.env.NODE_ENV` inside the store. That keeps the store's rail honest without making every local CLI invocation demand a production key.

Everything from my full verdict on `94d17145e` that I marked verified still stands: rail 1 with no surviving SKIP consultation, rail 2 with a clean caller sweep, and the health route on all four points.

RevealUIStudio:

Adversarial non-author review of head 7d88955 (GAP-417 items 1-2, owner-countersigned doctrine). Four checks, all pass.

### 1. Rail 1 — boot assert: PASS

`assertAuditStorageEnv` drops the `SKIP_ENV_VALIDATION` exemption for the signing key in production; the test that previously asserted not-throw under SKIP now asserts the throw, which is the correct prove-the-flip shape. SKIP's coverage of non-audit env validation is untouched. The error message names the doctrine so an operator hitting it in a kit knows SKIP will not help.

### 2. Rail 2 — store backstop: PASS

`DrizzleAuditStore.append` AND `appendBatch` both refuse a signer-less write when `NODE_ENV=production`, PGlite-proven with a rows-are-empty assertion after both rejections; dev/test unsigned behavior explicitly re-proven unchanged. The two rails compose correctly: boot refuses a keyless production process, and the store catches any residual path that constructs a signer-less store however it booted. Legitimate production writers all construct through `createAuditStore` (signer injected), so the backstop only fires on genuine misconfiguration.

### 3. env_file PEM normalization — coverage is complete: PASS

`normalizeEnvPem` is applied at every parse site of the key envs: signer composition, public-key resolution (explicit + derived), validate-startup's format check, and both offline verifiers. Grepped the head for any other reader of `REVEALUI_AUDIT_SIGNING_KEY` / `REVEALUI_AUDIT_PUBLIC_KEY`: the remaining hits are log strings, docs, env lists, and secret-path configs — no un-normalized parser exists; runtime consumers (setup route, anchor sweep, bootstrap/grant scripts) all route through the one composition point. The escape transform is safe: a valid PEM can never contain a literal backslash-n sequence, so multi-line keys pass through byte-identical, and the same-kid test proves escaped and multi-line keys compose the same signer.

### 4. Health signal: PASS, one non-blocking note

The unauthenticated arm reflects the real status code via a zero-I/O sync probe (`audit.isInMemoryStorage()`, present on the singleton since #2161) with a bare `{status}` body both ways — the public-issue-redaction posture holds, and no DB/Stripe work is added for anonymous callers. The authenticated arm's production probe-failure now correctly forces unhealthy (the one arm that could otherwise 200 while the audit surface is unobservable). NOTE (non-blocking): on the unauthenticated arm a probe-import failure falls back to healthy-200; acceptable per the in-code rationale (boot rails + authenticated arm carry the detailed signal), but worth remembering if monitors ever seem quiet during a partial outage.

### CI

The Server tsx boot failure on the first run was a Docker Hub registry pull timeout (`pgvector/pgvector:pg16`, three retries at container init — never reached the code); green on re-run. Unit Tests green. Only this gate holds.

PR #2167, opened 2026-07-25
chore(harnesses): de-regex yaml escaping, document normalizer posture

RevealUIStudio:
## Summary

Two KEEP-hygiene items from the GAP-421 harnesses routing audit (step 3, no owner gate):

- `packages/harnesses/src/content/generators/claude-code.ts:22` carried the package's only unmarked regex (a YAML-special-char class), inside the very module that authors the fleet's no-regex rule content. Replaced with a `Set`-based character check (the `no-regex.md` "character class checks -> explicit checks" alternative). Output is unchanged -- the committed content snapshot check still passes.
- `packages/harnesses/src/hooks/normalizers/cursor.ts` and `vscode.ts` have zero live callers today: no Cursor or VS Code harness home invokes `revealui-harnesses hook` yet. This is intentional under GAP-293 harness parity, but the audit found it was only inferable from context, not stated in-file -- added an explicit note to each docblock so a later dead-code sweep doesn't misread them.

No behavior change; no changeset (internal-only, no consumer-visible surface changed).

Routing ratified by owner directive 2026-07-25 (GAP-421). Steps 6 and 8 of the same audit (daemon-ownership ADR + the DELETE train that depends on it) are explicitly out of scope for this PR.

## Test plan

- [x] `pnpm --filter @revealui/harnesses build` -- clean
- [x] `pnpm --filter @revealui/harnesses test` -- 41 files / 542 tests pass (content snapshot test included)
- [x] `pnpm --filter @revealui/harnesses typecheck` -- clean
- [x] `pnpm gate:quick` -- PASS

### DIFF
--- a/packages/harnesses/src/content/generators/claude-code.ts
+++ b/packages/harnesses/src/content/generators/claude-code.ts
@@ -18,8 +18,40 @@ import { MANAGER_CONTENT_OUTPUT } from './types.js';
 
 const CONTENT = MANAGER_CONTENT_OUTPUT;
 
+// Characters that force a YAML scalar to be quoted (fleet no-regex hardline:
+// Set membership over a character class, per .claude/rules/no-regex.md).
+const YAML_QUOTING_CHARS = new Set([
+  ':',
+  '#',
+  '\n',
+  '"',
+  "'",
+  '{',
+  '}',
+  '[',
+  ']',
+  ',',
+  '&',
+  '*',
+  '?',
+  '|',
+  '>',
+  '!',
+  '%',
+  '@',
+  '`',
+]);
+
+function needsYamlQuoting(value: string): boolean {
+  if (value.trim() !== value) return true;
+  for (const char of value) {
+    if (YAML_QUOTING_CHARS.has(char)) return true;
+  }
+  return false;
+}
+
 function yamlEscape(value: string): string {
-  if (/[:#\n"'{}[\],&*?|>!%@`]/.test(value) || value.trim() !== value) {
+  if (needsYamlQuoting(value)) {
     return JSON.stringify(value);
   }
   return value;
--- a/packages/harnesses/src/hooks/normalizers/cursor.ts
+++ b/packages/harnesses/src/hooks/normalizers/cursor.ts
@@ -15,6 +15,14 @@
  *
  * `user_email` is read ONLY to prove the trust boundary: it is never copied
  * onto `HarnessHookIdentity` (design invariant I-1). It stays in `raw`.
+ *
+ * INTENTIONALLY UNREACHED TODAY (GAP-293 harness parity, GAP-421 §4.2): no
+ * Cursor harness home invokes `revealui-harnesses hook` yet, so this
+ * normalizer has zero live callers. It exists so `runHookCommand` /
+ * `normalizeHookEvent` (`../run-hook.ts`) already dispatch on
+ * `HarnessHookSource` for every editor the harness protocol targets, ahead of
+ * Cursor actually wiring the hook. Not dead code — do not delete without
+ * re-checking GAP-293's wiring status first.
  */
 
 import type {
--- a/packages/harnesses/src/hooks/normalizers/vscode.ts
+++ b/packages/harnesses/src/hooks/normalizers/vscode.ts
@@ -33,6 +33,14 @@
  *    built-in Copilot agent-mode tool set (code.visualstudio.com/docs/copilot/agents/agent-tools),
  *    not from the hooks reference itself -- best-effort, same posture as
  *    Cursor's `beforeReadFile` -> synthetic `Read` toolName fallback.
+ *
+ * INTENTIONALLY UNREACHED TODAY (GAP-293 harness parity, GAP-421 §4.2): no
+ * VS Code harness home invokes `revealui-harnesses hook` yet, so this
+ * normalizer has zero live callers. It exists so `runHookCommand` /
+ * `normalizeHookEvent` (`../run-hook.ts`) already dispatch on
+ * `HarnessHookSource` for every editor the harness protocol targets, ahead of
+ * VS Code actually wiring the hook. Not dead code -- do not delete without
+ * re-checking GAP-293's wiring status first.
  */
 
 import type {


RevealUIStudio:

Non-author review of head dd3948b. Three files, one functional change, all verified against the diff.

1. **The de-regex is behavior-identical.** Compared the removed character class `[:#\n"'{}[\],&*?|>!%@`]` against `YAML_QUOTING_CHARS` element by element: the same 19 characters, no additions, no omissions. The `value.trim() !== value` leading-whitespace condition is preserved (evaluated first now; the disjunction is commutative, so the quoting decision is unchanged for every input). Per-code-point Set membership and `.test` agree for all-BMP single characters, which is all this class ever contained. The committed content-snapshot check passing in CI independently confirms byte-identical generator output.
2. **Escaping strength is not weakened.** The quoted path still delegates to `JSON.stringify`, untouched — this PR only changes how the needs-quoting predicate is computed, not the escaping itself. No injection surface is opened.
3. **The normalizer docblocks are comment-only** and accurate: no Cursor/VS Code harness home invokes `revealui-harnesses hook` today, and the notes correctly anchor to GAP-293/GAP-421 so a dead-code sweep won't misfire. Zero runtime change in those files.

CI: all checks green except this gate.

RevealUIStudio:

Non-author review of `chore/gap-421-keep-hygiene` @ `dd3948b9d` (base `origin/test` @ `786f4733d`), run in a dedicated worktree. Full diff read: 33/1 lines in `claude-code.ts`, 8/0 in each normalizer.

### What I executed

**1. YAML escaping equivalence — proven, 0 mismatches over 662,246 inputs.**

I extracted the old predicate verbatim (`git show origin/test:packages/harnesses/src/content/generators/claude-code.ts`) and ran it head-to-head against the new `needsYamlQuoting` + `YAML_QUOTING_CHARS` over:

- **Exhaustive BMP sweep** — every code unit `U+0000`–`U+FFFF` in four positions (bare, `a<c>b`, `<c>abc`, `abc<c>`) = 262,144 cases.
- **Astral / surrogate cases** — emoji, `U+10000`, `U+10FFFF`, plane-14 tags, plus **lone high and low surrogates** and reversed pairs. This is the one place a code-unit regex and a code-point `for…of` could legitimately diverge; they do not, because every member of the set is single-code-unit ASCII, so a surrogate half can never alias one.
- **Hand-built YAML-injection corpus** — `name: injected`, `\n`/`\r`/`\r\n`/`U+2028`/`U+2029`/`U+0085` key-injection attempts, `---` document breaks, `!!python/object`, `%YAML 1.2`, anchors/aliases, block and folded indicators, backslash sequences, C0 controls, NUL, NBSP, RTL/CJK text, 5,000-char values with and without a trailing space.
- **400,000 randomized fuzz cases** over a YAML-hostile alphabet (all 19 set members plus `\r`, `\t`, `\\`, `-`, `~`, `.`, line separators, an astral pair, and bare surrogates), lengths 1–12.

Result: `inputs checked: 662246 / mismatches: 0`. The character sets are identical (19 members each) and the reordering of the `value.trim() !== value` check is inert under `||`. **The refactor is behavior-preserving, not merely "looks equivalent".**

On the threat model specifically: forging a new frontmatter key requires `\n` *and* `:`, and forging a list entry requires `\n`. Both are in the set, before and after. There is no input class where the new code emits a plain scalar that the old code quoted, so no new YAML-structure injection is reachable through a crafted rule/agent/command/skill name.

Worth stating for the record, since I probed for it: values containing only `\r`, `U+2028`, `U+2029`, `U+0085`, a bare `-`, or C0 controls still pass through unquoted. That is **identical in old and new** (I verified side by side), it is pre-existing, and it is not exploitable as key injection without a `:`. Out of scope here; flagging only so it is not mistaken for something this PR introduced.

**2. Normalizers — proven comment-only, byte-identical below the docblock.**

Rather than eyeballing the hunks, I stripped the leading `/** … */` from base and head of both files and hashed the remainder:

```
cursor.ts  post-docblock bytes: base=2577 head=2577  sha=99c29245662ee7ac/99c29245662ee7ac  IDENTICAL=true
vscode.ts  post-docblock bytes: base=2746 head=2746  sha=57e0cd8e80cedee9/57e0cd8e80cedee9  IDENTICAL=true
```

Zero deletions in both files; every added line falls inside the leading module docblock (`cursor.ts` 1–26, `vscode.ts` 1–44). **No validation or normalization semantics changed.** The trust-boundary invariants stay intact: `user_email` is still read-only-into-`raw` (I-1), and the existing suite still proves `null` / `undefined` / `'not an object'` inputs do not throw.

**3. Repo hygiene.**

- `pnpm --filter @revealui/harnesses test` — **41 files / 542 tests pass**, including `content-snapshot.test.ts` (the GAP-406 definition ↔ committed-snapshot lock).
- `pnpm --filter @revealui/harnesses typecheck` — clean.
- `biome check` on all three changed files — clean, no fixes applied.
- No-regex compliance: no regex literal, `new RegExp`, `.match(`, or `.replace(/` in any of the new code. The replacement uses exactly the alternative `no-regex.md` prescribes for character-class checks.
- Audit-first: the PR is step 3 of the GAP-421 harnesses routing audit and says so, with steps 6/8 explicitly scoped out. Compliant.
- No changeset, correctly — no consumer-visible surface changed. Verified against the emitted output, not just asserted.

### PR claims verified

- *"Output is unchanged, the committed content snapshot check still passes"* — **true**, and the 662k-input equivalence result is a stronger statement than the snapshot alone.
- *"no Cursor or VS Code harness home invokes `revealui-harnesses hook` yet"* — **true today**: no `.cursor/hooks.json` and no VS Code `plugin.json` are materialized anywhere in the tree, and `.revealui/manager.json` declares the adapters without emitting hook wiring.

### Suggestions (non-blocking)

- [`packages/harnesses/src/content/generators/claude-code.ts:23`](https://github.com/RevealUIStudio/revealui/blob/dd3948b9d/packages/harnesses/src/content/generators/claude-code.ts#L23) — the two injection-critical characters are the two the test suite does not lock. I scanned all 57 `name` / `description` / `argumentHint` literals in `src/content/definitions/`: 8 trigger the quoting branch, but every one of them trips on `,` or `>`, and **not one contains `:` or `\n`**. So the sha256 content snapshot would catch someone dropping `,` or `>` from `YAML_QUOTING_CHARS`, and would silently pass if someone dropped `:` or `\n`. That risk is new in kind, not degree: a 19-line editable array invites a "tidy up this list" edit in a way a compact character class did not. A three-line unit test in `claude-code-content-generator.test.ts` asserting that a description containing `:` and one containing `\n` both come back JSON-quoted would close it permanently. Worth doing, but it is a pre-existing coverage gap and this PR does not widen the actual behavior, so it does not block.

- [`packages/harnesses/src/hooks/normalizers/cursor.ts:18`](https://github.com/RevealUIStudio/revealui/blob/dd3948b9d/packages/harnesses/src/hooks/normalizers/cursor.ts#L18) and [`vscode.ts:36`](https://github.com/RevealUIStudio/revealui/blob/dd3948b9d/packages/harnesses/src/hooks/normalizers/vscode.ts#L36) — "zero live callers" is accurate about *wiring*, but a future reader can easily read it as *unreachable*, which is not the case. `CursorGenerator` and `VSCodeGenerator` are registered eagerly in `content/generators/index.ts:34-36`, and a single `revealui-harnesses content sync --generator cursor` writes `.cursor/hooks.json` binding `revealui-harnesses hook cursor` to all eleven Cursor events (`content/generators/cursor.ts:82`; VS Code equivalent at `vscode.ts:112`). The `hook` CLI subcommand also dispatches on an argv-supplied source into `normalizeHookEvent`, so the path is directly invocable today. Since these normalizers are exactly where untrusted editor payloads land, I would add half a sentence — "reachable via the CLI and one `content sync` away from being wired; treat payloads as untrusted regardless" — so the note documents dormancy without ever reading as an invitation to relax scrutiny.

### Nits

- [`packages/harnesses/src/content/generators/claude-code.ts:45`](https://github.com/RevealUIStudio/revealui/blob/dd3948b9d/packages/harnesses/src/content/generators/claude-code.ts#L45) — `needsYamlQuoting` allocates a one-character string per code point where the old regex scanned code units in place. Irrelevant here (build-time generation over frontmatter-length strings; I measured 5,000-char inputs with no issue) and both are linear with no backtracking, so there is no ReDoS trade either way. Noting it only so the choice is on the record as deliberate.

### Bottom line

Behavior-preserving refactor on a security-listed path, proven equivalent rather than assumed, plus two comment-only additions with byte-identical code beneath them. Nothing in this diff changes what reaches a YAML document or a hook decision. No blockers.

Review worktree removed after review. Per `disposition-actions.md` I am not merging and not applying labels.

PR #2169, opened 2026-07-25
refactor(harnesses): delete zero-consumer coordination daemon twin

RevealUIStudio:
## Summary

Executes the GAP-421 harnesses-routing audit's step 8 DELETE train, per the
internal daemon-ownership ADR (ratified by owner directive 2026-07-25, "the
word").

The RevDev daemon owns the harness socket and the coordination runtime
(`session.*`, `mail.*`, `files.*`, `tasks.*`, `events.*`). `@revealui/harnesses`
shipped a second, zero-consumer JSON-RPC daemon on the same socket path — this
PR removes that twin.

**CLI-first sequencing (the ADR's correction to the audit's own sequencing
note).** `cli.ts` imported `HarnessCoordinator`, and `coordinator.ts` imported
every DELETE-set module. The first mechanical step was trimming `cli.ts`:
removed the `start` subcommand and the `coordinate --init` daemon path (the
only two callers of `HarnessCoordinator`), which severed the DELETE set
cleanly with zero remaining production imports.

## Deleted (3,002 source LOC + 1,211 test LOC)

- `goals/` (533 LOC + its test)
- `server/rpc-server.ts` (1,018 LOC + its test)
- `server/spawner-service.ts` (289 LOC + its test)
- `server/shared-memory-client.ts` (192 LOC)
- `coordinator.ts` (325 LOC + its test + `__tests__/integration.test.ts`, which
  was entirely a `HarnessCoordinator` multi-agent test)
- `coordination/` (645 LOC: `ci-feedback.ts` + `merge-pipeline.ts`)
- `cli.ts`'s `start` command and `coordinate --init` branch, and
  `__tests__/cli.test.ts` (its only test was "start command invokes
  coordinator")
- The published `./goals` subpath export (package.json `exports` + tsup entry)

## Kept, contrary to the ADR's DELETE list — two consumers the audit missed

Auditing the actual import graph before deleting (audit-first) surfaced two
real consumers neither the audit nor the ADR accounted for. Per this task's
instruction to stop rather than force when a deletion breaks something
unexpected, both modules are **excluded** from this delete pass:

- **`storage/` (1,300 LOC).** `server/__tests__/http-gateway.test.ts` is a
  637-line security test suite (GAP-353, the gateway's fail-closed auth flow)
  that constructs the real `DaemonStore` and exercises real persistence
  (bootstrap-secret hashing, singleton-row upsert, token expiry/revocation).
  Deleting `storage/` would have required rewriting that suite against a fake
[...]

PR #2173, opened 2026-07-25
docs(harnesses): correct http-gateway docblock — daemon-store.ts was kept, not deleted

RevealUIStudio:
Comment-only follow-up to the revealui#2169 guardrail-2 review. The http-gateway docblock listed `storage/daemon-store.ts` among the daemon-ownership-ADR DELETED modules, but it was KEPT (one of the two STOPs); the GatewayStore interface decouples from the concrete store rather than covering a deleted one. Docblock corrected — no code change.

Touches a server/ path so the security-review gate will hold it for a non-author guardrail-2 verdict; comment-only diff.

### DIFF
--- a/packages/harnesses/src/server/http-gateway.ts
+++ b/packages/harnesses/src/server/http-gateway.ts
@@ -15,13 +15,15 @@ import { createServer, type IncomingMessage, type Server, type ServerResponse } 
 import { dirname, extname, join, normalize } from 'node:path';
 
 /**
- * Structural contracts covering this gateway's real dependencies on modules
- * deleted per the daemon-ownership ADR (2026-07-25): `server/rpc-server.ts`,
- * `storage/daemon-store.ts`, and `server/spawner-service.ts`. The RevDev daemon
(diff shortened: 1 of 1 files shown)


RevealUIStudio:

Non-author review of head f13a0bf. The diff is strictly comment-only: one docblock in `packages/harnesses/src/server/http-gateway.ts`, no code tokens touched (verified by reading the full diff — the only hunk edits prose between the `/** */` markers).

The correction is factually right, verified against `origin/test` rather than the PR prose: `storage/daemon-store.ts` exists on test (KEPT, one of the ADR's STOPs) while `server/rpc-server.ts` and `server/spawner-service.ts` are absent (DELETED per the daemon-ownership ADR) — exactly the split the new docblock describes and the old one misstated. A stale claim that a kept store was deleted is precisely the doc-drift class code-over-docs kills; good catch from the #2169 review to fold it here.

No auth, routing, or serving behavior can change from this diff. CI: all checks green except this gate.

PR #2179, opened 2026-07-26
fix(security): anchor the audit singleton on globalThis across bundled module copies

RevealUIStudio:
## What

Anchors the process-wide `audit` singleton (`packages/security/src/audit.ts`) on `globalThis` (keyed via `Symbol.for`) so every bundled copy of the module resolves the same `AuditSystem` instance.

## Why (live prod finding, GAP-338 follow-up)

`admin.revealui.com/api/health` returns **503 `{"status":"unhealthy"}`** on the current production deploy: the route bundle reports audit storage as in-memory in production, meaning admin route-level audit emits (login receipts included) evaporate on restart. This is exactly the cross-bundle failure the #2156 health check was built to catch — and it fired.

Root cause, reproduced locally on a clean production build (`next start`, DB env + Ed25519 signing key provisioned, same shape as the #2176 e2e env):

- The instrumentation chunk installs persistent storage successfully (`AUDIT SIGNING: ENABLED`, self-test reaches a real SQL query) — but into **its own copy** of the module graph.
- Turbopack bundles a separate copy of `@revealui/security` into the instrumentation chunk and route chunks (9 server chunks carried `InMemoryAuditStorage` after a clean build), so the swap never reaches the singleton the routes resolve.
- `serverExternalPackages: ['@revealui/security']` was tried first and does **not** take effect for this workspace package (chunks still bundled after a clean `--force` rebuild; `@revealui/config` externalizes fine, so this is specific to the package/subpath shape). Fighting the bundler is not durable; module identity cannot be trusted across bundles. `globalThis` is the process-wide anchor (the standard Next.js pattern for this duplication class).

## Proof

- **Red:** with base `audit.ts`, the new test (two independent module registries via `vi.resetModules()` + dynamic import) fails 2/2; local prod build serves `/api/health` → `503 {"status":"unhealthy"}`.
- **Green:** with this fix, the test passes 2/2; the same clean prod build + same env serves `/api/health` → `200 {"status":"healthy"}`.
- Full `@revealui/security` suite: 752 passed. `@revealui/auth` (owns the audit-storage boundary): 531 passed. Admin `instrumentation-audit-storage.test.ts`: 11 passed. Typecheck + Biome clean.

## Notes

- After merge + promote + deploy, the hosted prod `/api/health` should flip to 200 with no env changes (env parity was never the problem — the rails proved env present).
- Known residual (out of scope): module-level counters elsewhere in the package (e.g. audit write-failure ratios) are still per-copy under bundle duplication — monitoring granularity only, no correctness impact on the write path.
- Security-classed surface (`packages/security/`): **review-pending** — needs a recorded non-author guardrail-2 verdict before merge.

### DIFF
--- a/.changeset/audit-singleton-globalthis.md
+++ b/.changeset/audit-singleton-globalthis.md
@@ -1,0 +1,5 @@
+---
+'@revealui/security': patch
+---
+
+Anchor the process-wide `audit` singleton on `globalThis` (keyed via `Symbol.for`) so bundlers that duplicate the package's module graph across chunks (observed with Turbopack in the admin production build) still resolve one shared `AuditSystem` instance. Fixes the boot-time persistent-storage swap landing on a different module copy than the one routes resolve, which left production admin audit emits in-memory and `/api/health` reporting `audit-storage: unhealthy`.
--- a/packages/security/src/__tests__/audit-singleton-global.test.ts
+++ b/packages/security/src/__tests__/audit-singleton-global.test.ts
@@ -1,0 +1,57 @@
+import { describe, expect, it, vi } from 'vitest';
+
+/**
+ * GAP-338 follow-up: the `audit` singleton must be PROCESS-wide, not
+ * module-copy-wide.
+ *
+ * The admin production build bundles a separate copy of this package's module
+ * graph into the instrumentation chunk and each route chunk (Turbopack;
+ * `serverExternalPackages` does not take effect for the workspace package), so
+ * a plain module-level singleton splits: the boot-time persistent-storage swap
+ * lands on one copy while routes emit into their own in-memory copies —
+ * observed live as `/api/health` `audit-storage: unhealthy` on hosted prod
+ * (2026-07-25).
+ *
+ * `vi.resetModules()` + dynamic import simulates exactly that: two independent
+ * module registries each evaluating `audit.ts` from scratch. Without the
+ * `globalThis` anchor the two evaluations construct two different
+ * `AuditSystem` instances and this test FAILS (proven red before the fix);
+ * with the anchor both copies resolve the same instance, so a storage swap
+ * installed through either copy is visible to the other.
+ */
+describe('audit singleton across duplicated module copies', () => {
+  it('two independent module registries resolve the same AuditSystem instance', async () => {
+    vi.resetModules();
+    const firstCopy = await import('../audit');
+    vi.resetModules();
+    const secondCopy = await import('../audit');
+
+    expect(firstCopy.audit).toBe(secondCopy.audit);
+  });
+
+  it('a storage swap through one copy is visible to the other copy', async () => {
+    vi.resetModules();
+    const firstCopy = await import('../audit');
+    vi.resetModules();
+    const secondCopy = await import('../audit');
+
+    // A minimal non-in-memory storage stub: after swapping it in through the
+    // FIRST copy, the SECOND copy must stop reporting in-memory storage — the
+    // exact cross-bundle contract the admin /api/health `audit-storage` check
+    // verifies in production (instrumentation swaps, route observes).
+    const stub = {
+      write: async () => undefined,
+      query: async () => [],
+      count: async () => 0,
+    };
+    firstCopy.audit.setStorage(stub);
+    try {
+      expect(firstCopy.audit.isInMemoryStorage()).toBe(false);
+      expect(secondCopy.audit.isInMemoryStorage()).toBe(false);
+    } finally {
+      // Restore the default so the shared process-wide singleton does not leak
+      // the stub into other suites in this worker.
+      firstCopy.audit.setStorage(new firstCopy.InMemoryAuditStorage());
+    }
+  });
+});
--- a/packages/security/src/audit.ts
+++ b/packages/security/src/audit.ts
@@ -712,6 +712,38 @@ export class AuditReportGenerator {
 // (RFC 8785 canonicalization over the full row) — see `audit-signing.ts`.
 
 /**
- * Global audit system
+ * Global audit system — process-wide, anchored on `globalThis`.
+ *
+ * A plain module-level singleton is NOT process-wide under a bundler that
+ * duplicates this package's module graph across chunks. The admin production
+ * build does exactly that (Turbopack bundles a separate copy of
+ * `@revealui/security` into the instrumentation chunk and route chunks;
+ * `serverExternalPackages` does not take effect for this workspace package),
+ * so the GAP-338 boot-time persistent-storage swap landed on the
+ * instrumentation copy while every route emitted into its own pristine
+ * in-memory copy — observed live as `/api/health` `audit-storage: unhealthy`
+ * on hosted prod (2026-07-25), the exact cross-bundle failure the #2156
+ * health check exists to catch. Module identity cannot be trusted across
+ * bundles; `globalThis` + `Symbol.for` is the process-wide anchor, so every
+ * copy of this module resolves the same instance and a storage swap installed
+ * by any copy is seen by all of them.
  */
-export const audit = new AuditSystem(new InMemoryAuditStorage());
+const AUDIT_SYSTEM_GLOBAL_KEY = Symbol.for('revealui.security.audit-system');
+
+interface AuditSystemGlobal {
+  [AUDIT_SYSTEM_GLOBAL_KEY]?: AuditSystem;
+}
+
+const auditSystemGlobal = globalThis as AuditSystemGlobal;
+
+function resolveProcessWideAuditSystem(): AuditSystem {
+  const existing = auditSystemGlobal[AUDIT_SYSTEM_GLOBAL_KEY];
+  if (existing) {
+    return existing;
+  }
+  const created = new AuditSystem(new InMemoryAuditStorage());
+  auditSystemGlobal[AUDIT_SYSTEM_GLOBAL_KEY] = created;
+  return created;
+}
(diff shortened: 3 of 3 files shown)


RevealUIStudio:

Non-author adversarial review of the `globalThis` anchor. I tried to break it along six axes: cross-copy class identity, globalThis leakage/authority widening, test pollution, the API-server process, the declared residual, and Edge/HMR/`Symbol.for` collisions. The fix holds. Details below, with the two sharp edges it leaves behind (neither reachable today, both worth a follow-up).

### What I tried to refute, and what happened

**1. Cross-copy `instanceof` — the strongest attack, and it does not land.**

The anchored instance is constructed by whichever module copy evaluates first (`packages/security/src/audit.ts:743-746`). Its methods therefore close over *that copy's* module scope. `AuditSystem.isInMemoryStorage()` (`packages/security/src/audit.ts:129-131`) is class-identity dependent:

```ts
return this.storage instanceof InMemoryAuditStorage;
```

So if a *different* copy ever installed its own `new InMemoryAuditStorage()`, the probe would return `false` while storage was genuinely in-memory. That failure direction is the bad one: a **false-healthy** `/api/health` `audit-storage` check, which is exactly the signal this PR exists to restore.

It does not fire, and I checked every path rather than assuming:
- The only production `audit.setStorage(...)` in the repo is `packages/auth/src/server/audit-storage.ts:306`, which installs `DrizzleBackedAuditStorage` — not an `InMemoryAuditStorage` of any copy, so `instanceof` is correctly `false` regardless of which copy owns the class.
- The only `InMemoryAuditStorage` the singleton ever holds by default is the one constructed inside `resolveProcessWideAuditSystem()` at `packages/security/src/audit.ts:745` — same copy as the instance, so identity always matches and `true` stays truthful.
- No other `new InMemoryAuditStorage()` + `setStorage` pair exists outside tests.

Worth noting the new test at `packages/security/src/__tests__/audit-singleton-global.test.ts:54` restores with `new firstCopy.InMemoryAuditStorage()` — correct today only because `firstCopy` is the creating copy. Written with `secondCopy` it would leave the singleton permanently reporting "not in-memory." That is the sharp edge, not a bug in this PR.

**2. `AuditWriteError` identity across copies — checked, no regression.**

Every catch site is in the API server: `apps/server/src/middleware/audit.ts:92,96,97` and `apps/server/src/routes/cron/uptime-check.ts:60,62,63`, all importing from `@revealui/core/security` (which re-exports `@revealui/security/server` — `packages/core/src/security/index.ts`). `apps/server` builds as a single bundled graph (`apps/server/tsup.config.ts`: `bundle: true`, `noExternal: /^@revealui\/(?!ai|services|mcp)/`), and `packages/security` emits `AuditSystem`/`AuditWriteError` once into a shared chunk (`packages/security/dist/chunk-MR6AQU3U.js:13`) rather than per entry, so `.`/`./server` do not duplicate the class. Admin — the app that actually duplicates chunks — has **zero** `AuditWriteError` catch sites. So the anchor cannot change which class is thrown anywhere it is caught.

**3. globalThis exposure is not an authority widening.**

Any code already in the process can `import { audit } from '@revealui/security/server'` and call `setStorage`. The symbol adds a second way to reach an object that was never access-controlled, not new authority. It also never reaches a browser realm: `audit` is exported only from `packages/security/src/server.ts:32`, not the client-safe `.` barrel (`packages/security/src/index.ts` has no audit export), and `packages/security/tsup.config.ts` keeps the `node:`-bearing graph on the `./server` subpath alone. If it *did* ship to a page, `Symbol.for` would make it reachable from any third-party script — worth keeping that boundary as-is.

**4. Edge runtime — not exercised, and the guard is explicit.**

Every `export const runtime` in `apps/admin` is `'nodejs'`; there are no edge routes. `apps/admin/src/proxy.ts` imports only local CSP/CSRF helpers, so middleware never drags in the audit graph. `apps/admin/src/instrumentation.ts:30` returns early on `'EdgeRuntime' in globalThis`, and the audit install is behind `NEXT_RUNTIME === 'nodejs'` at line 250 via a dynamic `import('./instrumentation-node')` — the file itself documents that the guard is what excludes it from the Edge build (`apps/admin/src/instrumentation-node.ts:9-10`). No Edge `globalThis` assignment happens.

**5. Test pollution — nothing breaks today, but the premise in the new test's cleanup comment is wrong.**

I want to correct an assumption that is easy to make here, because I made it first and it is false. `packages/dev/src/vitest/create-config.ts:111-112` sets `pool: 'forks'`, `maxWorkers: 2`, and never sets `isolate`, so Vitest's default `isolate: true` applies — and in Vitest 4 that does **not** mean one process per test file. The installed worker loop (`vitest@4.1.10`, `dist/chunks/base.B6Opl8PE.js`) does:

```js
for (const file of files) {
  if (config.isolate) {
    moduleRunner.mocker?.reset();
    resetModules(workerState.evaluatedModules, true);
  }
  ...
}
```

That loop runs *inside one forked worker*, iterating the files assigned to it. `isolate: true` resets the **module registry** between files; it does not reset `globalThis`. With 19 files over 2 workers, most files share a process. So the anchored singleton now genuinely persists across test files — module-scope state still gets a fresh copy per file, `globalThis` state does not.

Nothing breaks today, and I verified that rather than inferring it:
- No suite anywhere asserts `isInMemoryStorage()` on the *singleton*. `packages/security/src/__tests__/audit.test.ts:430-447` exercises a locally constructed `AuditSystem`, not the export. `security.test.ts:6` imports `../audit` but only for the classes.
- `apps/admin/src/__tests__/instrumentation-audit-storage.test.ts:23-27` mocks `@revealui/auth/audit-storage` wholesale, so `installAuditStorage` is a `vi.fn()` and never reaches the real `setStorage`. That is what keeps `apps/admin/src/app/api/health/__tests__/health-routes.test.ts:83-89` — which does depend on the real singleton still being in-memory — order-independent.
- The one `isolate: false` config (`packages/test/vitest.integration.config.ts:44`) never touches audit storage.
- Full suite re-run independently on this branch: 19 files / 752 tests passed; `audit.test.ts` + `audit-singleton-global.test.ts` together, 29 passed.

The booby trap is in the new test's cleanup at `packages/security/src/__tests__/audit-singleton-global.test.ts:52-55`. Its comment says the restore stops the stub leaking "into other suites in this worker" — that is now a real concern rather than a hypothetical one, and the restore only half-works. If an earlier file in the same worker already created the global instance, `firstCopy` is a *fresh* copy whose `InMemoryAuditStorage` is a different class object, so `new firstCopy.InMemoryAuditStorage()` installs a storage the instance's own `instanceof` check does not recognize, and `audit.isInMemoryStorage()` stays `false` for every later file in that worker. Harmless right now because nothing asserts it — and a nondeterministic, distribution-dependent failure for the next person who writes the obvious test for this feature. This is the same root cause as finding 1, and the same fix closes both.

**6. API server — no regression, the anchor is a no-op there.**

`apps/server` runs one bundled graph, so it had one copy before and resolves the same object now. The installer chain at `apps/server/src/index.ts:1421-1422` (`assertAuditStorageEnv()` then `installAuditStorage()`) targets the identical instance the routes emit through, exactly as before.

**7. Residual scoping is honest.**

`recordAuditWriteResult` (`packages/security/src/audit-write-failures.ts:154`) keeps module-level rolling-window state, so under chunk duplication each copy carries its own 500-attempt window with a 20-attempt minimum before the 1% escalation can fire. The effect is a diluted and delayed alarm, never a wrong write-path outcome — which is what the PR body claims. Accepting that as scoped, but it is the same defect class with the same one-line remedy, so it should not sit unowned.

**8. The test is a real proof, not a tautology.**

`vi.resetModules()` + dynamic import does produce genuinely distinct class objects per registry, so `expect(firstCopy.audit).toBe(secondCopy.audit)` is a true cross-registry assertion. The second test's `secondCopy.audit.isInMemoryStorage()` is implied by the first (same object) — redundant but harmless.

### Non-blocking follow-ups

- **Make the storage probe identity-free — highest-value follow-up.** Replace the `instanceof` in `packages/security/src/audit.ts:130` with a brand the duplicated copies agree on (a `readonly kind = 'in-memory'` discriminant on `InMemoryAuditStorage`, or a `Symbol.for`-keyed marker). The PR fixed identity for the *instance* and left an identity-dependent check inside it. Branding closes the false-healthy window permanently, makes the new test's cleanup at `audit-singleton-global.test.ts:52-55` actually restore, and lets a reset-to-in-memory helper exist safely.
- **Version-scope the symbol key.** `Symbol.for('revealui.security.audit-system')` at `packages/security/src/audit.ts:731` is unversioned. In-workspace that is fine (`workspace:*` everywhere, one version). For an external npm consumer that ends up with two `@revealui/security` versions in one process, first-loaded now wins process-wide and a newer copy could call a method the older instance lacks — a `TypeError` class that does not exist today, where each version simply kept its own instance. `…audit-system.v1` plus a shape guard on the resolved value removes it.
- **No changeset.** `.changeset/` on this branch holds only `config.json`, and `@revealui/security` is published at `0.6.0`. Per `.claude/rules/versioning.md`, a meaningful behavior change to a published 0.x package is a minor bump routed through `pnpm changeset`. This is a release-pipeline gap, not a security one, but it should land before this reaches npm consumers.

### Verdict

The diagnosis matches the code, the fix is the right shape for it (module identity genuinely cannot be trusted across Turbopack chunks), the boot-order is symmetric so it works whichever copy evaluates first, and the failure modes I could construct all require a code path that does not exist in this repo. Approving. Merge disposition and the `sec-review:approved` label remain the owner's.

PR #2182, opened 2026-07-26
fix(server): legacy anchor floor so the audit sweep clears the pre-enforcement era (GAP-427)

RevealUIStudio:
## Summary

Anchoring has never completed in prod (GAP-427): the pre-enforcement era left 170 unsigned `audit_log` rows (seq 1..1122, 123 of them interleaved above the first signed row), and every unsigned row is a permanent contiguity hole for the anchor sweep. Since the both-rails promotion guarantees no new unsigned rows, that era is closed, so the sweep now derives a per-tenant **legacy floor** (the highest unsigned `seq`) and starts contiguity at `floor + 1` when the tenant has no anchors yet.

## What changed

- `apps/server/src/jobs/audit-anchor-sweep.ts`: new `legacyUnsignedFloor()` (derived at sweep time, no new state). Consulted **only when a tenant has zero anchors**; once anchors exist, strict `last + 1` contiguity governs unchanged, so a hole appearing above an existing anchor still stalls the sweep and raises the gap metric.
- **No retro-signing.** Legacy rows (unsigned + the signed rows interleaved below the floor) stay outside every Merkle root. A backfilled signature would be indistinguishable from tampering, which is the property the receipts exist to prevent.
- `docs/security/AUDIT_RECEIPTS.md`: new "Legacy floor" section documenting the pre-floor era honestly (row-level verify still works below the floor; no root covers those rows).

## Behavior change note (reviewed intentionally)

The prior test `skips when signed seqs jump after last anchor (unsigned hole)` locked the stall-at-first-hole behavior for a tenant with **no anchors** — exactly the state GAP-427 rules is wrong (that fixture is a closed legacy era). It is reworked to lock the surviving semantics: an unsigned hole appearing **above an existing anchor** stalls, never jumps.

## Tests (prove-red)

New PGlite test `GAP-427: first anchor starts above the closed unsigned legacy era (floor)` with fixture unsigned(1), signed(2), unsigned(3), signed(4), signed(5):

- **Base code (red, run with the fix checked out to origin/test):** sweep anchors the interleaved legacy prefix `[2..2]`, then gap-stalls forever; assertion `seqFrom = 4` fails with `expected 2 to be 4`.
- **With the fix (green):** floor = 3, the single anchor covers exactly the signed era `[4..5]`; suite 12/12.

Note on the mechanism: with zero anchors, `planContiguousBatch` applies no gap check, so the actual prod stall shape is "anchor the first contiguous signed run, then stall at the first interleaved unsigned row" rather than "stall at seq 1". The fixture reproduces that real shape.

## After deploy

First real prod anchor landing (`audit_anchors` count > 0 after a sweep tick) is the observation that closes GAP-427 and unblocks Stage 4 anchoring.

### DIFF
--- a/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
+++ b/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
@@ -237,14 +237,15 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     expect(anchors.some((a) => a.tenant === '')).toBe(false);
   });
 
-  it('skips when signed seqs jump after last anchor (unsigned hole)', async () => {
+  it('skips when an unsigned hole appears above an existing anchor (strict contiguity)', async () => {
     const signedStore = new DrizzleAuditStore(db, canonicalSignerFn(priv));
     const unsignedStore = new DrizzleAuditStore(db); // no signer → signature null
-    await signedStore.append(makeEntry({ id: 's1', tenant: 'acct_max' }));
-    await unsignedStore.append(makeEntry({ id: 'u2', tenant: 'acct_max' }));
-    await signedStore.append(makeEntry({ id: 's3', tenant: 'acct_max' }));
 
-    // First pass anchors contiguous prefix [seq 1] only (size/lag ready).
+    // Anchor seq 1 first, THEN an unsigned row lands above it. Post-GAP-417
+    // rails this cannot happen in prod; if it does, it is a genuine integrity
+    // signal — the sweep must stall on it, never jump it (GAP-427 floor only
+    // applies while a tenant has no anchors at all).
+    await signedStore.append(makeEntry({ id: 's1', tenant: 'acct_max' }));
     const first = await runAuditAnchorSweep({
       db,
       signer: cryptoSigner,
@@ -254,6 +255,9 @@ describe('runAuditAnchorSweep (PGlite)', () => {
       now: () => new Date('2026-07-23T12:00:10.000Z'),
     });
     expect(first.anchorsInserted).toBe(1);
+
+    await unsignedStore.append(makeEntry({ id: 'u2', tenant: 'acct_max' }));
+    await signedStore.append(makeEntry({ id: 's3', tenant: 'acct_max' }));
 
     // last=1; next signed is seq 3 (seq 2 unsigned). Gap → no further insert.
     const second = await runAuditAnchorSweep({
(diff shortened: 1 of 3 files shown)


RevealUIStudio:

Reviewed independently from the PR head (`b00f3fb`) against `origin/test`. I did not author this change.

## What I verified

**Tests, run by me (not read from the PR body).**

- Green, PR head: `pnpm --filter server exec vitest run src/jobs/__tests__/audit-anchor-sweep.test.ts` to **12 passed (12)**.
- Red, base code + PR tests: reverted only `apps/server/src/jobs/audit-anchor-sweep.ts` to `origin/test` and reran. Exactly one failure, the new GAP-427 test, `AssertionError: expected 2 to be 4` at `audit-anchor-sweep.test.ts:312`. The run also printed the real prod stall shape the PR describes: `anchored tenant=acct_max seq=2..2` then `gap for tenant=acct_max lastAnchored=2 nextSeq=4 - skip`. So the fixture reproduces the interleaved-legacy stall, not a strawman. File restored to HEAD afterward, working tree clean.

**1. Correctness of the floor gate.** The floor is consulted only at `audit-anchor-sweep.ts:262`, `anchored > 0 ? anchored : await legacyUnsignedFloor(...)`. `lastAnchoredSeq` reads `max(audit_anchors.seq_to)` for the tenant; `seq_to` is `notNull`, is only ever written by this job from a real `audit_log.seq` (bigserial, starts at 1), and the table carries `audit_anchors_seq_range_check (seq_to >= seq_from)` plus `leaf_count > 0`. There is no path that can persist `seq_to <= 0`, so `anchored > 0` is a faithful "has anchors" test. Drizzle's `max()` maps through the column decoder for a bigint `mode: 'number'` column, so the `typeof m === 'number'` guard does not silently fall through to 0 on a driver string. The floor can only raise `last`, never lower it, and candidate selection stays `gt(auditLog.seq, last)`, so no row below the floor can enter a root.

Can the floor skip live rows? Only if an unsigned row were written above signed rows *while the tenant still has zero anchors*. I checked the rail rather than trusting the premise: `packages/db/src/audit-store.ts:116` `refuseUnsignedInProduction()` throws on any signer-less `append` under `NODE_ENV=production`, backing the boot-time assert. So the legacy era is genuinely closed and the floor is stable.

**2. Security.** The diff adds one `SELECT` and nothing else. No `UPDATE`, no `DELETE`, no signature write, no backfill anywhere in the changed file. Legacy rows are untouched, which is the GAP-427 ruling. The gap warn plus `revealui_audit_anchor_seq_gaps_total` survive intact for holes above an existing anchor (`audit-anchor-sweep.ts:287`), and they also still fire when the floor is engaged but the next signed row is not `floor + 1`, so a floor-era stall is loud, not silent.

**3. The test rework.** The assertion that was dropped (a no-anchors tenant stalling at the first unsigned hole) is precisely the behavior GAP-427 rules wrong, so removing it loses nothing worth keeping. The replacement locks the property that actually matters going forward, strict `last + 1` above an existing anchor, and the red run confirms it is a real lock. The all-signed path (floor 0, behavior identical to before) is covered by the 11 pre-existing tests, which pass unchanged.

**4. Edge cases.** Tenant with unsigned rows and no signed rows is never selected (the tenant scan requires `isNotNull(signature)`). Tenant fully signed derives floor 0, so behavior is byte-identical to base. Null-tenant rows are correctly excluded from the derivation by `eq(auditLog.tenant, tenant)`: the sweep never anchors null-tenant rows, and folding a null-tenant `seq` into a tenant's floor would wrongly exclude that tenant's own signed rows from every root. Exclusion is the right call.

**5. Docs.** `docs/security/AUDIT_RECEIPTS.md` matches the code line for line, including "only consulted while a tenant has no anchors at all" and "a hole above an existing anchor still stalls and raises the gap metric". Zero em dashes in the file, and no internal paths in the added copy. The "no root covers them" admission is the honest framing.

## Findings (none blocking)

1. **Precondition worth confirming before you call GAP-427 closed.** The floor engages only while `audit_anchors` has zero rows for the tenant. My red run shows base code anchors the interleaved legacy prefix (`seq=2..2`) on its first pass. If the prod sweep has ever run with a signer present, an equivalent legacy-prefix anchor may already exist, `anchored > 0` wins, and this fix is a no-op that leaves prod stalled exactly where it is. Recommend confirming `select count(*) from audit_anchors where tenant = ...` is 0 before the deploy tick, so the "first anchor lands" observation is meaningful either way.

2. **`floor + 1` is strict, and the floor only removes same-tenant unsigned holes.** Any non-tenant row sitting at `floor + 1`, or any null-tenant or other-tenant row interleaved above the floor, is still a permanent contiguity hole for that tenant (a pre-existing property of contiguity over a global bigserial, not introduced here). It fails loudly with the gap warn naming `lastAnchored` and `nextSeq`, so it is safe, but it means the fix unblocks prod only if the row at `floor + 1` is the tenant's own signed row. Worth one glance at the prod rows immediately above 1122 rather than discovering it from a warn line.

3. **The lag surface goes quiet about the sub-floor era.** `apps/server/src/routes/audit.ts:135-147` computes `unanchoredSignedCount` as `seq > lastAnchoredSeqTo`. Once the first floor-based anchor lands, the signed rows interleaved below the floor stop being counted as unanchored and stop being covered by any root, so a consumer reading "everything at or below `lastAnchoredSeqTo` is anchored" would be wrong. The prose in AUDIT_RECEIPTS.md is honest about this; the API is not. Suggest a follow-up that exposes the floor (or an explicit unrooted-below-floor count) in the lag payload, given the receipts positioning.

4. **No observability that the floor engaged.** Nothing logs or counts the fact that a sweep started at a derived floor. A one-line `logger.info` (or a counter) at `audit-anchor-sweep.ts:262` when `anchored === 0 && floor > 0` would make the one-time transition auditable rather than inferable from the first anchor's `seq_from`. Related: no test covers a floor above the entire signed era, which returns `skipped` with no metric and no warn. Post-rails that state should be unreachable, so this is polish, not a hole.

None of the above is a correctness or security defect in the diff. The change is conservative, fail-closed, does not touch legacy rows, and is proven red then green. Approving on the code. Merge and the `sec-review:approved` label remain the owner's call.

PR #2185, opened 2026-07-26
feat(server): audit lag surface names the legacy floor + floor-engaged observability (GAP-429)

RevealUIStudio:
## Summary

Follow-ups from the guardrail-2 review of #2182 (legacy anchor floor), findings 3 and 4: the anchors API went silent about signed rows below the pre-enforcement floor, and nothing observable recorded that the sweep engaged the floor.

## What changed

- **Lag surface honesty** (`apps/server/src/routes/audit.ts`): `GET /api/audit/anchors` lag object gains `legacyFloor` (0 when no legacy era) and `preFloorSignedCount`. Signed rows at or below the floor are row-verifiable but never enter a Merkle root (no retro-signing, per the floor ruling) — they are now named in the API instead of silently excluded. Existing lag fields keep their exact semantics (unanchored tail = signed rows above the last anchor).
- **Floor-engaged observability** (`apps/server/src/jobs/audit-anchor-sweep.ts`): new `tenantsFloorEngaged` result field, `revealui_audit_anchor_floor_engaged_total` counter, and a log line carrying the floor value. A floor above the entire signed era (nothing to anchor) is now visible rather than a silent skip.
- **One derivation, no copy**: `legacyUnsignedFloor()` is exported from the sweep and reused by the route.
- **Docs**: `AUDIT_RECEIPTS.md` Lag section documents the new fields and the tail-vs-pre-floor split.

## Tests (prove-red)

21/21 green with the change. Reverting only the two source files to `origin/test` fails exactly the four tests that assert the new surface (missing `legacyFloor` / `preFloorSignedCount` / `tenantsFloorEngaged`), including:

- route: legacy-era fixture (unsigned 1, signed 2, unsigned 3, anchored era 4..5) reports `legacyFloor=3`, `preFloorSignedCount=1`, `unanchoredSignedCount=0`
- route: all-signed history reports `legacyFloor=0`, `preFloorSignedCount=0`
- sweep: floor above the entire signed era reports engagement, anchors nothing, no error
- sweep: floor test asserts engagement on pass one and non-engagement once anchors exist

### DIFF
--- a/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
+++ b/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
@@ -301,8 +301,10 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     };
     const first = await runAuditAnchorSweep(sweepOptions);
     expect(first.anchorsInserted).toBe(1);
+    expect(first.tenantsFloorEngaged).toBe(1); // GAP-429: engagement is observable
     const second = await runAuditAnchorSweep(sweepOptions);
     expect(second.anchorsInserted).toBe(0);
+    expect(second.tenantsFloorEngaged).toBe(0); // anchors exist now — floor not consulted
 
     // Floor = 3 (highest unsigned seq): the one anchor covers exactly the
     // signed era. Legacy rows 1..3 (unsigned + interleaved signed) stay
@@ -311,6 +313,31 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     expect(anchors).toHaveLength(1);
     expect(anchors[0]?.seqFrom).toBe(4);
     expect(anchors[0]?.seqTo).toBe(5);
+  });
+
+  it('GAP-429: floor above the entire signed era reports engagement and anchors nothing', async () => {
+    const signedStore = new DrizzleAuditStore(db, canonicalSignerFn(priv));
+    const unsignedStore = new DrizzleAuditStore(db);
+    // Signed seq 1 sits below the floor (unsigned seq 2 closed the era); there
+    // is no signed era above the floor, so the sweep has nothing to anchor —
+    // but the engagement must still be visible, not a silent skip.
+    await signedStore.append(makeEntry({ id: 's1', tenant: 'acct_max' }));
+    await unsignedStore.append(makeEntry({ id: 'u2', tenant: 'acct_max' }));
+
+    const result = await runAuditAnchorSweep({
+      db,
+      signer: cryptoSigner,
+      batchSize: 1,
+      canAnchorTenant: async () => true,
+      recordMeter: false,
+      now: () => new Date('2026-07-23T12:00:10.000Z'),
+    });
+
+    expect(result.tenantsFloorEngaged).toBe(1);
+    expect(result.anchorsInserted).toBe(0);
+    expect(result.tenantsSkipped).toBe(1);
+    expect(result.errors).toEqual([]);
+    expect(await db.select().from(auditAnchors)).toHaveLength(0);
   });
 
   it('no-ops without a signer (unsigned mode)', async () => {
--- a/apps/server/src/jobs/audit-anchor-sweep.ts
+++ b/apps/server/src/jobs/audit-anchor-sweep.ts
@@ -75,12 +75,18 @@ const mWaiting = metrics.counter(
   'revealui_audit_anchor_waiting_total',
   'Tenants with unanchored rows still under batch-size and max-lag thresholds',
 );
+const mFloorEngaged = metrics.counter(
+  'revealui_audit_anchor_floor_engaged_total',
+  'Tenant sweep passes that started from a derived legacy floor (GAP-427/429)',
+);
 
 export interface AnchorSweepResult {
   tenantsConsidered: number;
   anchorsInserted: number;
   tenantsSkipped: number;
   tenantsWaiting: number;
+  /** Tenants whose pass started from a derived legacy floor (no anchors yet). */
+  tenantsFloorEngaged: number;
   nullTenantSignedRows: number;
   errors: string[];
 }
@@ -152,6 +158,7 @@ export async function runAuditAnchorSweep(
     anchorsInserted: 0,
     tenantsSkipped: 0,
     tenantsWaiting: 0,
+    tenantsFloorEngaged: 0,
     nullTenantSignedRows: 0,
     errors: [],
   };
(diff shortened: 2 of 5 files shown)


RevealUIStudio:



Reviewed independently from the PR head (`d687aa7`) against `origin/test` (`832a77f`). I did not author this change. Same reviewer as the #2182 verdict, of which this is the follow-up.

## What I verified

**Tests, run by me.**

- Green, PR head: `pnpm --filter server exec vitest run src/jobs/__tests__/audit-anchor-sweep.test.ts src/routes/__tests__/audit-anchors.test.ts` to **21 passed (21)**.
- Red: reverted only `apps/server/src/jobs/audit-anchor-sweep.ts` and `apps/server/src/routes/audit.ts` to `origin/test` and reran. **Exactly 4 failures**, all on the new surface and nothing else: `lists anchors for the tenant and reports lag`, `GAP-429: lag names the legacy floor and the sub-floor signed rows`, `GAP-427: first anchor starts above the closed unsigned legacy era (floor)`, `GAP-429: floor above the entire signed era reports engagement and anchors nothing`. Messages are `expected undefined to be 1 / +0 / 3`. Both files restored to HEAD, `git status --porcelain` empty, `git diff HEAD` on both paths empty.
- Also clean by me: `tsc --noEmit` on `server` (no output) and `biome check` on all four changed files (no fixes applied).

Honest note on the red check: for an additive observability PR the assertions can only fail by absence, so this proves the tests bind to the new code, not that a behavioral bug was fixed. That is the right ceiling for this change class, not a weakness in the tests.

**1. Route change, tenant scoping.** `tenant` still comes from `getUserAccountId(db, user.id)` at `audit.ts:112` and is unchanged. Both new reads are scoped to it: `legacyUnsignedFloor` filters `eq(auditLog.tenant, tenant)`, and the pre-floor count filters `eq(auditLog.tenant, tenant)` plus `isNotNull(signature)` and `lte(seq, legacyFloor)` (`audit.ts:157-174`). No user-supplied value reaches either query, so there is no cross-tenant path. Auth and entitlement gating are untouched: the diff does not modify `apps/server/src/index.ts:876-883`, where `writeProtected` and `requireFeature('auditLog', { mode: 'entitlements' })` are mounted on both the `/api/audit/anchors` and `/api/v1/audit/anchors` prefixes. Both queries are read-only; nothing in the diff writes, signs, or mutates an audit row.

**2. The `anchorTenantBatch` refactor.** Semantics are identical. `const floor = anchored > 0 ? 0 : await legacyUnsignedFloor(...)` short-circuits exactly as before, so the floor query still runs only when `anchored === 0`, and `const last = anchored > 0 ? anchored : floor` collapses to the merged `#2182` expression in both branches. I enumerated every exit from the function: `signed.length === 0`, the gap branch, the not-ready branch, the `seqFrom/seqTo undefined` branch, and the success return. All five go through `done()`. The declared return type makes this compiler-enforced rather than reviewer-enforced, and the typecheck is clean, so an unwrapped return cannot land later either. Outcome values themselves are unchanged, and the caller still maps `inserted` / `waiting` / everything-else the same way.

`floorEngaged = floor > 0` correctly excludes a tenant with no anchors and no unsigned rows, which is the "no legacy era" case rather than an engagement.

**3. API shape.** `/anchors` is a plain `app.get` returning `c.json`, not a `createRoute` + `app.openapi` registration, so there is no response schema to drift and no runtime response validation to break. `packages/contracts/src/generated/contracts.ts` carries only the `audit_anchors` **table** row/insert contracts, not the HTTP lag object. The sole cross-repo reference to the endpoint is prose in `apps/marketing/app/content/claims-evidence.ts:534`. Two additive fields on a JSON body are safe here.

**4. Docs.** The new Lag bullets match the code: `legacyFloor` is 0 when there is no legacy era (the derivation returns 0 on a null `max`), and `preFloorSignedCount` counts signed rows at or below the floor. The added sentence about the tail-vs-pre-floor split is exactly what `audit.ts:141` and `audit.ts:157-174` do. Zero em dashes in the file, no internal paths in the added copy.

## Findings (none blocking)

1. **`legacyUnsignedFloor` on the request path has no supporting index.** `packages/db/src/schema/audit-log.ts` has `audit_log_tenant_idx` on `tenant` alone and nothing involving `signature`. The worst case is the *healthy* case: a tenant with no unsigned rows forces Postgres to examine that tenant's whole row set to prove no NULL `signature` exists, and it happens on every `GET /anchors` (the response is `no-store`, so nothing absorbs it). Today prod's legacy era is ~1122 rows and this is invisible; on a tenant with a large audit history it becomes a per-request scan on a paid endpoint. The floor is immutable once the era is closed, so both fixes are cheap: a partial index (`on audit_log (tenant, seq desc) where signature is null`) or a short-lived cache of the derived value. Worth a follow-up rather than a change here.

2. **A route now imports from `jobs/`.** `audit.ts:24` pulls `legacyUnsignedFloor` from the sweep module, which has module-scope side effects: seven `metrics.counter()` / `gauge()` registrations now evaluate inside the API process too. I checked the failure mode rather than assuming it: `MetricsCollector.counter()` is a plain `Map.set` plus a fresh `Counter` (`packages/core/src/observability/metrics.ts:45-57`), so a second registration cannot throw the way a prom-client registry would, and `tsup.config.ts` builds `src/index.ts` and `src/worker.ts` as entries that ship in separate containers (`Dockerfile` vs `Dockerfile.worker`), so there is no in-process second copy of the module to strand a stale counter (the `#2179` failure class). No cycle either: the sweep's graph reaches only `lib/metering.ts` and `lib/account-entitlement.ts`, neither of which imports routes. So this is layering taste, not a defect. `apps/server/src/lib/` (next to `audit-anchor-proof.ts`) is the natural home for a derivation shared by a job and a route, and sharing one derivation is clearly better than duplicating the query.

3. **The floor-engaged log repeats on every poll.** `mFloorEngaged.inc()` plus the `logger.info` fire once per tenant per tick while the tenant has no anchors. For the normal transition that is one or two lines. For the state the new test locks in (floor above the entire signed era, nothing anchorable) it is a line per poll interval, default 60s, indefinitely. The counter repeating is arguably the point since the rate makes a stuck tenant visible; the log line is the part worth demoting or emitting once per floor value.

4. **`floorEngaged` is lost when a tenant throws.** The `catch` in `runAuditAnchorSweep` increments `errors` only, so a pass that engaged the floor and then failed downstream reports `tenantsFloorEngaged` unchanged. Errors are counted separately, so nothing is hidden, but the two signals disagree on that path.

Nit: `legacyFloor` exposes a raw value from a globally shared `bigserial`, so a customer can infer coarse fleet-wide write ordering. `lastAnchoredSeqTo` already exposed exactly that, so this adds no new class, and the value returned is the tenant's own row. Noting it only so it is on the record.

Nothing here is a correctness or security defect. The tenant scoping is intact, the gating is untouched, the refactor is provably behavior-preserving and compiler-enforced, and the change does what #2182's review asked for. Approving on the code. Merge and the `sec-review:approved` label remain the owner's call.

PR #2188, opened 2026-07-26
perf(db): partial index for the audit legacy-floor derivation + floor-log throttle (GAP-429 item 4)

RevealUIStudio:
## Summary

Closes out the review residual from #2185: the floor derivation ran unindexed on the anchors request path, and the floor-engaged log repeated every poll tick.

## What changed

- `audit_log` gains `audit_log_tenant_unsigned_idx` ON `(tenant, seq)` WHERE `signature IS NULL`. The healthy no-legacy case previously scanned the tenant's whole row set to return 0 on a no-store endpoint. The unsigned era is closed post-enforcement, so the index stays tiny. The Drizzle schema models the same index via `.where()`.
- Migration `0030` is hand-written idempotent SQL (`CREATE INDEX IF NOT EXISTS` with a partial predicate, following the 0008 precedent) with a `_custom.json` entry. A generate attempt re-emitted the pre-existing 0028/0029 snapshot drift on top of the stale 0027 snapshot, confirming the snapshot debt is shared, not new — noted in the entry.
- The floor-engaged log is throttled to once per tenant per process; the `revealui_audit_anchor_floor_engaged_total` counter and the `tenantsFloorEngaged` result field still report every pass.

## Tests (prove-red)

14/14 sweep tests green. Reverting only the sweep file to `origin/test` fails exactly the new throttle test with `expected [ ...(2) ] to have a length of 1 but got 2` (base logs on both passes). The partial index is exercised by every PGlite schema load in the suite and applied for real by the Drizzle Migrations CI job.

### DIFF
--- a/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
+++ b/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
@@ -13,7 +13,7 @@ import { type AuditEntry, type AuditRowSignerFn, DrizzleAuditStore } from '@reve
 import type { Database } from '@revealui/db/client';
 import { auditAnchors } from '@revealui/db/schema';
 import { eq } from 'drizzle-orm';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   createTestDb,
   type TestDb,
@@ -338,6 +338,41 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     expect(result.tenantsSkipped).toBe(1);
     expect(result.errors).toEqual([]);
     expect(await db.select().from(auditAnchors)).toHaveLength(0);
+  });
+
+  it('GAP-429: floor-engaged log fires once per tenant per process (metric every pass)', async () => {
+    // Unique tenant: the throttle Set is module-level, so tenants used by
+    // earlier tests in this file may already be marked as logged.
+    const tenant = 'acct_throttle_only';
+    const signedStore = new DrizzleAuditStore(db, canonicalSignerFn(priv));
+    const unsignedStore = new DrizzleAuditStore(db);
+    await signedStore.append(makeEntry({ id: 'th-s1', tenant }));
+    await unsignedStore.append(makeEntry({ id: 'th-u2', tenant }));
+
+    const { logger } = await import('@revealui/core/observability/logger');
+    const infoSpy = vi.spyOn(logger, 'info');
+    try {
+      const sweepOptions = {
+        db,
+        signer: cryptoSigner,
+        batchSize: 1,
+        canAnchorTenant: async () => true,
+        recordMeter: false,
+        now: () => new Date('2026-07-23T12:00:10.000Z'),
+      };
+      const first = await runAuditAnchorSweep(sweepOptions);
+      const second = await runAuditAnchorSweep(sweepOptions);
+      // Engagement stays observable in the result on EVERY pass...
+      expect(first.tenantsFloorEngaged).toBe(1);
+      expect(second.tenantsFloorEngaged).toBe(1);
+      // ...but the log line fires only once for the tenant.
+      const floorLogs = infoSpy.mock.calls.filter((call) =>
+        String(call[0]).includes(`legacy floor engaged tenant=${tenant}`),
+      );
+      expect(floorLogs).toHaveLength(1);
+    } finally {
+      infoSpy.mockRestore();
+    }
   });
 
   it('no-ops without a signer (unsigned mode)', async () => {
--- a/apps/server/src/jobs/audit-anchor-sweep.ts
+++ b/apps/server/src/jobs/audit-anchor-sweep.ts
@@ -258,6 +258,11 @@ export async function legacyUnsignedFloor(db: Database, tenant: string): Promise
 
 type TenantOutcome = 'inserted' | 'waiting' | 'skipped';
 
+// GAP-429: a tenant stuck in the no-anchors state (e.g. floor above the entire
+// signed era) re-engages the floor on every poll tick — log once per tenant
+// per process; the counter still increments every pass.
+const floorLoggedTenants = new Set<string>();
+
 async function anchorTenantBatch(
   db: Database,
   signer: Ed25519AuditRowSigner,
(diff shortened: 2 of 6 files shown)


RevealUIStudio:



Reviewed independently from the PR head (`4e89ad7`) against `origin/test` (`98d0cad`). I did not author this change. Same reviewer as the #2182 and #2185 verdicts; this builds my #2185 findings 1 and 3.

## What I verified

**Tests, run by me.**

- Green, PR head: `pnpm --filter server exec vitest run src/jobs/__tests__/audit-anchor-sweep.test.ts` to **14 passed (14)**.
- Red: reverted only `apps/server/src/jobs/audit-anchor-sweep.ts` to `origin/test` and ran `-t "floor-engaged log fires once"`. Fails with `AssertionError: expected [ …(2) ] to have a length of 1 but got 2`, which is the throttle being absent, not a shape mismatch. File restored to HEAD, `git status --porcelain` empty, `git diff HEAD` empty.
- `pnpm validate:migrations`: `31 SQL files, 31 journal entries, all checks passed`. I also ran the validator's own suite (`scripts/validate/__tests__/migration-journal.test.ts`, 25 passed) so a passing verdict from it means something. `db:assert-migration-count` needs `POSTGRES_URL` and is a deploy-time check, not runnable here.

**Migration correctness.** SQL and schema agree on all three axes: name `audit_log_tenant_unsigned_idx`, columns `(tenant, seq)`, predicate `signature IS NULL`. Journal entry is `idx: 30`, `version: "7"`, `when: 1785084298662` (monotonic against 0029's `1784700000000`), tag matching the filename. The `_custom.json` entry carries rationale plus the `missingSnapshot` / `snapshotDebtNote` pair the other hand-written entries use, and the validator accepts it.

Worth recording: `packages/test/src/utils/drizzle-test-db.ts` builds every PGlite test database by loading and applying `packages/db/migrations` in order. So the 14 green tests above already apply `0030` to a fresh database. The SQL is exercised, not just linted.

**Index shape, measured rather than reasoned.** I ran the exact migration statement against PGlite on a 40k-row table across four tenants plus a 200-row unsigned era, with `ANALYZE`, and took plans before and after:

| Query | Before | After |
|---|---|---|
| floor, tenant *with* a legacy era | Bitmap Heap Scan, 385 buffers, 3.26 ms | `Index Only Scan Backward using audit_log_tenant_unsigned_idx`, 2 buffers, 0.22 ms |
| floor, tenant with **no** unsigned rows (the case I flagged in #2185) | Bitmap Heap Scan, 383 buffers, `Rows Removed by Filter: 10000`, 1.87 ms | `Index Only Scan Backward`, 1 buffer, `Heap Fetches: 0`, 0.07 ms |

So yes: `(tenant, seq) WHERE signature IS NULL` is the right index, and Postgres takes the backward scan on the second key exactly as intended. Index size is 16 kB against a 3008 kB table, and it only ever covers the closed era, so it does not grow. Re-running the migration statement is a clean no-op, and `pg_indexes` afterward shows exactly one index, defined as `USING btree (tenant, seq) WHERE (signature IS NULL)`.

**Throttle.** `mFloorEngaged.inc()` sits outside the `if (!floorLoggedTenants.has(tenant))` guard, and `floorEngaged` is still returned on every pass, so neither the counter nor `tenantsFloorEngaged` is throttled. The new test asserts both directly (`first` and `second` each report 1) alongside the single log line, which is the right pair of assertions. Memory is bounded by the number of tenants that ever engage the floor, one short string each, and a tenant stops being added once it anchors. The unique-tenant choice in the test is correct and the comment explains why; if a future test reused `acct_throttle_only` earlier in the file this test would fail loudly rather than pass for the wrong reason, which is the safe direction.

## Findings (none blocking)

1. **The other request-path query is still unindexed, and it is the one prod will actually run.** `preFloorSignedCount` in `apps/server/src/routes/audit.ts` filters `signature IS NOT NULL`, the complement of this index's predicate, so it gets no benefit. In the same measured setup it stayed a Bitmap Heap Scan with `Rows Removed by Filter: 10175`, meaning a full scan of that tenant's rows. It runs precisely when `legacyFloor > 0`, which is prod's state (floor 1122). Net effect: this PR removes the scan for healthy tenants and leaves it for legacy-era tenants. Note the planner did not choose `audit_log_seq_idx` for the `seq <= floor` range even though the range was highly selective, so this is planner-dependent and worth an `EXPLAIN` against real prod volume rather than an assumption. The count is immutable once the era is closed, so caching it or adding a composite `(tenant, seq)` are both cheap paths.

2. **The index build takes a lock that blocks audit writes.** Plain `CREATE INDEX` holds a `SHARE` lock on `audit_log` for the duration of one full table scan, during which INSERTs block. That is a scan of the whole table, not just the 170 matching rows, and `audit_log` is written on every agent action. `CONCURRENTLY` is not an option inside the migration runner (drizzle wraps migrations in a transaction), so the chosen form is right for the file. Flagging it as a deploy note: if prod's `audit_log` is large, consider building this index manually with `CONCURRENTLY` before the migration runs, so the runner's `IF NOT EXISTS` finds it already present and no-ops.

3. **`tenantsFloorEngaged` never reaches the tick log.** The tick summary at `audit-anchor-sweep.ts:408-412` prints tenants, inserted, waiting, skipped, nullTenant, and errors, but not the new field. With the log line now throttled to once per process, the metric is the only per-tick evidence that the floor is engaged. Adding `floorEngaged=${r.tenantsFloorEngaged}` to that existing line restores per-tick log visibility at zero additional spam, which is the cheapest way to keep what the throttle traded away.

4. **`stopAuditAnchorSweep()` does not clear `floorLoggedTenants`.** A stop/start cycle in the same process keeps the suppression. Nothing in production depends on it, but clearing it there makes the module's lifecycle honest and the state easier to test.

5. **The Set is keyed by tenant, not by tenant and floor value.** If a floor ever changed, the new value would never be logged. Post-enforcement it cannot change, so this is theoretical, but `${tenant}:${floor}` is a one-token change that removes the caveat.

Nit: schema-versus-SQL equivalence here is human-verified, not machine-verified, because the drizzle snapshot is deliberately not regenerated (documented shared debt, and the PR is explicit that a generate attempt re-emitted the 0028/0029 drift). I confirmed the SQL by executing it and reading back `pg_indexes`; I did not independently render drizzle's own DDL for the schema-side `.where()`. That gap belongs to the snapshot debt, not to this PR.

Nothing here is a correctness or security defect. The migration is idempotent, validator-clean, applied by the test suite itself, and the index is empirically the right one for the query it targets. Approving on the code. Merge and any labels remain the owner's call.

Issue #828, opened 2026-05-12
fix(billing): gate automatic_tax on Stripe Tax active check


RevealUIStudio:
**Severity:** P0 — blocks `STRIPE_LIVE_MODE=true` (`phase:0` ethics gate).
**Source:** Pre-flip moral-readiness audit, 2026-05-11 (tracker #826).
**Audit-class:** billing.

## Problem

Every checkout session in `apps/server/src/routes/billing.ts:633,1401,1525,1657` passes `automatic_tax: { enabled: true }` unconditionally. If the live Stripe account hasn't activated Stripe Tax (per #407 it has not), Stripe returns `stripe_tax_inactive` and the customer sees a 500 on the first checkout. No boot-time check guards this.

## Fix

- Add `validateStripeTaxActive(stripe)` in `apps/server/src/lib/validate-startup.ts`: call `stripe.tax.settings.retrieve()`, require `status === 'active'` when `STRIPE_LIVE_MODE=true`.
- Add a `billing-readiness` cron alarm so post-boot dashboard-side deactivation surfaces immediately.
- Optionally: gate `automatic_tax` to `enabled: false` when the boot check failed but live-mode isn't on (dev/test parity).

## Acceptance

- [ ] Boot fails with a clear error when Stripe Tax inactive + live mode on
- [ ] Cron alarm fires if Stripe Tax is deactivated post-boot
- [ ] Documented in `docs/runbooks/stripe-live-flip.md` (or wherever pre-flip ops live)

Blocked-by #407 (owner-action: activate Stripe Tax on Live account).

PR #2200, opened 2026-07-26
feat(billing): warn at boot when Stripe Tax is active but flag is off

RevealUIStudio:
## Summary
- GAP-437: warns when this Stripe account's Tax Settings are active but `STRIPE_TAX_ENABLED` is off, so live Checkout sessions aren't silently collecting zero sales tax.
- Root incident (2026-07-26): prod collected a live $49 TN charge with zero sales tax because `STRIPE_TAX_ENABLED` was never set while Stripe Tax + a TN registration were both active. `routes/billing.ts:102` correctly gates `automatic_tax` on this flag (#828); the deployment just never flipped it.
- **The daily `billing-readiness` cron (`apps/server/src/routes/cron/billing-readiness.ts`) is now the authoritative control.** It runs on the same Vercel deployment that actually builds Checkout sessions and emails `REVEALUI_ALERT_EMAIL` on failure — this is the fix for the review's B1 finding (see below). The Vercel serverless API has no boot hook at all (cold starts run only module-level code, by design — see the comment above the production block in `index.ts`); there is no "wire it into the API's boot" option, which is why the daily cron is the right home rather than a workaround.
- `validateStripeTaxConfigAtStartup` in `apps/server/src/lib/validate-startup.ts` is kept as a **secondary, advisory** signal on the Fly worker's boot chain (and the local-dev boot chain, for symmetry) — early feedback on a worker redeploy, not proof the money path is safe. It is structurally fail-open: the fetch AND the warning emitter run inside one try/catch, so nothing in this function can ever reject a caller's boot chain, and it now runs last, after every fail-closed gate (`auditStorageSelfTest`, `runHostedLicenseCanary`, `initializeLicense`) instead of ahead of them.
- All Stripe calls on this path go through the shared `protectedStripe` wrapper (GAP-131: DB-backed circuit breaker + bounded timeout/retry) — added a `tax.settings.retrieve` method to `packages/services/src/stripe/stripeClient.ts` so this call site matches every other one in the repo instead of using a raw client with the SDK's ~80s default timeout.
- A Stripe API failure while checking tax settings never blocks boot (worker) and never counts as a cron failure/alert (an inability to check is not evidence of misconfiguration) — but it does now log one non-secret line (message only, never the error object) so a scope-denied key or persistent outage is visible instead of this control silently going dark.
- Registered `STRIPE_TAX_ENABLED` in the documented env matrix (`docs/ENVIRONMENT-VARIABLES-GUIDE.md`) and the Fly sync manifest skip-list (`scripts/sync/revvault-fly.toml` — it's a plain boolean, not a secret, so it has no vault path) so the Vercel `api` project and the Fly `revealui-worker` app have one documented place to stay in sync instead of drifting silently.

## Review status

**REVIEW-PENDING under guardrail 2 — `apps/server/src/lib/validate-startup.ts` is on the fleet security-sensitive path list. Do not merge until a fresh non-author APPROVE verdict is recorded** (checked via the fleet security checker).

This PR previously received a REQUEST-CHANGES verdict (3 blockers). Mapping below; awaiting a fresh review.

| Blocker | Fix | Test |
|---|---|---|
| **B1** — the control didn't observe the deployment that mints Checkout sessions (Fly worker only; Vercel API has no boot hook) | Moved the authoritative check into the daily `billing-readiness` cron, which runs on the Vercel deployment that builds Checkout sessions; kept the worker-boot check as a secondary signal and said so honestly in this PR body | New `billing-readiness cron — Stripe Tax flag (GAP-437)` describe block: live+active+flag-off alerts, flag-on silent, pending-status silent, API error tolerated (warning not failure), not-live-mode skipped entirely |
| **B2** — raw Stripe client (no `protectedStripe`, no timeout floor) sitting ahead of fail-closed boot gates, emitter not covered by the fail-open try/catch | Added `tax.settings.retrieve` to `protectedStripe`; `fetchLiveStripeTaxSettings` now calls it instead of `getStripe()` directly; moved the worker/dev-block call to the END of the chain (after `initializeLicense()`), `.catch(() => {})`'d at the call site; wrapped the ENTIRE validator body (fetch + emitter) in one try/catch, with the error-logging line itself also guarded so a broken stderr stream can't escalate | New tests: `fetchLiveStripeTaxSettings` calls `protectedStripe.tax.settings.retrieve` (not raw `getStripe`); "never throws even when the emitter itself throws (structural fail-open)" |
| **B3** — PR body leaked a private-repo script path and a private-repo PR link, failing the body gitleaks scan | Rewrote this section to reference "the fleet security checker" and "GAP-438 (internal tracker)" with no paths or repo links | N/A (body-only fix) |

Suggestions also taken: one-line non-secret error log (above), test coverage for `fetchLiveStripeTaxSettings`, structural fail-open on the emitter, and the env-registry entries above.

## Test plan
- [x] New/updated tests in `apps/server/src/lib/__tests__/validate-startup.test.ts`: proved RED first against the pre-fix code (3 failing: the updated API-error-log assertion, the emitter-fail-open test, and the `protectedStripe` call-site test), then implemented → all green.
- [x] New tests in `apps/server/src/routes/cron/billing-readiness.test.ts` (GAP-437 describe block): proved RED first (3 failing — no `stripe:tax-flag` check existed), then implemented → all green.
- [x] `pnpm --filter server exec vitest run src/lib/__tests__/validate-startup.test.ts src/routes/cron/billing-readiness.test.ts` — 142/142 passing (131 + 11; the original PR's 128 plus this rework's additions).
- [x] `pnpm --filter server exec vitest run src/__tests__/index.startup.test.ts` — 14/14 passing (boot-chain shape unaffected).
- [x] `pnpm --filter @revealui/services test` — 126 passed / 3 skipped, unaffected by the new `tax.settings.retrieve` wrapper method.
- [x] `pnpm turbo run typecheck --filter=@revealui/services --filter=server` — clean.
- [ ] Full pre-push gate — pending re-run before this is presented as merge-ready again.
- [x] Pre-existing unrelated failure in `webhook-downgrade-cap.test.ts` (503 vs 200), reproduced identically on a pristine `origin/test` scratch worktree with zero relation to this diff — tracked as GAP-438 (internal tracker), not part of this PR's scope.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


RevealUIStudio:

Non-author adversarial review of the full diff (4 files, +223/-1), the touched files in context, and the surrounding deployment topology. Tests were re-run and mutation-checked on the PR head in a scratch worktree.

### What holds up

- **No secret or key material reaches any log.** `emitStripeTaxFlagWarning` writes a static banner; the `catch` swallows the error object entirely rather than formatting it. Nothing derived from `STRIPE_SECRET_KEY` is emitted.
- **The guard set is correct and matches the established convention.** `SKIP_ENV_VALIDATION` / `NODE_ENV !== 'production'` / `detectDeploymentMode() !== 'hosted'` / `STRIPE_LIVE_MODE !== 'true'` is byte-for-byte the `validateBillingCatalogAtStartup` gate (`apps/server/src/lib/validate-startup.ts:690-693`). Stripe is genuinely never called in test mode.
- **Fail-open on API rejection is real, not just claimed**, and the diff is purely additive. No existing validator behavior is weakened.
- **The tests are not vacuous.** I ran `apps/server/src/lib/__tests__/validate-startup.test.ts` on the PR head: 128/128 pass, and the 10 new tests match the claim. Three mutations were all caught: deleting the `STRIPE_LIVE_MODE` guard (2 failures), dropping the `status === 'active'` condition (1 failure), removing the try/catch (1 failure). The red-then-green claim is credible.

### Blockers

**B1. The control does not observe the deployment that mints checkout sessions.**
`vercel.json` rewrites `/(.*)` to `/api`, and `apps/server/src/index.ts:1526` is the production block for that serverless handler. That is the process where `routes/billing.ts:102` reads `STRIPE_TAX_ENABLED` and builds Checkout sessions for customers. The new validator is deliberately excluded from it. The `index.ts` wiring added here sits inside the `NODE_ENV !== 'production' && NODE_ENV !== 'test'` dev block, so its own production guard makes it unreachable there. That leaves `worker.ts` as the only live call site, and the worker is a separate Fly app with separately-applied secrets.

Consequences, both in the exact shape of the incident this PR cites:
- Setting `STRIPE_TAX_ENABLED=true` on the API deployment fixes the money path but does not silence the worker banner, because the flag is not on the revvault-to-Fly sync manifest (`scripts/sync/revvault-fly.toml`) and not in `apps/server/src/lib/required-env.ts`. It appears in exactly one place in the repo: `routes/billing.ts:102`.
- Setting it on the worker alone silences the banner while checkout stays untaxed. The control then reports all clear over a live defect.

Additionally, per the fleet lane tracking the Fly migration, the `revealui-worker` machine was last recorded as stopped with the deploy blocked, so today this guard would execute in zero production processes. Please re-verify that against live infra rather than the doc, but the structural point stands either way.

**Suggested fix (extend before create):** `apps/server/src/routes/cron/billing-readiness.ts` already runs daily at 06:00 UTC on the Vercel deployment, already loads Stripe through `getServices()`, already validates billing env, and already emails `REVEALUI_ALERT_EMAIL` on failure. Adding the Tax Settings check there gives you the right process, the right env, a channel a human actually reads, and daily re-evaluation instead of once per deploy, at zero request-path cost.

**B2. The Stripe call bypasses the mandated resilience wrapper and sits ahead of two fail-closed gates in the boot chain.**
`fetchLiveStripeTaxSettings` calls `services.getStripe().tax.settings.retrieve()` raw. Every other Stripe call site in the repo goes through `protectedStripe` per GAP-131 (`routes/billing.ts`, `routes/webhooks.ts`, `routes/pricing.ts`, `routes/gdpr.ts`, `routes/marketplace.ts`), which supplies a 10s `withTimeout` and a DB-backed circuit breaker. The raw client inherits the SDK defaults instead: `timeout` 80000ms and `maxNetworkRetries` 2 in stripe@22.3.2, so a black-holed connection can hang for roughly four minutes.

In `worker.ts` the new step is inserted before `auditStorageSelfTest()` (the fail-closed audit integrity gate, ADR section 2a), `runHostedLicenseCanary()`, and `initializeLicense()`, while `serve()` binds the port synchronously and starts accepting traffic regardless of chain progress. So a slow Stripe endpoint delays the audit round-trip proof and license initialization for minutes on a process that is already serving. The docstring's "same secret key and pinned API version as every other Stripe call site" is accurate about the key and version but glosses over the wrapper every other site uses.

**Suggested fix:** move the advisory check to the end of the chain (after `initializeLicense()`), and either route it through `protectedStripe` (adding a `tax.settings.retrieve` method) or wrap it in an explicit short timeout.

**B3. The PR body leaks a private-repo path and is failing the body scan.**
`scan / Scan body via gitleaks` is RED on rule `revealui-private-repo-path` (see `.gitleaks.issues.toml:45`). The Review-status paragraph embeds a private fleet script path, and the test-plan section links a private-repo PR. This is a public repo, so please edit the body to remove both before merge. Note the check's own warning that editing does not remove content from GitHub edit history.

### Suggestions

- **The swallow is fully silent, which makes a broken control indistinguishable from a healthy account.** `tax.settings.retrieve` requires the Tax Settings read scope; a restricted key without it returns a permission error, the `catch` returns, and there is no warning and no log line ever again. Emit one line on the failure path (message only, no secrets) so the control's own health is observable.
- **`fetchLiveStripeTaxSettings` has no test coverage.** Every test injects a stub, so the `getServices()` null branch and the actual call shape are unexercised. A small test asserting the null-module branch throws (and is therefore absorbed by the caller) would close that.
- **Fail-open is asserted but not structurally enforced.** The try/catch wraps only `fetchTaxSettings()`. Anything thrown by `emitStripeTaxFlagWarning` would reject the chain and hit the `worker.ts` catch that calls `process.exit(1)`, which turns an advisory warning into a boot-kill path. Wrapping the whole body, or attaching `.catch(() => {})` at both call sites, makes the comment in `worker.ts:77-79` true by construction.
- **Register `STRIPE_TAX_ENABLED` somewhere.** As long as it lives only at `routes/billing.ts:102`, the API and worker deployments can diverge on it indefinitely. Adding it to the Fly sync manifest, or to a documented env matrix, is what actually prevents the recurrence.

### Nits

- The PR body describes the check as wired into `index.ts`, which reads as coverage it does not provide; the dev-block placement is symmetry only.
- The `import` of the new symbol sits mid-file in the test at roughly line 1159. Hoisting works, but the file's other imports are at the top.

Happy to re-review promptly once B1 through B3 are addressed. The underlying detection logic is sound and well tested; my objection is that the control is mounted on the wrong process and, as placed, can slow two fail-closed gates.

RevealUIStudio:

Re-review of the delta from the tree I first reviewed (`1aa678d5`) to current head (`16de1cd2`), non-author. **This resolves my standing REQUEST-CHANGES.** All three blockers are addressed, and I verified each structurally rather than by test name, plus mutation-tested the new tests.

### B1 — control now sits on the money path: RESOLVED

Verified end to end rather than by claim:

- `billing-readiness` is registered in the `JOBS` array in `apps/server/src/routes/cron/dispatch.ts`, and `dispatch` is the single cron entry in `apps/server/vercel.json` (`/api/cron/dispatch`, `0 6 * * *`). That runs in the serverless deployment that serves `/api/*`, which is the same process where `routes/billing.ts:102` reads `STRIPE_TAX_ENABLED` and builds Checkout sessions. The control and the defect now live in the same env.
- The alert path is the real one, not a parallel one: the check pushes `ok: false` into `results`, which flows into `failures`, which triggers both `sendCronFailureAlert` and the `sendEmail({ to: ALERT_EMAIL })` block, with `ALERT_EMAIL` resolving from `REVEALUI_ALERT_EMAIL`. The response also flips to 503. Identical treatment to every other hard-fail check in that file.
- `services` is already in scope at the insertion point (loaded once at section 4), and the check is correctly gated on `services && STRIPE_LIVE_MODE === 'true'`, so it cannot throw on a missing optional peer and never calls Stripe in test mode.
- The boot-time check is now honestly labelled secondary in all four places a reader could land: the function docstring, `worker.ts`, the `index.ts` dev block, and the cron's own inline comment. The docstring's "treat a worker-boot warning as an early heads-up, not proof the money path is safe" is exactly the right framing.

The env-registration suggestion also landed in three places: the docs env matrix, an explanatory non-required note in `required-env.ts`, and the Fly manifest skip-list with the divergence rationale spelled out.

### B2 — bounded call, correct ordering, structural fail-open: RESOLVED

- `tax.settings.retrieve` was added to `protectedStripe` in `packages/services/src/stripe/stripeClient.ts` and goes through `callWithResilience`, so it inherits the 10s per-attempt timeout and the DB-backed breaker instead of the SDK's 80s default. `fetchLiveStripeTaxSettings` calls the wrapper; a test asserts `getStripe` is never invoked, which pins the property rather than describing it.
- Ordering is fixed in both chains: the check now runs inside the terminal `.then((tier) => ...)`, after `auditStorageSelfTest()`, `runHostedLicenseCanary()` and `initializeLicense()`. Nothing fail-closed sits behind it any more.
- Fail-open is now structural, not incidental. The whole body (fetch, condition, emitter) is inside one try/catch, the error-path stderr write is itself wrapped, and both call sites add `.catch(() => {})`. Three independent layers, any one of which is sufficient.
- I mutation-tested this rather than trusting the test name: rewiring the function so the emitter path escapes the catch fails 3 tests. The structural-fail-open test is real.

### B3 — body scrubbed: RESOLVED

`scan / Scan body via gitleaks` is green, and a direct grep of the current body returns zero matches for private-repo paths or links. The only remaining red check is `Security review gate`, which is this gate itself and should clear once the verdict is recorded.

### Test verification

Ran on the PR head in a scratch worktree: 142/142 across both touched test files, with `billing-readiness.test.ts` at 11/11 (5 pre-existing price-parity plus 6 new tax). The 11/11 claim checks out.

Four mutations, all caught:
- cron ignores `status === 'active'` → 1 failure
- cron downgrades the drift result to `ok: true` → 2 failures (confirms the tests bind the alert path, not just the string)
- cron drops the `STRIPE_LIVE_MODE` guard → 1 failure
- boot validator's emitter escapes the try/catch → 3 failures

My earlier coverage gap is closed too: `fetchLiveStripeTaxSettings` now has both the missing-module branch and the protectedStripe-not-getStripe assertion, and the error-path test pins that exactly one non-secret line is written and that it is not the loud banner.

### Non-blocking notes for later

1. **Dispatcher time budget.** All jobs run sequentially inside one 30s function with no per-job budget guard, and this adds a Stripe round trip that can consume roughly 30s on its own in the worst case (3 attempts at 10s plus backoff), ahead of `sweep-grace-periods`, which is the job that revokes access from non-payers. This is a pre-existing class rather than something introduced here, since the price-parity loop already makes up to three such calls, but the margin got thinner. A single-attempt variant for advisory checks, or a deadline guard in the dispatcher, would be the durable fix.
2. **Persistent lookup failure is quiet by design.** A scope-denied key produces a `warnings` entry, which only reaches email when some other check fails; otherwise it is a daily `logger.warn`. The reasoning in the comment is sound and I am not asking for a change now, but if this control becomes load-bearing, escalating N consecutive warning days into a failure would close the last blind spot.
3. Per the fleet lane tracking the Fly migration, the worker was last recorded stopped, so the secondary boot check may execute nowhere today. That no longer affects the verdict now that the authoritative control is the cron, but it is worth knowing the secondary signal may be dormant.

Good rework. The control is on the right process, bounded, correctly ordered behind the fail-closed gates, and the tests hold under mutation.

PR #2201, opened 2026-07-26
feat(license): boot at Free tier on unlicensed plain self-host (GAP-436)

RevealUIStudio:
## Summary

Owner ruling 2026-07-26 (GAP-436): on the **plain** self-host path, booting with **no license key** now runs at Free (OSS) tier instead of refusing to start. This closes the wall the GAP-430 Railway marketplace template hit (its README previously said "there is currently no license-free 'just try it' mode for a container deploy").

**HARD SCOPE GUARD, unchanged:** RevForge-stamped Fleet kits (forge mode) keep the hard license requirement. Nothing relaxes unless a deployment explicitly sets `REVEALUI_ALLOW_UNLICENSED_SELF_HOST=true` — RevForge-stamped kits never set this flag, so their enforcement is the exact same code path as before this PR.

### Audit findings (file:line, before this PR)

- `packages/core/src/license.ts` `initializeLicense` (:556-568) already resolved to Free tier for a missing key **at the singleton level** — this was never the wall.
- The actual hard boot refusal lives in **two duplicated gates**, both keyed off `detectDeploymentMode(env) !== 'hosted'` (i.e. "forge" = any self-hosted deployment, stamped or plain — there was no existing signal to tell them apart):
  1. `apps/server/src/lib/validate-startup.ts` `validateStartup` (:140-159 `REQUIRED_IN_PRODUCTION_FORGE`, :278-286 dispatch) — required `REVEALUI_LICENSE_KEY` + `REVEALUI_LICENSE_PUBLIC_KEY` presence in `NODE_ENV=production` boots (this is what `worker.ts`, the Docker/Railway entrypoint, runs first).
  2. `apps/server/src/lib/validate-startup.ts` `validateLicenseAtStartup` (:571-587, pre-PR) — threw `LICENSE VALIDATION FAILED: REVEALUI_LICENSE_KEY is required for RevForge deployments` whenever the key was absent, in ANY `NODE_ENV`.
  3. **A third, independently-duplicated copy** of the same enforcement in `apps/admin/src/instrumentation.ts` (:34-93, pre-PR) — flagged in its own TODO as "the ~25-line duplication is preferable to a new package boundary." Missed by GAP-430's earlier audit; found and fixed here, since a fix that only touched apps/server would still crash-loop the Railway template's `admin` service.
- `forge/stamp.sh` (revforge repo) never sets `REVEALUI_DEPLOYMENT_MODE` — a genuinely-stamped kit is detected purely via **absence** of `REVEALUI_LICENSE_PRIVATE_KEY`, the identical signal a plain self-host boot produces. There is no existing "this kit was RevForge-stamped" marker in the env.

### Why a new explicit opt-in flag, not a new deployment-mode value

Because forge-mode detection cannot currently distinguish "stamped kit whose license got stripped/never issued" from "plain self-host boot with no license, by design," relaxing the *default* fallback would silently weaken enforcement for every self-hosted deployment — exactly what the HARD SCOPE GUARD forbids. Widening `REVEALUI_DEPLOYMENT_MODE`'s enum (`z.enum(['hosted','forge'])` in `packages/config/src/schema.ts:127`) to a third value would also touch every hosted/forge dispatch point across auth, sign-up, chat routes, and admin instrumentation — a much larger blast radius than this gap asked for.

Instead: a single new, narrowly-scoped, explicit env var — `REVEALUI_ALLOW_UNLICENSED_SELF_HOST=true` — that a deployer/template must deliberately set. Nothing changes for anyone who doesn't know it exists. Concretely:

- `validateStartup`'s `REQUIRED_IN_PRODUCTION_FORGE` list drops `REVEALUI_LICENSE_KEY` / `REVEALUI_LICENSE_PUBLIC_KEY` **only** when `mode === 'forge' && REVEALUI_ALLOW_UNLICENSED_SELF_HOST === 'true'`. Every other forge requirement (KEK, SECRET, AUDIT_SIGNING_KEY, CORS_ORIGIN, etc.) is untouched.
- `validateLicenseAtStartup` skips its throw **only** when the key is completely absent AND the flag is `'true'` — logging `no license key — running Free (OSS) tier`. A key that IS supplied is still fully verified (signature, expiry, domain binding) regardless of the flag, so the flag can never be used to downgrade a genuinely stamped kit whose key was stripped as an attack.
- The identical relaxation was mirrored into `apps/admin/src/instrumentation.ts`'s duplicated gate, for the same reason the duplication exists at all (admin boots the same Forge check independently).

### What did NOT change

- `detectDeploymentMode` / `DeploymentMode` type / the hosted-vs-forge dispatch used elsewhere (auth, billing, sign-up, chat routes) — zero changes.
- Any check unrelated to license material (KEK, SECRET, audit signing key, Stripe, etc.).
- Verification of a license key that IS present — invalid/expired/wrong-keypair still throws exactly as before, flag or no flag.
- `apps/admin/src/lib/utils/env-validation.ts`'s separate production-required list (SESSION_COOKIE_DOMAIN, Stripe price IDs, email transport) — unrelated to licensing, out of scope for this gap.

## Test plan (red → green, both proven by stash/revert + rerun)

**apps/server** (`apps/server/src/lib/__tests__/validate-startup.test.ts`, 127/127 passing):
- [x] `validateStartup` — forge mode with `REVEALUI_ALLOW_UNLICENSED_SELF_HOST=true` and no license keys: does NOT throw (new; proven red on pre-PR code: threw `Missing production-required env vars: REVEALUI_LICENSE_KEY, REVEALUI_LICENSE_PUBLIC_KEY`).
- [x] **REGRESSION GUARD** — forge mode, flag unset, no license keys: still throws `forge mode.*REVEALUI_LICENSE_KEY` (pre-existing test, unchanged, still green).
- [x] **REGRESSION GUARD** — flag set, but KEK/SECRET/AUDIT_SIGNING_KEY still missing: still throws (flag only relaxes license material).
- [x] `validateLicenseAtStartup` — no key + flag=true: resolves undefined, logs `no license key — running Free (OSS) tier` (new; proven red: pre-PR code rejected with `LICENSE VALIDATION FAILED`).
- [x] **REGRESSION GUARD** — no key + flag unset: still rejects `REVEALUI_LICENSE_KEY is required for RevForge` (pre-existing assertion, unchanged).
- [x] **REGRESSION GUARD** — flag=true but a key IS supplied and it's invalid: still rejects `REVEALUI_LICENSE_KEY is invalid` (flag never bypasses verification of a present key).
- [x] Flag has no effect in hosted mode (already a no-op).

**apps/admin** (new file `apps/admin/src/__tests__/instrumentation-license-gate.test.ts`, 4/4 passing, proven red against pre-PR `instrumentation.ts` via git-stash):
- [x] `register()` boots (no `process.exit`) with no key + flag=true, logs the same boot message.
- [x] **REGRESSION GUARD** — flag unset: still calls `process.exit(1)` with the `REVEALUI_LICENSE_KEY is required for RevForge` message.
- [x] **REGRESSION GUARD** — flag=true, key present but invalid: still exits(1) with `REVEALUI_LICENSE_KEY is invalid`.
- [x] Hosted mode: no-op regardless of flag.

Also ran clean: `apps/server/src/lib/` full suite (459/459), `packages/core` license + deployment-mode suites (98/98, untouched), `scripts/sync/__tests__/prod-required-coverage.test.ts` (3/3, untouched), `tsc --noEmit` on both `server` and `admin` (no new errors — remaining errors are pre-existing unbuilt-workspace-package noise unrelated to this change), Biome clean, full `pnpm gate` pre-push hook (Security Gate + CI Gate) green.

## Docs updated

- [`docs/ENVIRONMENT-VARIABLES-GUIDE.md`](docs/ENVIRONMENT-VARIABLES-GUIDE.md) — new `REVEALUI_ALLOW_UNLICENSED_SELF_HOST` row under License.
- [`deployment/railway/README.md`](deployment/railway/README.md) — "Who this is for" + template now documents the Free-tier no-license path.

## Security review — REVIEW-PENDING under guardrail 2

This touches the boot-time license enforcement gate (`apps/server/src/lib/validate-startup.ts` + `apps/admin/src/instrumentation.ts`), a guardrail-2 security-sensitive surface (`security-pr-review-check.js --diff origin/test` confirms). **Do not merge until a non-author APPROVE verdict is recorded on this PR** (per `dup-work-and-review-gates.md`). I am not presenting this as merge-ready — the next step is review, not merge.


RevealUIStudio:

Non-author adversarial review. Reviewed at `7101912e0` in a scratch worktree with workspace packages built; all claims below are empirically verified, not read off the PR body.

**The security design is sound and I could not break it.** The one blocker is a docs/wiring inconsistency that ships the PR's own advertised flow broken. It fails *closed*, so it is not a vulnerability — but it defeats the gap's purpose and is a two-line fix.

---

### Blocker

**1. The Railway template's Free-tier path crash-loops `admin` if you follow the README.**

The PR body correctly argues that a server-only fix "would still crash-loop the Railway template's `admin` service" — and then the docs never tell the operator to set the flag on `admin`:

- [`deployment/railway/README.md:21-22`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/deployment/railway/README.md#L21-L22) — "set `REVEALUI_ALLOW_UNLICENSED_SELF_HOST=true` on **the `api` service**".
- [`deployment/railway/README.md:190-217`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/deployment/railway/README.md#L190-L217) — neither the `api` nor the `admin` env table has a `REVEALUI_ALLOW_UNLICENSED_SELF_HOST` row, and both still list `REVEALUI_LICENSE_KEY` / `REVEALUI_LICENSE_PUBLIC_KEY` as **required**.
- [`docs/ENVIRONMENT-VARIABLES-GUIDE.md:285`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/docs/ENVIRONMENT-VARIABLES-GUIDE.md#L285) — scope column says `api`.

`admin`'s gate is independent ([`apps/admin/src/instrumentation.ts:50`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/apps/admin/src/instrumentation.ts#L50)): no `REVEALUI_LICENSE_PRIVATE_KEY` → forge mode; no `REVEALUI_LICENSE_KEY`; flag unset on that service → `else` branch → `process.exit(1)`. Railway envs are per-service (the README itself says "Same value as `api`'s" for every shared var), so an operator following the README literally gets a booting `api` and a crash-looping `admin`.

Fix: say "on **both** the `api` and `admin` services", add the flag row to both service tables, mark the two license vars optional-when-flag-set, and widen the env-guide scope to `api, admin`.

---

### What I attacked and could not break (evidence)

**Present-but-invalid key still throws — verified, not just asserted.** Mutation-tested rather than trusting the suite name. I mutated the source five ways and re-ran; every relaxation is caught:

| Mutation | Result |
|---|---|
| A — `validateStartup` drops the flag requirement (forge always relaxes) | **5 failed** / 122 |
| B — `validateLicenseAtStartup` drops the flag requirement | **3 failed** / 124 |
| C — flag also bypasses a *present-but-invalid* key | **1 failed** / 126 |
| D — admin drops the flag requirement | **1 failed** / 3 |
| E — admin flag bypasses a present-but-invalid key | **1 failed** / 3 |

Baseline green: `validate-startup.test.ts` 127/127, `instrumentation-license-gate.test.ts` 4/4. The regression guards genuinely bite; a future session that relaxes forge mode will break CI.

**Flag parsing is strictly fail-closed.** Probed `'TRUE'`, `'True'`, `'1'`, `'yes'`, `' true '`, `'true '`, `''` against the real `validateLicenseAtStartup` — every one still throws `REVEALUI_LICENSE_KEY is required`. Only exact `'true'` relaxes.

**Whitespace key does not slip through.** `REVEALUI_LICENSE_KEY='   '` + flag=true is truthy, so it takes the verification path and throws `REVEALUI_LICENSE_KEY is invalid, expired beyond grace, ...`. Empty-string key is treated as absent (→ Free), which is behaviorally identical to unset and not a new route.

**No hosted-mode leak.** `plainSelfHostOptIn` is gated on `mode === 'forge'` ([`validate-startup.ts:282`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/apps/server/src/lib/validate-startup.ts#L282)) and the hosted branch of the ternary is evaluated first; `validateLicenseAtStartup` returns early on non-forge; the admin gate is inside `=== 'forge'`. `REQUIRED_IN_PRODUCTION_HOSTED` never contained `REVEALUI_LICENSE_KEY`/`_PUBLIC_KEY` anyway ([`required-env.ts:36`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/apps/server/src/lib/required-env.ts#L36) has only `_PRIVATE_KEY`). Probed explicit `MODE=hosted` + flag=true: hosted's required list is unaffected. Flag also does not relax non-license forge requirements — probed `REVEALUI_KEK` still throws.

**Gate inventory is complete — no fourth copy missed.** Grepped every `REVEALUI_LICENSE_KEY` / `detectDeploymentMode` / `LICENSE VALIDATION FAILED` site. The forge boot gates are exactly the three the PR names, and the relaxation is mirrored identically. The others are correctly untouched: `license-canary.ts:47` is hosted-only; `revforge-license.ts` is the mint side; `packages/core/src/license.ts:556` already resolved Free on an absent key; `PoweredByRevealUI` correctly shows the badge with no key.

### On the stamped-kit dodge — honest assessment

A stamped-kit operator *can* delete the baked key and set the flag. I don't think that's a material regression:

- The boot gate was never the revenue control. `initializeLicense()` has always returned `'free'` for an absent key ([`license.ts:556-568`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/packages/core/src/license.ts#L556-L568)); entitlement is enforced per-feature off that tier. An operator who takes this route lands on Free and loses exactly the Pro/Enterprise features they paid for. That is the correct outcome, not a bypass.
- Env is operator-controlled in self-host, and the repo is source-visible. The old gate stopped the *naive* "just remove the license" case; it never stopped a determined operator.
- Verification of a supplied key — signature, expiry, keypair match, domain binding — is untouched, so the flag can't downgrade a kit whose key was stripped as an attack while claiming to still be licensed.

What genuinely changes is that an expired/revoked kit can now be kept alive at Free instead of hard-stopping, and nothing distinguishes "plain self-host Free" from "formerly-licensed kit that dropped to Free". That's a revenue-*observability* gap, not a vulnerability — worth a follow-up gap (a boot-log/telemetry marker when the flag is honored), not a change here.

### Non-blocking suggestions

- [`apps/admin/src/instrumentation.ts:63`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/apps/admin/src/instrumentation.ts#L63) — the new `await import('@revealui/core/observability/logger')` sits **outside** any `try`, in a block whose own comment says instrumentation must never throw (the pre-PR code kept its only dynamic import inside a `try`). The module resolves fine today, so this is theoretical, but wrapping it (or reusing the `Promise.all` import below) preserves the block's stated invariant.
- Follow-up gap for the TODO at [`instrumentation.ts:46`](https://github.com/RevealUIStudio/revealui/blob/7101912e0/apps/admin/src/instrumentation.ts#L46): this PR is the second time the "~25-line duplication" had to be edited in lockstep across two apps. The duplication held fine here, but the next edit is the one that drifts.

Re-request review once the README/env-guide `admin` scope is corrected; the code itself needs no change for me to approve.

RevealUIStudio:

Re-review of the delta only: `7101912e0..c2a91d0a7` ("docs(railway): flag must be set on both api and admin services"). Non-author.

### Blocker resolved

All three sites I flagged are corrected, and the fix explains the *why* rather than just the *what*:

- [`deployment/railway/README.md:19-24`](https://github.com/RevealUIStudio/revealui/blob/c2a91d0a7/deployment/railway/README.md#L19-L24) — now "**both the `api` and `admin` services**", with the reason inline: "Railway env vars are per-service; the admin app runs its own copy of the license gate and will crash-loop without the flag." That last clause is exactly the failure mode I reproduced, so a future reader who is tempted to drop the flag from one service knows what happens.
- [`README.md:192-210`](https://github.com/RevealUIStudio/revealui/blob/c2a91d0a7/deployment/railway/README.md#L192-L210) (`api` table) — both license vars moved to "required unless running Free tier", plus a new `REVEALUI_ALLOW_UNLICENSED_SELF_HOST` row that restates the both-services requirement *and* the invalid-key property.
- [`README.md:211-226`](https://github.com/RevealUIStudio/revealui/blob/c2a91d0a7/deployment/railway/README.md#L211-L226) (`admin` table) — same treatment, with its own flag row.
- [`docs/ENVIRONMENT-VARIABLES-GUIDE.md:285`](https://github.com/RevealUIStudio/revealui/blob/c2a91d0a7/docs/ENVIRONMENT-VARIABLES-GUIDE.md#L285) — scope column `api` → `api, admin`, plus "Must be set on BOTH the api and admin processes".

An operator following the README now lands on a working Free-tier deploy on both services.

### Nothing new appeared

`git diff --name-only 7101912e0..c2a91d0a7` is **two `.md` files and nothing else** — `apps/`, `packages/`, and `scripts/` are byte-identical to the tree I reviewed. So every property I established last round still holds unchanged, without needing to re-run anything:

- Present-but-invalid key still throws (mutation-tested, 5/5 relaxations caught).
- Flag is strictly fail-closed on value (`'TRUE'`, `'1'`, `' true '`, `''` all still refuse).
- Whitespace key takes the verification path and throws.
- No hosted-mode leak; flag does not relax non-license forge requirements.
- Gate inventory complete — the three forge gates are mirrored identically, no fourth copy.
- Baseline suites green at 127/127 and 4/4.

I also checked the new prose against the code rather than taking it at face value: "the admin app runs its own copy of the license gate" matches [`apps/admin/src/instrumentation.ts:50`](https://github.com/RevealUIStudio/revealui/blob/c2a91d0a7/apps/admin/src/instrumentation.ts#L50), and "a present-but-invalid license key still fails boot regardless of this flag" matches the behavior I verified empirically. No new claims that the code doesn't support.

### Nit (do not hold the merge for this)

The two license rows in the `api` table say "Omit for the Free-tier path **below**" — the Free-tier path is actually §"Who this is for" *above* (line 15); there is no Free-tier section below. The flag row sits two rows down in the same table, so nobody gets lost. Fix it if you touch the file again.

The revenue-observability follow-up from my first review still stands as a separate gap (nothing distinguishes "plain self-host Free" from "formerly-licensed kit that dropped to Free" in the boot logs) — out of scope here, not a merge condition.

Cleared from my side. Merge disposition and the `sec-review:approved` label remain the owner's; I have applied no label and performed no merge.

PR #2202, opened 2026-07-26
fix(apify-actor): resolve guardrail-2 blockers on GAP-431 receipt actor

RevealUIStudio:
## Summary

Fix-forward on the REQUEST-CHANGES verdict recorded on #2198 after it merged ahead of review: https://github.com/RevealUIStudio/revealui/pull/2198#issuecomment-5085062163

Each blocker below has a failing test added first (reproducing the reviewer's finding), then the fix, then green.

### Blocker 1 — receipt provenance (`src/receipt/sign.ts`, `src/receipt/verify.ts`, `src/receipt/signable.ts`, `src/types.ts`, `src/main.ts`)

**Finding:** the signed payload was `{actionLog, algorithm, timestamp}` with no binding to the actor/run/build, so anyone could mint a self-consistent, "valid" receipt for a run that never happened.

**Fix:** `RunContext` (`actorId`/`actorRunId`/`actorBuildId`, sourced from `Actor.getEnv()`, `null` for local runs) is now inside the JCS-canonicalized signed payload and the `Receipt` schema, so it cannot be swapped after signing. `main.ts` publishes the receipt to the run's own `Actor.pushData`/`Actor.setValue` dataset/KV, a location only the real running actor can write to. README rewritten to state plainly: standalone verification proves integrity (tamper-evidence), not provenance; provenance requires cross-checking `receipt.actorRunId` against that run's own platform record.

**Test:** `src/__tests__/receipt-provenance.test.ts` (red before: fields didn't exist; green after: fields are embedded, signed, and tamper-checked).

### Blocker 2 — verify-what-you-read (`src/types.ts`)

**Finding:** `ActionLogEntrySchema`/`ReceiptSchema` were non-strict, so Zod silently stripped an injected unknown key during `safeParse`. The signature check covered the stripped data; every consumer reads the caller's original, unstripped object, so an injected field survived into what gets rendered while `verifyReceipt` still said `valid: true`.

**Fix:** both schemas are now `.strict()`.

**Test:** `src/__tests__/receipt-unknown-key-injection.test.ts`, reproducing the exact injected-field attack (red before: `valid: true`; green after: `valid: false`).

### Blocker 3 — SSRF guard (`src/agent/tools.ts`)

**Finding:** `isBlockedHost` compared against `url.hostname` with no bracket-stripping (so its own `'::1'` check could never match WHATWG's `[::1]`), no IPv6 handling at all, and single-address checks instead of full ranges. 8 of the reviewer's 13 probe targets slipped through.

**Fix:** rewrote the guard — strips IPv6 brackets, parses and range-checks both IPv4 and IPv6 (including IPv4-mapped addresses like `::ffff:127.0.0.1`), covers full CIDR ranges (127.0.0.0/8, 169.254.0.0/16, 100.64.0.0/10, fc00::/7, fe80::/10, etc.) instead of single hardcoded addresses, and resolves DNS for non-literal hostnames so a public name with a static record pointed at a blocked address (the `*.nip.io` case) is also blocked. True DNS-rebinding (the answer changing between check and fetch) remains out of scope and is now stated honestly in the README rather than blocked.

**Test:** `src/__tests__/tools.test.ts` — all 13 of the reviewer's probe targets as a parameterized table (red before: 8 pass through; green after: all 13 blocked).

### Hygiene — `@revealui/ai` optional dependency (`src/main.ts`, new `src/agent/load-llm-client.ts`)

**Finding:** declared in `optionalDependencies` but hard-imported at `main.ts:1`, so the actor couldn't start without it even in `verify-receipt` mode, which never needs it.

**Fix:** extracted to `loadLLMClient()`, a lazy dynamic `import()` with a clear, actionable error if it can't resolve. `verify-receipt` mode is now genuinely unaffected by whether `@revealui/ai` is installed.

**Test:** `src/__tests__/load-llm-client.test.ts` (happy path + mocked load failure).

### Guardrail-2 list widening

Added `packages/apify-actor-governed-run/` to `scripts/validate/security-paths.shared.json` (GAP-404 canonical source — audited; it is NOT under `packages/harnesses/`, that was the reviewer's placeholder guess) so this whole package is caught by guardrail-2 mechanically on future changes. Confirmed working: `security-pr-review-check.js --diff origin/test` now classifies this branch as security-sensitive.

## REVIEW-PENDING under guardrail 2

**The standing REQUEST-CHANGES on #2198 resolves only when a fresh APPROVE verdict is recorded on THIS PR.** Do not merge on green CI alone — CI passing is not a recorded verdict (that is the exact #1711 lesson this fix inherits). This PR should be reviewed by a non-author session before merge.

## What I could not fully address from the verdict

- **Suggestion (not a blocker): unbounded tool-calls-per-step / unbounded `verify-receipt` payload size / raw `err.message` echoed to OUTPUT** — the reviewer's "Suggestions" section items were left as-is; they were not blockers and are reasonable follow-up gaps, not addressed in this fix-forward PR to keep the diff focused on the three numbered blockers + the hygiene item the task specified.
- **True DNS-rebinding (TCP connection pinning)** — explicitly out of scope per the reviewer's own tiered guidance ("if that is more than a v0.1 scaffold should carry..."); the static-record bypass class (the actual reviewed bug) is fully closed, and the residual gap is now stated honestly in the README rather than silently left as a broader disclaimer than necessary.
- **Pre-existing, unrelated gate failure**: `pnpm gate --changed` on this branch fails one check, "Contracts OpenAPI mirror drift" (`packages/openapi/contracts.openapi.json`). Verified this exists identically on a clean, untouched checkout of `origin/test` before any of my changes (already tracked as GAP-436). Not touched by this PR.

## Test plan

- [x] `receipt-provenance.test.ts` — red before fix, green after
- [x] `receipt-unknown-key-injection.test.ts` — red before fix, green after
- [x] `tools.test.ts` 13-target SSRF probe table — red before fix (8/13 slip through), green after (13/13 blocked)
- [x] `load-llm-client.test.ts` — new coverage for the lazy-import hygiene fix
- [x] Full package suite: 57/57 green (`pnpm --filter @revealui/apify-actor-governed-run test`)
- [x] `pnpm --filter @revealui/apify-actor-governed-run typecheck` clean
- [x] `pnpm --filter @revealui/apify-actor-governed-run lint` clean
- [x] `pnpm gate --changed` green except the pre-existing, unrelated GAP-436 drift noted above
- [x] `security-pr-review-check.js --diff origin/test` now correctly flags this PR as security-sensitive (confirms the list widening works)

### DIFF
--- a/packages/apify-actor-governed-run/README.md
+++ b/packages/apify-actor-governed-run/README.md
@@ -3,26 +3,41 @@
 If an agent did it, there's a receipt.
 
 This actor runs an AI agent task for you and hands back both the result and a
-cryptographically signed, offline-verifiable receipt of every action the
-agent took: every model call and every tool call, in order, with a signature
-you can check without ever talking to us again.
+cryptographically signed receipt of every action the agent took: every model
+call and every tool call, in order, with a signature you can check without
+ever talking to us again.
 
 ## What you get
 
 - **A completed task.** Give the actor a task and your own LLM API key
   (Anthropic or OpenAI), and it runs an agent loop until the task is done or
   it hits a step limit.
-- **A signed receipt.** Every model call and tool call is recorded into an
-  action log, and the whole log is signed with a fresh Ed25519 key generated
-  for that run. The public key travels with the receipt, so anyone can verify
-  it offline with nothing but the receipt itself.
+- **A signed, tamper-evident receipt.** Every model call and tool call is
+  recorded into an action log, and the whole log, together with the
+  identifiers of the Apify run that produced it, is signed with a fresh
+  Ed25519 key generated for that run. Anyone can check the signature offline
+  with nothing but the receipt itself, and confirm that nothing in it changed
+  after signing.
+- **A platform-attributable run record.** The actor also publishes the
+  receipt to that run's own dataset and key-value store on Apify, the same
+  places the rest of the run's output lives. Nobody but the actual running
+  actor process can write there for that run, so a verifier who wants proof
+  the run really happened, not just that the receipt is internally
+  consistent, can fetch that record from the Apify API for the `actorRunId`
+  embedded in the receipt and confirm it matches.
 - **A free way to check any receipt.** Run this actor again in
   "verify-receipt" mode with a receipt you already have, and it tells you
   whether the signature is valid. This mode is free.
 
-What the receipt is: a cryptographically signed and offline-verifiable record
-of the actions an agent took. What it is not: a compliance certification of
-any kind. This actor makes no SOC 2, ISO, or other compliance claim.
+**What a receipt proves, and what it doesn't.** Standalone verification
+(`verifyReceipt`, or "verify-receipt" mode) proves the action log was not
+altered after signing. It does not, by itself, prove the run happened on
+Apify at all. Anyone can generate their own Ed25519 keypair and sign a
+fabricated action log claiming any `actorRunId` they like, and that receipt
+will still verify. Proving provenance, that a specific receipt corresponds to
+a run RevealUI's actor actually executed, needs the extra step above: checking
+`receipt.actorRunId` against that run's own dataset or key-value record on
+Apify. This actor makes no SOC 2, ISO, or other compliance claim either way.
 
 ## Input
 
@@ -90,13 +105,22 @@ price-only adjustment. Example `PUT` body shape (see
 
 1. Every model call and tool call the agent makes is appended to an action
    log, in order, along with a final entry recording the run's output.
-2. The action log is canonicalized with RFC 8785 JSON Canonicalization (the
-   same canonicalizer RevealUI's own internal audit log uses, so the byte
-   representation is deterministic regardless of key order) and signed with a
-   fresh Ed25519 keypair generated for that run.
-3. The receipt embeds the action log, the signature, the public key, the
-   algorithm, and a timestamp. Nothing about verifying it depends on
-   RevealUI's infrastructure or on this actor still existing.
+2. The action log, together with the run's `actorId`, `actorRunId`, and
+   `actorBuildId` (read from `Actor.getEnv()`; all three are `null` for a run
+   outside the Apify platform, such as local `pnpm dev`), is canonicalized
+   with RFC 8785 JSON Canonicalization (the same canonicalizer RevealUI's own
+   internal audit log uses, so the byte representation is deterministic
+   regardless of key order) and signed with a fresh Ed25519 keypair generated
+   for that run.
+3. The receipt embeds the action log, the run identifiers, the signature, the
+   public key, the algorithm, and a timestamp. Checking the signature depends
+   on none of RevealUI's infrastructure and none of this actor still existing.
+   Checking provenance instead of just integrity does depend on the run's own
+   record still being retrievable from Apify.
+4. The actor publishes that same receipt to the run's dataset
+   (`Actor.pushData`) and key-value store (`Actor.setValue('OUTPUT', ...)`),
+   both of which only the actual running actor process can write to for that
+   run.
 
 ```ts
 import { verifyReceipt } from '@revealui/apify-actor-governed-run';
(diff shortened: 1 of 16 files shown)


RevealUIStudio:

Re-attacked by the same non-author session that raised the three blockers on #2198, using the original methods against `7cd00fef1` in a clean worktree, not by reading the diff and trusting it. **All three blockers are genuinely closed.**

**This is the resolving verdict for the standing REQUEST-CHANGES on merged #2198.** That earlier verdict is hereby resolved by this APPROVE.

Note the `Security review gate` FAILURE on this PR is expected and correct: this PR adds `packages/apify-actor-governed-run/` to `scripts/validate/security-paths.shared.json`, so the gate immediately caught its own PR and now demands a recorded verdict. This comment is that verdict. Per disposition discipline I have applied no label and merged nothing; clearing the gate is the owner's action.

### Blocker 1 (provenance) — CLOSED, and it fails closed

`actorId` / `actorRunId` / `actorBuildId` are inside the JCS payload (`signable.ts:28-35`), not merely attached. Re-running my forged-receipt mint plus a fail-closed battery:

```
PASS  valid=false  receipt with "actorId"      DELETED  (old-format receipt)
PASS  valid=false  receipt with "actorRunId"   DELETED
PASS  valid=false  receipt with "actorBuildId" DELETED
PASS  valid=false  receipt with "actorId"      SWAPPED after signing
PASS  valid=false  receipt with "actorRunId"   SWAPPED after signing
PASS  valid=false  receipt with "actorBuildId" SWAPPED after signing
PASS  valid=false  receipt with actorRunId     NULLED  after signing
```

Absence fails closed because the fields are `z.string().nullable()` (required, nullable) rather than `.optional()` (`types.ts:51-53`) — the right choice, and a pre-existing v0.1 receipt cannot be replayed against the new verifier. The local-run `null` path still verifies, which is correct and tested.

My original forgery still succeeds (`valid=true` for a self-minted receipt claiming an invented `actorRunId`) and **that is the honest, unavoidable property of an ephemeral self-embedded key** — the fix was never going to be to make offline verification prove provenance. What I asked for was that the product stop claiming otherwise, and the README now does exactly that, in plain terms: "Anyone can generate their own Ed25519 keypair and sign a fabricated action log claiming any `actorRunId` they like, and that receipt will still verify." Provenance is correctly relocated to cross-checking `actorRunId` against the run's own dataset/KV record, which `main.ts:68-69` publishes and which only the real running actor can write for that run. Integrity and attribution are now separated and each claimed accurately. This is the right resolution.

### Blocker 2 (signature coverage) — CLOSED

Both schemas are `.strict()` (`types.ts:16`, `types.ts:55`). My exact injection repro now rejects at both levels:

```
PASS  valid=false  unknown key injected onto a signed action-log ENTRY
PASS  valid=false  unknown key injected at receipt TOP LEVEL
```

`receipt-unknown-key-injection.test.ts` reproduces the attack with an accurate comment explaining why silent stripping made "verified" diverge from "what you read". Verified content and readable content are now the same object.

### Blocker 3 (SSRF) — CLOSED for the reviewed class

All 13 of my original probe targets are blocked, including the bracket bug (`stripIPv6Brackets`, `tools.ts:26`) and the static-DNS-record class that the old disclaimer wrongly scoped as rebinding-only. I extended the table by 15 targets, including cases hitting the DNS path against **live** DNS rather than the suite's mock:

```
BLOCKED  http://localtest.me/              DNS name -> ::1  (DNS + IPv6 path together)
BLOCKED  http://127.0.0.1.nip.io/          DNS name -> 127.0.0.1
BLOCKED  http://10.0.0.1.nip.io/           DNS name -> RFC1918
BLOCKED  http://[0:0:0:0:0:ffff:7f00:1]/   uncompressed IPv4-mapped loopback
BLOCKED  http://[fe80::1]/ [fc00::1]/ [fdff:ffff::1]/   IPv6 link-local + fc00::/7 bounds
BLOCKED  http://[::ffff:0:0]/              IPv4-mapped 0.0.0.0
BLOCKED  http://LOCALHOST/                 case-folding
allowed  http://172.32.0.1/                public, correctly NOT over-blocked
allowed  http://100.128.0.1/               public, correctly NOT over-blocked
```

The CIDR boundaries are right in both directions, which matters as much as the blocks: 172.32 and 100.128 stay reachable. Unparseable IPv6 and DNS resolution failure both fail closed (`tools.ts:90`, `tools.ts:144`). The suite mocks `node:dns/promises` so it stays hermetic in CI, and my live-DNS run independently confirms the real resolver path.

Two residual gaps, both one-liners, neither reaching cloud metadata or any private range:

- **`http://[::]/` is allowed** (`isBlockedIPv6`, `tools.ts:88-107`). The IPv6 unspecified address connects to localhost on Linux, and its IPv4 twin `0.0.0.0/8` *is* blocked at `tools.ts:74` — so this is an oversight rather than a decision. Add an all-zero-groups check alongside the `::1` test.
- **`http://[64:ff9b::a9fe:a9fe]/` is allowed** — the NAT64 well-known prefix embedding the metadata address. Only exploitable behind a NAT64 gateway, so exotic, but it is the same embed-IPv4-in-IPv6 trick as the `::ffff:` case that is now handled.

I am not holding the verdict on these: the entire metadata / private / link-local / CGNAT / static-DNS class is closed, and the residual reaches only the container's own loopback. Worth fixing before Store submission.

### Verification performed

`57/57` pass in a clean worktree once the workspace is built. Note `load-llm-client.test.ts` fails to resolve `@revealui/ai` without a prior `turbo build`; CI's build-first step (`ci.yml:381`) covers it, same known pattern as the receipt suites. The lazy loader (`load-llm-client.ts`) makes the `optionalDependencies` declaration honest rather than textual, which resolves that hygiene finding properly.

### Carried forward (not blocking, from the #2198 suggestion list)

- `main.ts:73-74` still echoes raw `err.message` into the KV `OUTPUT`; provider SDK errors can carry request context. Redact to known-safe messages.
- Tool calls per step remain unbounded (`run-governed-task.ts:130`); `verify-receipt` is free with an unbounded `receipt` payload; `run-completed` is still charged when `stopReason` is `charge-limit`.

Good, disciplined fix-forward: every blocker addressed at the root rather than papered over, each with a prove-red test carrying an explanatory comment, and the honesty fix in the README is the part that most needed to land. The remaining items are hardening, not correctness.

PR #2206, opened 2026-07-26
fix(apify-actor): close :: and NAT64 SSRF residuals (GAP-431)

RevealUIStudio:
## Summary

Guardrail-2 reviewer left an APPROVE-with-residuals verdict on #2202
(https://github.com/RevealUIStudio/revealui/pull/2202#issuecomment-5085198453)
flagging two non-blocking SSRF bypasses in the `web_fetch` tool's
`isBlockedHost` guard (`packages/apify-actor-governed-run/src/agent/tools.ts`),
to close before Apify Store submission. This PR closes both.

| Residual | Why it slipped through | Fix |
|---|---|---|
| `http://[::]/` | The IPv6 unspecified address `::` routes to localhost on Linux. Its IPv4 twin `0.0.0.0/8` was already blocked by `isBlockedIPv4`, but `isBlockedIPv6` had no all-zero-groups check. | `isBlockedIPv6` now blocks `::` outright (all 8 groups zero) before the `::1` check. |
| `http://[64:ff9b::a9fe:a9fe]/` | The NAT64 well-known prefix `64:ff9b::/96` embeds an IPv4 address in its low 32 bits (here, the cloud metadata address `169.254.169.254`) — the same embed trick already closed for `::ffff:`-mapped addresses. | Extracted a shared `embeddedIPv4()` helper from the existing `::ffff:` handling and reused it for the `64:ff9b::/96` prefix, recursively checking the embedded address through `isBlockedIPv4` (same pattern, not a new special case). |

## Red → green evidence

Both cases were added to the existing reviewer's probe-target test table
(`src/__tests__/tools.test.ts`) FIRST as failing tests, confirmed red against
the pre-fix guard:

```
❯ src/__tests__/tools.test.ts (25 tests | 2 failed)
  × blocks: IPv6 unspecified address (:: routes to localhost on Linux) (http://[::]/)
  × blocks: NAT64 well-known prefix (64:ff9b::/96) embedding the cloud metadata IPv4 address (http://[64:ff9b::a9fe:a9fe]/)

AssertionError: expected 'fetch failed' to match /not allowed/
```

(the `fetch failed` error confirms the guard let the request through to the
real `fetch()` call rather than blocking it — the bypass was real, not a test
artifact).

After the fix, same suite:

```
Test Files  1 passed (1)
     Tests  25 passed (25)
```

## No over-blocking

Added a companion "confirmed-reachable addresses" test group covering the
two addresses the reviewer confirmed must stay reachable — `172.32.0.1` (just
past the 172.16.0.0/12 private range) and `100.128.0.1` (just past the
100.64.0.0/10 CGNAT range) — stubbing `fetch` so the guard is exercised
without a real network call. Both pass unblocked, both before and after this
change.

## Verification run

- `pnpm vitest run` in `packages/apify-actor-governed-run`: 9 test files, 61
  tests, all green.
- `pnpm typecheck` / `pnpm lint` (biome): clean.
- `pnpm gate --changed`: all phases pass **except** "Contracts OpenAPI mirror
  drift" — confirmed pre-existing and unrelated (reproduces identically on a
  pristine `origin/test` checkout with zero diff to `packages/openapi/`;
  tracked separately as GAP-436 in the internal tracker). This PR touches neither
  `packages/openapi` nor `packages/mcp`.
- Pre-push quality gate (phase 1, changed-scoped) ran automatically and
  passed.

## Review status

**REVIEW-PENDING under guardrail 2 — do not merge until a non-author APPROVE
verdict is recorded.** This touches `packages/apify-actor-governed-run/`,
which is on the fleet's widened security-sensitive path list (SSRF-adjacent
code). The fleet security-path self-check (`--diff origin/test`) confirms:
security-sensitive, gate held pending a recorded verdict.

### DIFF
--- a/packages/apify-actor-governed-run/src/__tests__/tools.test.ts
+++ b/packages/apify-actor-governed-run/src/__tests__/tools.test.ts
@@ -75,12 +75,43 @@ describe('webFetchTool', () => {
       ['short-form IPv4 (normalizes to loopback)', 'http://127.1/'],
       ['canonical loopback', 'http://127.0.0.1/'],
       ['canonical cloud metadata endpoint', 'http://169.254.169.254/'],
+      // Guardrail-2 APPROVE-with-residuals on revealui#2202
+      // (https://github.com/RevealUIStudio/revealui/pull/2202#issuecomment-5085198453):
+      // two more IPv6 bypasses the 13-target probe above did not cover.
+      ['IPv6 unspecified address (:: routes to localhost on Linux)', 'http://[::]/'],
+      [
+        'NAT64 well-known prefix (64:ff9b::/96) embedding the cloud metadata IPv4 address',
+        'http://[64:ff9b::a9fe:a9fe]/',
+      ],
     ];
 
     it.each(mustBeBlocked)('blocks: %s (%s)', async (_label, url) => {
       const result = await webFetchTool.execute({ url });
       expect(result.success).toBe(false);
       expect(result.error).toMatch(/not allowed/);
+    });
+  });
+
+  describe('SSRF guard -- confirmed-reachable addresses must not be over-blocked', () => {
+    const mustStayReachable: Array<[label: string, ip: string]> = [
+      ['just past the 172.16.0.0/12 private range', '172.32.0.1'],
+      ['just past the 100.64.0.0/10 CGNAT range', '100.128.0.1'],
+    ];
+
+    it.each(mustStayReachable)('does not block: %s (%s)', async (_label, ip) => {
+      // Stub fetch so the guard is exercised without a real network call --
+      // what's under test is that isBlockedHost lets the request through to
+      // fetch() at all, not the network behavior beyond that point.
+      const fetchSpy = vi
+        .spyOn(globalThis, 'fetch')
+        .mockResolvedValue(new Response('ok', { status: 200 }));
+      try {
+        const result = await webFetchTool.execute({ url: `http://${ip}/` });
+        expect(fetchSpy).toHaveBeenCalled();
+        expect(result.success).toBe(true);
+      } finally {
+        fetchSpy.mockRestore();
+      }
     });
   });
 
--- a/packages/apify-actor-governed-run/src/agent/tools.ts
+++ b/packages/apify-actor-governed-run/src/agent/tools.ts
@@ -80,16 +80,30 @@ function isBlockedIPv4(ip: string): boolean {
   return false;
 }
 
-/** Blocked IPv6 ranges: loopback (::1), unique local addresses (fc00::/7,
- * covers fd00::/8), link-local (fe80::/10), and IPv4-mapped addresses
- * (::ffff:a.b.c.d), whose embedded IPv4 address is recursively checked
- * against `isBlockedIPv4` -- this is what closes the
- * `[::ffff:127.0.0.1]` / `[::ffff:a9fe:a9fe]` bypasses. */
+/** Render the last two 16-bit groups of an expanded IPv6 address as the dotted
+ * IPv4 address they embed -- shared by the IPv4-mapped (`::ffff:a.b.c.d`) and
+ * NAT64 (`64:ff9b::a.b.c.d`) embed forms below. */
+function embeddedIPv4(groups: number[]): string {
+  const g6 = groups[6] ?? 0;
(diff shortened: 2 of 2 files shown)


RevealUIStudio:

Re-attacked `57b1198e2` in a clean worktree by the same non-author session that flagged these two residuals on #2202. Both are closed, no over-blocking introduced, prior-round fixes intact.

### The two residuals I flagged — both CLOSED

`isBlockedIPv6` (`tools.ts:88-116`) now blocks the all-zero `::` address (`groups.every(isZero)`, before the `::1` check) and detects the NAT64 well-known prefix `64:ff9b::/96` (`groups[0] === 0x64 && groups[1] === 0xff9b && groups[2..5] all zero`), extracting its embedded IPv4 via the shared `embeddedIPv4` helper and running it through the existing `isBlockedIPv4`. Factoring the IPv4-mapped and NAT64 extraction into one helper is the right shape — one place decides what an embedded v4 address is.

My original two bypasses plus the requested variants, run against the actual compiled guard:

```
BLOCKED  http://[::]/                        orig residual: IPv6 unspecified
BLOCKED  http://[64:ff9b::a9fe:a9fe]/        orig residual: NAT64 -> metadata
BLOCKED  http://[::0]/                       variant: ::0 spelled out
BLOCKED  http://[0:0:0:0:0:0:0:0]/           variant: fully-expanded ::
BLOCKED  http://[::]:80/                     variant: :: with explicit port
BLOCKED  http://[64:ff9b::7f00:1]/           variant: NAT64 -> 127.0.0.1
BLOCKED  http://[64:ff9b::a00:1]/            variant: NAT64 -> 10.0.0.1 (RFC1918)
BLOCKED  http://[64:ff9b::c0a8:1]/           variant: NAT64 -> 192.168.0.1
BLOCKED  http://[64:ff9b::6464:64c8]/        variant: NAT64 -> 100.100.100.200 (CGNAT)
BLOCKED  http://[64:ff9b:0:0:0:0:a9fe:a9fe]/ variant: NAT64 fully expanded
BLOCKED  http://[64:ff9b::0.0.0.0]/          variant: NAT64 -> 0.0.0.0
BLOCKED  http://[::1]/                        regression: ::1 loopback
BLOCKED  http://[::ffff:a9fe:a9fe]/           regression: IPv4-mapped metadata
```

The fully-expanded forms confirm the check keys off the parsed group values, not the textual form, so compression tricks don't help an attacker.

### No over-blocking

Public addresses that merely resemble the blocked shapes stay reachable:

```
allowed  http://[64:ff9b::808:808]/         NAT64 -> 8.8.8.8 (public)
allowed  http://[2001:4860:4860::8888]/     ordinary public IPv6
allowed  http://[64:ff9c::7f00:1]/          64:ff9c (NOT the NAT64 prefix) -> public
allowed  http://172.32.0.1/                 just past 172.16/12
allowed  http://100.128.0.1/                just past 100.64/10
```

The `64:ff9c::` case matters: the prefix match is exact on both groups, so an address one hex digit off the well-known prefix is correctly treated as a normal public v6 address rather than a NAT64 embed. The PR's own `mustStayReachable` group locks the two IPv4 CIDR boundaries.

### Verification

`61/61` pass in a clean worktree (after the standard `turbo build` of `@revealui/security` + `@revealui/ai`; same known build-first requirement CI covers at `ci.yml:381`). Both residuals are added as red-first tests inside the reviewer's probe table with an accurate provenance comment, and the new confirmed-reachable `it.each` group guards against regression to over-blocking.

The `Security review gate` FAILURE on this PR is correct and expected: `packages/apify-actor-governed-run/` is now in `security-paths.shared.json` (added by #2202), so the gate demands a recorded verdict — this comment is it. The earlier `Scan body via gitleaks` FAILURE (21:05:08Z) is superseded by a later SUCCESS run (21:09:54Z) after a body edit, which is the benign superseded-duplicate pattern, not a live leak. Per disposition discipline I applied no label and merged nothing; clearing the gate is the owner's action.

Clean, minimal, root-cause fix with tests that would have caught the bug. Nothing outstanding on the SSRF surface. The non-blocking hardening items carried forward from #2202 (raw `err.message` in the KV `OUTPUT`, unbounded tool-calls-per-step, free `verify-receipt` with an unbounded payload, `run-completed` charged on `charge-limit`) are out of scope for this PR and remain as follow-ups.

PR #2210, opened 2026-07-27
build(apify-actor): standalone single-directory Docker publish output (GAP-431)

RevealUIStudio:
## Summary

Repackages `packages/apify-actor-governed-run` so it can be published to the Apify Store as a **self-contained directory** whose Docker build needs only its own directory as context, sourcing `@revealui/*` deps from the **public npm registry** instead of `workspace:*`. Owner-chosen route (2026-07-26). This is **packaging/build only** -- no changes to signing, SSRF-guard, or agent-loop logic, which is already guardrail-2 APPROVED (revealui#2202, revealui#2206).

## Audit (cited)

- [packages/apify-actor-governed-run/package.json](packages/apify-actor-governed-run/package.json) -- runtime deps: `@revealui/security` (workspace:*, required), `@revealui/ai` (workspace:^, optionalDependency, lazy-loaded), `apify@^3.7.2`, `zod: catalog:`.
- [packages/apify-actor-governed-run/tsup.config.ts:18](packages/apify-actor-governed-run/tsup.config.ts#L18) -- `external: ['apify', '@revealui/ai']`. **Verified by inspecting the actual build output** (`dist/chunk-*.js`) that the pre-existing comment on this line was wrong: `@revealui/security` and `zod` are not bundled inline -- tsup auto-externalizes everything listed in `package.json` `dependencies` by default, so both stay real `import`s in the compiled output. Fixed the comment (no functional change -- the `external: [...]` array itself is byte-identical, `git diff` confirms).
- [packages/apify-actor-governed-run/src/main.ts](packages/apify-actor-governed-run/src/main.ts), [src/agent/load-llm-client.ts](packages/apify-actor-governed-run/src/agent/load-llm-client.ts) -- confirmed `@revealui/ai` is dynamically imported only in `run-task` mode; `verify-receipt` mode needs neither `@revealui/ai` nor an LLM key.
- Old `Dockerfile` (pre-this-PR) built from the monorepo root context (`COPY packages/ai/package.json`, `packages/security/package.json`, etc. + `pnpm deploy --legacy`) -- that's the thing this PR replaces for the publish path. It stays as-is for local/dev reference; the new standalone Dockerfile lives in the generated `.apify-build/` output.
- `npm view` re-verified at design time and again at build time (the build script does this live): `@revealui/security@0.6.0`, `@revealui/ai@0.9.0` (-> `@revealui/contracts@0.8.1`, `core@0.12.2`, `db@0.10.0` -> `config@0.6.0`, `resilience@0.2.4`, `utils@0.3.6`), `apify@3.7.2`, `zod@4.4.3` -- all published and installable, dependency graph resolves entirely from the public registry with zero `workspace:*` leakage.

## Design

Kept the actor in the monorepo (`workspace:*`) for local dev + the existing vitest suite -- unchanged. Added [scripts/deploy/build-apify-actor-standalone.mjs](scripts/deploy/build-apify-actor-standalone.mjs) (aliased as a root `build:apify` script), which:

1. Runs the existing `tsup` build.
2. Emits `packages/apify-actor-governed-run/.apify-build/` (gitignored, regenerated on every run -- avoids hand-maintaining two Dockerfiles/package.jsons in lockstep, which is exactly the drift class `extend-before-create`/`code-over-docs` exist to prevent):
   - `dist/` -- compiled actor, unmodified.
   - `package.json` -- standalone. `@revealui/security` + `@revealui/ai` pinned exact to their live-resolved npm versions (`npm view <pkg> version` at build time, hardcoded fallback if offline); `apify`/`zod` as normal caret deps. `@revealui/ai` stays `optionalDependencies`, matching the source package's lazy-load design.
   - `Dockerfile` -- single-stage, `FROM node:24-alpine`, copies `package.json`, installs with `npm install --omit=dev`, copies `dist`, drops to a non-root user, `CMD node dist/main.js`. No monorepo file copies, no pnpm workspace.
   - `.dockerignore`, `README.md`, `.actor/actor.json` + `.actor/input_schema.json` (copied as-is; `actor.json`'s relative `dockerfile`/`readme` fields resolve correctly against the output layout).

Chose "generate on demand" over "commit a hand-maintained second package" because the alternative requires remembering to re-sync two Dockerfiles and two dependency lists every time `@revealui/security`/`@revealui/ai` publish a new version -- a manual step that will drift.

## Verify

- **Docker build, from the single output directory only** (Docker 28.1.1, daemon up): built `revealui/apify-actor-governed-run:standalone-test` from `packages/apify-actor-governed-run/.apify-build` as the entire build context -- succeeded, the dependency install pulled 263 packages from the public registry in ~40s (no monorepo files in the build context at all).
- **Dependency install tree** (top-level, resolved from the registry): `@revealui/ai@0.9.0`, `@revealui/security@0.6.0`, `apify@3.7.2`, `zod@4.4.3` -- plus their declared transitive `@revealui/{contracts,core,db,resilience,utils,config}` packages, all from npm. Zero `workspace:*`.
- **Smoke-ran the built image** in `verify-receipt` mode (free, no LLM key needed) against a real signed receipt generated with the actor's own `signActionLog` (workspace dev path): `{"valid": true}`, exit 0. Then tampered one field of the same receipt and re-ran: `{"valid": false, "reason": "signature verification failed"}`, exit 0 -- proves the standalone image's `@revealui/security`-backed Ed25519 verification is genuinely functioning end-to-end, not a stub.
- **Existing vitest suite, unchanged workspace dev path**: the package's own test script (`vitest run`) -> 9 files / 61 tests, all passing.
- **Fleet gate**: PASS. Two unrelated pre-existing failures surfaced along the way and were resolved/ruled out, not caused by this diff:
  - "Project manager check" / "Build artifacts" failures were this fresh worktree simply not having run a full monorepo build yet (turbo build-order issue, not a regression) -- resolved by running the full build.
  - "Contracts OpenAPI mirror drift" (`packages/openapi/contracts.openapi.json`, ~10k line diff) is **pre-existing on `origin/test` itself** -- verified via `git stash` back to the pristine base commit and re-running the same check, which fails identically with zero relation to this PR's files. Left untouched; out of scope for a packaging-only PR.
  - A dependency audit on the standalone output surfaces 4 pre-existing high-severity advisories in `apify`'s own `crawlee`/`minimatch`/`brace-expansion` chain -- unrelated to this packaging change (same in the monorepo's existing `apify@3.7.2` dependency), not something this PR should "fix" by downgrading `apify`.

`git diff --stat origin/test`:
```
 .gitignore                                       |  4 ++++
 package.json                                     |  1 +
 packages/apify-actor-governed-run/tsup.config.ts | 16 +++++++++++++---
 scripts/deploy/build-apify-actor-standalone.mjs   | (new file)
```
No `src/*.ts` logic files touched. The `tsup.config.ts` diff is comment-only -- `git diff` shows the `external: [...]` array line itself is byte-identical before/after.

## Fix-forward: guardrail-2 REQUEST-CHANGES blocker resolved (2026-07-27)

Reviewer's blocker ([PR comment](https://github.com/RevealUIStudio/revealui/pull/2210#issuecomment-5086230242)): `@revealui/security` + `@revealui/ai` versions came from `npm view <pkg> version` (= latest on the registry at build time). The serialized pin was exact, but the *source* floated -- the next publish of either package would silently ship an unreviewed version through `apify push`, bypassing guardrail-2 on the receipt-signing surface.

**Fix** in [scripts/deploy/build-apify-actor-standalone.mjs](scripts/deploy/build-apify-actor-standalone.mjs):
- Promoted the two first-party pins to a new `REVIEWED_VERSIONS` constant (`@revealui/security: '0.6.0'`, `@revealui/ai: '0.9.0'`) -- now the sole, authoritative source for what the generated `package.json` ships. `npm view` is no longer consulted to choose these values.
- Added `warnIfPinIsStale()`: still calls `npm view` for these two packages, but now only to print a warning (`npm has X@Y but this build pins the guardrail-2-reviewed Z; bump the pin in a reviewed PR to adopt Y`) when the registry has moved past the pin. It never changes what ships, and a failed lookup (offline) is not a build error.
- `apify` + `zod` (third-party, not on the guardrail-2 sensitive-path list) keep resolving to npm latest via the existing `resolveVersion()`, unchanged.
- Added a code comment on `REVIEWED_VERSIONS` explaining why: shipping an unreviewed signing-dependency version through `apify push` would bypass the guardrail-2 gate; a pin bump is now a deliberate, reviewed one-line PR edit to this file.

**Nit addressed** (reproducible image build): the build script now also runs `npm install --package-lock-only --omit=dev` inside the generated `.apify-build/` output after writing `package.json`, and the Dockerfile now runs `npm ci --omit=dev` against that lockfile instead of `npm install --omit=dev`. The lockfile lives inside the gitignored `.apify-build/` directory (regenerated every run, same as the rest of that output) -- `apify`/`zod`'s transitive tree no longer re-resolves on every image build.

**Re-proof:**
- `pnpm build:apify` -- succeeded end to end (tsup build, drift checks logged no warnings since the registry currently matches the reviewed pins, lockfile generated, Dockerfile written).
- Generated `packages/apify-actor-governed-run/.apify-build/package.json` -- confirmed `"@revealui/security": "0.6.0"` and `"@revealui/ai": "0.9.0"` exact (no caret), matching `REVIEWED_VERSIONS` byte-for-byte.
- `docker build .` from `.apify-build/` as the entire context (Docker 28.1.1) -- succeeded; `npm ci --omit=dev` installed 263 packages from the lockfile in ~11s.
- Smoke-ran the built image in `verify-receipt` mode against a freshly self-signed receipt (`signActionLog`, workspace dev path, mounted as `INPUT.json` via `CRAWLEE_STORAGE_DIR`): valid receipt -> `{"valid": true}`; same receipt with one action-log entry appended after signing -> `{"valid": false, "reason": "signature verification failed"}`. Confirms genuine end-to-end Ed25519 verification through the standalone, npm-ci-installed `@revealui/security` build, not a stub.
- Independently re-verified the drift-warning path against the live registry (deliberately mismatched pin vs. `zod`'s real latest) -- warns as designed; a matching pin logs nothing.
- Existing vitest suite: 9 files / 61 tests, all still passing.
- `pnpm gate --changed`: same result class as the original PR -- everything passes except the pre-existing "Contracts OpenAPI mirror drift" check, independently re-confirmed failing identically on a clean `origin/test` checkout with zero relation to this PR's files (unrelated, out of scope).
- `git diff --stat origin/test` -- confirmed zero `packages/apify-actor-governed-run/src/**/*.ts` changes; only `scripts/deploy/build-apify-actor-standalone.mjs` changed since the reviewer's pass (plus the pre-existing `.gitignore`/`package.json`/`tsup.config.ts` changes already reviewed).

**The standing REQUEST-CHANGES resolves only when a fresh APPROVE verdict is recorded** -- this fix-forward commit does not self-clear guardrail-2. Not presented as merge-ready.

## Security

`packages/apify-actor-governed-run` is on the guardrail-2 sensitive-paths list. The fleet security-PR checker confirms: **SECURITY-SENSITIVE, touched `packages/apify-actor-governed-run/tsup.config.ts`**.

**REVIEW-PENDING under guardrail 2 -- do not merge until a non-author APPROVE verdict is recorded.** Only packaging/build/Docker changed here: no signing, SSRF-guard, or agent-loop logic touched (see `git diff --stat` above -- every changed line is build tooling, `.gitignore`, or a corrected code comment). This is not presented as merge-ready.

## Owner actions

- Review the fix-forward commit and record a fresh guardrail-2 verdict (`APPROVE` marker or `sec-review:approved` label) before merge -- this PR does not self-clear. The prior REQUEST-CHANGES stands until a new APPROVE is recorded.
- After merge, publish by running the standalone build script, then pointing the Apify CLI's push command's directory flag at the generated `packages/apify-actor-governed-run/.apify-build` output.


RevealUIStudio:

Packaging-only review of the standalone repackaging (GAP-431). Scope: confirm the reviewed signing/SSRF/agent logic (approved on #2202, #2206) is unchanged and that the standalone build faithfully ships the same audited code. I am not the author.

### What is clean (verified)

1. **No actor logic changed.** `git diff origin/test...origin/feat/gap-431-standalone-actor --stat` touches only 4 files: `.gitignore`, root `package.json` (adds `build:apify` script), `packages/apify-actor-governed-run/tsup.config.ts`, and the new `scripts/deploy/build-apify-actor-standalone.mjs`. **Zero** `packages/apify-actor-governed-run/src/**/*.ts` files changed. The `tsup.config.ts` diff is comment-only; the behavior line `external: ['apify', '@revealui/ai']` is byte-identical to `origin/test`. The reviewed signing/SSRF/agent code is untouched.

2. **Dependency set is faithful.** The generated `package.json` mirrors the source package: `@revealui/security` + `apify` + `zod` as deps, `@revealui/ai` as an optionalDependency (lazy, run-task mode only). No dependency added or dropped.

3. **Dockerfile introduces no new exposure.** Runs as a non-root system user (`uid 1001 actor`, `USER actor` after `chown`), `npm install --omit=dev`, no secrets baked in, no `curl|sh`, no env that disables the SSRF guard or alters tool config. `CMD ["node", "dist/main.js"]` only.

4. **No secret-leak vector in the artifact.** `.dockerignore` excludes `.env`, `.env.*`, `.git`, `node_modules`, logs. The build script copies only `dist/` + static Apify metadata into a freshly-reset `.apify-build/`; it never copies `.env` or credentials, and `.gitignore` excludes `.apify-build/` so the artifact is not committed. `execFileSync('npm', ['view', <const>, ...])` takes only hardcoded package names — no injection surface.

### Blocker

**[scripts/deploy/build-apify-actor-standalone.mjs:47-64, 108-121] The security-signing dependency versions are resolved to `npm view <pkg> version` (latest-at-build-time), not pinned to the reviewed versions.**

The serialized pin is exact (no caret) for `@revealui/security` and `@revealui/ai` — that part is correct. But the *value* comes from `resolveVersion()` → `npm view <pkg> version`, which returns whatever is **latest** on the public registry at build time. Today that happens to equal the audited versions (`npm view` returns `@revealui/security 0.6.0`, `@revealui/ai 0.9.0`, matching #2202/#2206), so today's artifact is faithful. But the mechanism does not enforce it:

- The moment `@revealui/security` (the receipt-signing crypto surface) or `@revealui/ai` publishes a new version, the next `pnpm build:apify` silently pins the standalone actor to that newer, **unreviewed** version and `apify push`es it.
- That regeneration is a local build + push. It does **not** pass through CI or a guardrail-2 verdict. So unreviewed signing-dependency code can reach the published actor with no review gate — the exact invariant guardrail-2 exists to protect, and precisely the "floating range on the security-signing dependency" hazard this review was asked to flag. Resolve-to-latest is a floating resolution even though the emitted string is exact.
- The tell is that the reviewed versions already live in the script as `FALLBACK_VERSIONS` (`@revealui/security: '0.6.0'`, `@revealui/ai: '0.9.0'`) — but only as a network-failure fallback. Author intent ("pin to the reviewed versions") and implementation ("pin to latest") diverge.

**Suggested fix (cheap):** make the reviewed versions the authoritative pin for the two first-party security deps — hoist `@revealui/security`/`@revealui/ai` to explicit constants used directly in the generated `dependencies`/`optionalDependencies`, and demote `npm view` to a drift *check* that warns (or fails) when latest differs from the pinned reviewed version, rather than sourcing the pin from it. Bumping the actor to a newer signing dependency then becomes a deliberate one-line edit that rides a reviewed PR — which is where the guardrail-2 verdict should land.

### Nit

- **[Dockerfile, generated]** The image uses `npm install --omit=dev` with no committed lockfile, so `apify`/`zod` float within their caret ranges (`^3.7.2` / `^4.4.3`) at each image build. Third-party and standard practice, not blocking — but `npm ci` against a generated lockfile would make the runtime image reproducible. `zod` participates in schema validation; worth keeping in mind if you tighten later.

Once the first-party signing deps are pinned to the reviewed versions (not resolve-to-latest), this is a clean packaging change and I would flip to APPROVE.

RevealUIStudio:

Re-review of commit `281f1e73d` against my prior reviewed commit. **This resolves my standing REQUEST-CHANGES.** Only `scripts/deploy/build-apify-actor-standalone.mjs` changed beyond the prior review (the `.gitignore`/root `package.json`/`tsup.config.ts` diffs are byte-identical to what I already cleared; still zero `src/**/*.ts` logic diff).

### Blocker resolved

The floating resolution of the signing surface is gone. A new `REVIEWED_VERSIONS` constant (`@revealui/security: 0.6.0`, `@revealui/ai: 0.9.0`) is now the **sole** authoritative source for those two pins: `securityVersion`/`aiVersion` read from it, and the generated `package.json` emits `@revealui/security: "0.6.0"` (dependencies) and `@revealui/ai: "0.9.0"` (optionalDependencies) **exact, no caret**. `npm view` no longer sources them — it is demoted to `warnIfPinIsStale()`, which only prints a drift warning when the registry has moved past the pin and never changes what ships (returns silently on lookup failure, never throws). Bumping to a newer signing version is now a deliberate edit to this file that rides a reviewed PR, which is exactly where the guardrail-2 verdict belongs.

### Nit resolved

Step 5 generates `package-lock.json` via `npm install --package-lock-only --omit=dev` (lockfile-only: resolves the tree from registry metadata, no tarball download, no install-script execution at generation time), and the Dockerfile now `COPY package.json package-lock.json` + `RUN npm ci --omit=dev`. `apify`/`zod` are frozen per-build by the lockfile rather than re-resolving their caret ranges at each image build. The lockfile is written into gitignored `.apify-build/`, not committed.

### No new exposure introduced by the delta

- `npm view` / `npm install --package-lock-only` are invoked with hardcoded package-name args only — no injection surface.
- Dockerfile unchanged in posture: non-root `USER actor` (uid 1001), `--omit=dev`, no secrets baked in, no `curl|sh`, SSRF guard and tool config untouched.
- `.dockerignore` still excludes `.env`/`.env.*`/`.git`; the build copies only `dist/` + static Apify metadata into a freshly-reset dir.

The reviewed signing/SSRF/agent logic (revealui#2202, #2206) is unchanged; the standalone build now faithfully and durably ships the audited `security@0.6.0` / `ai@0.9.0` code. Clean packaging change.

PR #2212, opened 2026-07-27
fix(security): sign audit payload in jsonb-storage form (GAP-442)

RevealUIStudio:
## The bug

Every prod audit write was failing with `canonicalizeJcs: undefined is not JSON-representable` (200+ consecutive on the Fly worker's first boot; the audit-write-failure-ratio alarm fired). Trigger: the worker's own header-less `/api/health` probes produce `data.read` events whose payload carries an undefined `actor.ip` / `userAgent` / `requestId`, and `canonicalizeJcs` (correctly) throws on undefined.

## Why the fix is at the signing boundary, not the middleware

The root cause is deeper than the crash. The `payload` column is `jsonb`, stored as `JSON.stringify(payload)`, which **drops** undefined keys (turns undefined array elements into null). So a payload signed with a nested undefined would sign bytes the verifier can **never** reproduce from the DB readback — a permanently un-verifiable signature. Fixing only the one middleware call site would leave that latent verify-mismatch for every other caller.

## The fix

Normalize **only the payload** to its jsonb-storage form (a JSON round-trip) before signing, so sign-time bytes equal verify-time (DB) bytes. The structural columns (`id`/`sequence`/`eventType`/`severity`/`agentId`) stay strict — a signed row genuinely must have them, so undefined there is still a loud error. A clean payload round-trips to an equivalent object and canonicalizes identically, so **existing signatures are unchanged**.

## Prove-red

3 new tests throw the exact prod `canonicalizeJcs` error on base, pass with the fix; a 4th locks that clean payloads are byte-for-byte unchanged (regression guard); a 5th covers the undefined-array-element→null case. 48/48 across the four audit suites (signing, merkle, jcs, anchor-verify).

Review-pending: audit signing surface (`packages/security`) — non-author guardrail-2 verdict before merge.

### DIFF
--- a/packages/security/src/__tests__/audit-signing.test.ts
+++ b/packages/security/src/__tests__/audit-signing.test.ts
@@ -85,6 +85,49 @@ describe('Ed25519 audit signing (GAP-355 Stage 3)', () => {
     expect(v1).toBe(v2);
   });
 
+  describe('GAP-442: payload is signed in jsonb-storage form (undefined dropped)', () => {
+    it('a payload with a nested undefined field signs without throwing', () => {
+      // Pre-fix this threw "canonicalizeJcs: undefined is not JSON-representable"
+      // — the exact prod failure on header-less /api/health data.read events.
+      const payload = {
+        tool: 'read',
+        actor: { id: 'anonymous', ip: undefined },
+        requestId: undefined,
+      };
+      expect(() => auditSignableBytes(makeRow({ payload }))).not.toThrow();
+    });
+
+    it('signs the undefined-dropped form so sign-time bytes == DB readback', () => {
+      // Postgres jsonb stores JSON.stringify(payload), which drops undefined
+      // keys. The signature must be over THAT form, or it can never verify
+      // against the stored row. Bytes for {a,b:undefined} must equal bytes for
+      // the round-tripped {a} the DB actually holds.
+      const withUndef = auditSignableBytes(makeRow({ payload: { a: 1, b: undefined } }));
+      const stored = auditSignableBytes(
+        makeRow({ payload: JSON.parse(JSON.stringify({ a: 1, b: undefined })) }),
+      );
+      expect(Buffer.from(withUndef).equals(Buffer.from(stored))).toBe(true);
+    });
+
+    it('a clean payload (no undefined) produces unchanged bytes (regression guard)', () => {
+      // The normalization must not perturb existing signatures: a payload with
+      // no undefined round-trips to an equivalent object and canonicalizes the
+      // same, so already-signed rows keep verifying.
+      const clean = { tool: 'read', args: { b: 2, a: 1 } };
+      const direct = auditSignableBytes(makeRow({ payload: clean }));
+      const roundtripped = auditSignableBytes(
+        makeRow({ payload: JSON.parse(JSON.stringify(clean)) }),
+      );
+      expect(Buffer.from(direct).equals(Buffer.from(roundtripped))).toBe(true);
+    });
+
+    it('an undefined array element is signed as null (jsonb semantics)', () => {
+      const withHole = auditSignableBytes(makeRow({ payload: { items: [1, undefined, 3] } }));
+      const asNull = auditSignableBytes(makeRow({ payload: { items: [1, null, 3] } }));
+      expect(Buffer.from(withHole).equals(Buffer.from(asNull))).toBe(true);
+    });
+  });
+
   it('verification fails for an unresolvable kid', () => {
     const row = makeRow();
     const { value } = signer.sign(auditSignableBytes(row));
--- a/packages/security/src/audit-signing.ts
+++ b/packages/security/src/audit-signing.ts
@@ -62,6 +62,38 @@ export function auditTimestampString(ts: Date | string): string {
   return d.toISOString();
 }
 
+/**
+ * Normalize the `payload` to the EXACT shape Postgres jsonb stores it as, before
+ * signing. The `payload` column is `jsonb`, and the driver writes it as
+ * `JSON.stringify(payload)` — which DROPS object keys whose value is `undefined`
+ * and turns `undefined` array elements into `null` (ES JSON semantics). So a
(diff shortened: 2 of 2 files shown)


RevealUIStudio:



Crown-jewel review (audit-log signing surface). Head `914348a`, base `origin/test` `c1775d5`. I did not author this change. The whole risk is signature stability: if the round-trip changes the signed bytes for any clean payload, every already-signed prod row (~1666) stops verifying. It does not. Reasoned proof plus empirical proof below.

## The non-regression argument (why no clean-payload byte can change)

The only field the diff changes in `auditSignableBytes` is `payload`: old signed `canonicalizeJcs(payload ?? null)`, new signs `canonicalizeJcs(toJsonbStorageForm(payload))` where `toJsonbStorageForm(x) === JSON.parse(JSON.stringify(x))`. Every other signed field is byte-for-byte unchanged.

The verifier has *always* consumed the DB readback, not the in-memory object. Confirmed in code: the sign path passes `entry.payload` (the live object, which is how `undefined` got in — `audit-store.ts:212`), while the verify/read path reconstructs from `row.payload` parsed out of the `jsonb` column (`audit-store.ts:318`). A `jsonb` readback is exactly the JSON round-trip form. So for **any row that currently verifies**, it must already be true that `canonicalizeJcs(rawPayload) === canonicalizeJcs(jsonbReadback) === canonicalizeJcs(roundTrip(rawPayload))` — that equality is the very precondition that let it verify. The new sign path now feeds that same round-trip form, so new-sign-bytes == old-verify-bytes == old-sign-bytes for every verifying row. The change can therefore only ever *fix* a non-verifying/throwing row, never break a verifying one. That is airtight, not hand-waved: the verifier never saw the raw object to begin with.

## Empirical confirmation (I ran it, did not take it on faith)

Byte-equality battery over `canonicalizeJcs(x)` vs `canonicalizeJcs(roundTrip(x))` across 17 clean payloads chosen to stress exactly the cases flagged: `-0` as a value and inside an array, large safe integer, floats incl. `1e21`/`1e-7`, deeply nested arrays/objects, unicode + control-char escaping, unicode keys, keys already-sorted vs not, string-that-looks-like-a-number, top-level null/array/object, anchor-shaped array of objects, falsy-but-defined values (`0`/`''`/`false`). **0 failures — all 17 byte-identical.** `-0` is a non-issue: `jcs.ts:53` already normalizes `-0`→`0`, matching `JSON.stringify(-0)`.

The three legitimately-differing cases behave as the argument predicts, and none could have verified before:
- nested `undefined` (the prod bug): old **threw** `canonicalizeJcs: undefined is not JSON-representable`; new signs `{"actor":{"id":"anon"},"evt":"read"}` — the stored form.
- `undefined` array element: old threw; new signs it as `null` (ES/jsonb semantics), matching the DB.
- **Date in payload**: old canonicalized to `{"at":{}}` (Dates have no enumerable own props), but the jsonb readback is the ISO string, so old sign≠old verify — a Date-payload row **never verified under the old code**. New signs `{"at":"2026-07-26T00:00:00.000Z"}`, matching storage. This is a fix, not a regression.

**Tests:** ran the four files myself — 48/48 pass. Prove-red: reverted only `audit-signing.ts` to `origin/test` (kept the new tests at HEAD), reran `-t "GAP-442"` — exactly 3 fail with `canonicalizeJcs: undefined is not JSON-representable`, and the 4th (the clean-payload regression guard) passes on base, which is itself evidence the clean path is unchanged. Restored via `git checkout HEAD --`; `git status` clean, `git diff HEAD` empty.

## The other four checks

2. **Sign==verify symmetry.** `auditSignableBytes` is the single shared builder (`audit-signer-env.ts:101` signs, `verifyAuditRow`/`audit-merkle.ts:238` verify). Both now normalize identically, and per the argument above a pre-fix clean row still verifies. Confirmed.

3. **Is normalizing only `payload` sufficient?** Yes, and scoping it there is the *correct* call, not a gap. `tenant`/`taskId`/`sessionId`/`policyViolations` carry `?? null`/`?? []` guards; `id`/`sequence`/`eventType`/`severity`/`agentId` are strict on purpose — they are non-nullable structural columns, so an `undefined` there is a genuine signer bug that must fail closed on a signing surface, not be silently masked. The store sources them from the `AuditEntry` (`agentId = actor.id ?? 'anonymous'`, etc.), so the middleware already guarantees them; if that ever regressed, the strict throw is the behavior you want. The free-form `jsonb` payload is the only field that can legitimately carry arbitrary nested `undefined` from event data, and it is the only one relaxed. Also checked the top-level parity: `toJsonbStorageForm` returns `null` for both `null` and `undefined` payloads, exactly matching the old `?? null`, and passes falsy-but-defined values (`0`/`''`/`false`) through unchanged.

4. **Merkle root canonicalization is untouched.** The diff is two files (`audit-signing.ts` + its test); `audit-merkle.ts` is not in it. Merkle leaves canonicalize signature strings, not payloads, so the strict `canonicalizeJcs` still guards the root. No weakening.

## Note (non-blocking, not introduced by this PR)

`toJsonbStorageForm` models the jsonb storage form with a JS `JSON.parse(JSON.stringify())`. For the `undefined`-dropping and key-order concerns that model is exact (and `canonicalizeJcs` re-sorts keys regardless). The one place a JS round-trip could in theory diverge from Postgres's actual jsonb serialization is exotic number formatting (extreme exponents / high precision) — but this PR does **not** change number behavior at all: `JSON.parse(JSON.stringify(n)) === n` for every finite number, so `canonicalizeJcs` is identical to the old raw-object path for numbers. Any residual jsonb-number exposure is pre-existing, equally present before this PR, and already documented as out-of-scope in `jcs.ts`'s number caveat (audit payloads don't carry such values). Worth a one-line acknowledgement in the function comment someday; not a blocker and not a regression.

Verdict: the fix is correct, the strict structural guarantee and the Merkle root are preserved, and no clean-payload byte changes — the ~1666 existing signatures keep verifying. Approving on the code. Merge and the `sec-review:approved` label remain the owner's call.

PR #2223, opened 2026-07-27
fix(auth): fresh unlicensed self-host sign-in + relax admin env validation for OSS self-host

RevealUIStudio:
## Problem (found deploying the Railway template as a real stranger would)
A fresh, unlicensed Free-tier self-host seeds the first admin but then fails login with "Unexpected error in signIn" (packages/auth/src/server/auth.ts outer catch, which swallowed the error). Separately, the admin crash-loops on boot demanding SESSION_COOKIE_DOMAIN + Stripe price IDs + Google email vars an OSS self-host does not use.

## Root cause
The DB-storage calls in signIn (rate-limit, account-lock, clearFailedAttempts) were unguarded. `clearFailedAttempts` runs immediately after a SUCCESSFUL credential check; a storage hiccup on the fresh DB threw into the generic outer catch and turned a valid login into "Unexpected error", with the error swallowed and undiagnosable.

## Fix
- Guard each DB-touching sign-in step: genuine failures fail closed with a logged, diagnosable `database_error`; the post-success bookkeeping (`clearFailedAttempts`) is best-effort so it can never sink a valid login. Outer catch now logs the real error.
- Relax admin env validation: SESSION_COOKIE_DOMAIN, Stripe price IDs, and Google email are optional for a plain unlicensed OSS self-host; critical vars (REVEALUI_KEK etc.) stay hard-required. Admin now boots without SKIP_ENV_VALIDATION.

## Verification
Red→green regression `oninit-bootstrap-signin.test.ts` (freshly-seeded unlicensed admin signs in); auth suite 536 green; env-validation tests added.

**Design note / crown-jewel surface:** this changes the auth sign-in path. **REVIEW-PENDING under guardrail 2 — do not merge until a non-author APPROVE verdict is recorded.**

### DIFF
--- a/apps/admin/src/app/api/auth/sign-in/route.ts
+++ b/apps/admin/src/app/api/auth/sign-in/route.ts
@@ -144,7 +144,10 @@ async function signInHandler(request: NextRequest): Promise<NextResponse> {
 
     return response;
   } catch (error) {
-    logger.error('Error signing in', { error });
+    logger.error('Error signing in', error instanceof Error ? error : undefined, {
+      message: error instanceof Error ? error.message : String(error),
+      name: error instanceof Error ? error.name : 'unknown',
+    });
     return createErrorResponse(error, {
       endpoint: '/api/auth/sign-in',
       operation: 'sign_in',
--- a/apps/admin/src/lib/utils/__tests__/env-validation.test.ts
+++ b/apps/admin/src/lib/utils/__tests__/env-validation.test.ts
@@ -1,0 +1,145 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { validateRequiredEnvVars } from '../env-validation.js';
+
+/**
+ * GAP-430/GAP-440: a fresh, unlicensed self-host deploy (e.g. the Railway
+ * marketplace template) never uses Stripe checkout, cross-subdomain session
+ * cookies, or the Workspace-backed signup verification email — those are
+ * hosted-SaaS-only concerns. Requiring their env vars in production blocked
+ * a stranger's deploy from booting at all without SKIP_ENV_VALIDATION=true.
+ */
+
+const ENV_KEYS = [
+  'REVEALUI_SECRET',
+  'REVEALUI_PUBLIC_SERVER_URL',
+  'POSTGRES_URL',
+  'DATABASE_URL',
+  'REVEALUI_KEK',
+  'REVEALUI_FLEET_MODE',
+  'REVEALUI_LICENSE_PRIVATE_KEY',
+  'REVEALUI_DEPLOYMENT_MODE',
+  'SESSION_COOKIE_DOMAIN',
+  'NEXT_PUBLIC_STRIPE_PRO_PRICE_ID',
+  'NEXT_PUBLIC_STRIPE_MAX_PRICE_ID',
+  'NEXT_PUBLIC_STRIPE_ENTERPRISE_PRICE_ID',
+  'GOOGLE_SERVICE_ACCOUNT_EMAIL',
+  'GOOGLE_PRIVATE_KEY',
+] as const;
+
+const saved: Record<string, string | undefined> = {};
+
+const VALID_KEK = 'a'.repeat(64);
+
+function setCriticalRequiredVars(): void {
+  process.env.REVEALUI_SECRET = 'a'.repeat(32);
+  process.env.REVEALUI_PUBLIC_SERVER_URL = 'https://admin.example.com';
+  process.env.POSTGRES_URL = 'postgresql://user:pass@host:5432/db';
+  process.env.REVEALUI_KEK = VALID_KEK;
+}
+
+describe('validateRequiredEnvVars', () => {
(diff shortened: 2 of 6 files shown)


RevealUIStudio:

Adversarial review by a non-author session (crown-jewel auth sign-in surface). Read the full diff, traced the control flow in `packages/auth/src/server/auth.ts`, and ran the tests in a scratch worktree.

### 1. Fail-closed analysis (the load-bearing question) — PASS

The critical property holds: **the pre-credential gates fail closed; only post-success bookkeeping is best-effort.**

- **Rate-limit check** (`checkRateLimit`) — on throw returns `{ success: false, reason: 'database_error' }` and `return`s. It does **not** fall through to grant a session. Fail-closed. ✅
- **Account-lock check** (`isAccountLocked`) — same shape: throw → `database_error` → `return`. A storage error can no longer let a locked account through, because a throw ends the request before any session is minted. Fail-closed. ✅
- **`clearFailedAttempts`** — best-effort (catch-and-log, continue), but it runs **only after `bcrypt.compare` already returned `true`** (valid credentials), and it is post-success bookkeeping, not a gate. Its worst-case failure mode leaves the failed-attempt counter **higher** than it should be (a future lockout triggers slightly sooner) — strictly more conservative, never a bypass. An attacker without valid credentials never reaches this line. ✅
- **Failure path is still fail-closed too:** `recordFailedAttempt` on the invalid-credentials path remains **unguarded** — if it throws, it propagates to the outer catch → `unexpected_error` → no session. So a storage error on the failure path does not silently skip brute-force accounting *and* grant access; it denies. `getClient`, the user query, `bcrypt.compare`, and `rotateSession` all likewise fail closed. ✅

No path was found where a thrown storage error now lets login proceed. The stated blocker condition does not occur.

### 2. Env-validation relaxation — correctly scoped, PASS

`SESSION_COOKIE_DOMAIN`, the three Stripe price IDs, and the Google Workspace email vars are now wrapped in `if (isSaasHosted)`, where `isSaasHosted = isHostedDeployment(process.env)` (`@revealui/core/deployment-mode`). `isHostedDeployment` returns hosted iff `REVEALUI_DEPLOYMENT_MODE=hosted` **or** `REVEALUI_LICENSE_PRIVATE_KEY` is present — so a licensed/hosted prod deploy **still requires all of them** (verified by `env-validation.test.ts`, which asserts the licensed path reports every one as missing). Unlicensed OSS self-host / Fleet kits skip them, which matches their actual runtime (no self-billing, no cross-subdomain cookies, no Workspace mail).

Critically, `REVEALUI_KEK`, `REVEALUI_SECRET`, `POSTGRES_URL`/`DATABASE_URL`, and `REVEALUI_PUBLIC_SERVER_URL` stay in `baseRequired` for **all** paths, and the KEK 64-hex format check is unchanged. Secret-bearing requirements were not weakened. ✅

Minor, non-blocking: a hosted deploy that removes the private key for signer isolation must set `REVEALUI_DEPLOYMENT_MODE=hosted` explicitly, else it is detected as `forge` and the hosted-only vars become optional. This is the existing GAP-260 posture-detection contract (and `deploymentModeKeyConsistencyError` guards the inverse), not a new weakness introduced here.

### 3. No secret leakage in the new logging — PASS

The outer catch and the per-step guards log only `error.message`, `error.name`, and `error.stack` of the caught error object (route.ts logs message + name). No password, session token, KEK, or license key material is passed to the logger. JS stack traces carry function/file:line, not variable contents. ✅

### 4. Regression tests exercise the real path and are red-without-fix — PASS

- `oninit-bootstrap-signin.test.ts` (new, PGlite) seeds the first admin exactly the way `onInit` does via the dynamic-SQL collections engine (raw INSERT, `created_at` from the SQL DEFAULT), then signs in — reproducing GAP-430. A second case drives `DatabaseStorage` (the production rate-limit/brute-force backend), not the in-memory fallback. Both pass. ✅
- The unit tests in `auth.test.ts` are red-without-fix by construction: the prior test asserted `unexpected_error` for a thrown rate-limit check; it now asserts `database_error` for the same input — i.e. the old (unguarded) code produces the old assertion and fails the new one. The `clearFailedAttempts`-throws case asserts `success === true`, which the pre-fix bare `await` (propagating to the outer catch) would fail. ✅

### 5. Test run — PR-relevant suite green

Ran in a scratch worktree (`origin/fix/gap-440-...`): `env-validation.test.ts` 6/6, `oninit-bootstrap-signin.test.ts` 2/2, and all four new `signIn` guard tests pass. The remaining auth-package file failures are **environmental, not PR regressions** — unbuilt workspace deps (`@revealui/core/observability/logger`, `@revealui/core/admin/utils/csrf` resolution errors before a full monorepo build; CI builds first). The 3 `signUp` unit failures are **pre-existing and unrelated to `signIn`**: `signUp` source is untouched by this diff, and I isolated the test's logger-mock hoist change (reverting it to a fresh `vi.fn()` leaves `signUp` still failing 3/3), ruling out the PR as the cause.

Verdict: the fix is correct and the security-relevant gates fail closed. Approving.

Issue #1644, opened 2026-06-27
chore(core): remove Vercel Blob storage fallback after R2 cutover completes


RevealUIStudio:
## Summary

The Vercel Blob storage provider is retained only as a migration-window fallback during the GAP-208 Cloudflare R2 cutover. R2 is the canonical backend; Blob is used only when R2 env is absent. Remove once R2 is mandatory across all environments.

## Current state

- Provider: `packages/core/src/storage/vercel-blob-provider.ts`
- Factory dispatch: `packages/core/src/storage/index.ts:12,36,52-53`
- Fallback selection: `apps/server/src/lib/storage.ts:29-30` (R2 if configured, else Blob)
- Config surface: `packages/config/src/modules/storage.ts:38`, `packages/config/src/schema.ts:45` (`BLOB_READ_WRITE_TOKEN`)
- Tests: `packages/core/src/storage/__tests__/vercel-blob-provider.test.ts`, `.../factory.test.ts`
- Known limitation while retained: Blob has no per-object metadata/Cache-Control API, so `opts.metadata` / `opts.cacheControl` are silently ignored.

## Gating

Blocked until R2 is configured in every environment and no installation depends on `BLOB_READ_WRITE_TOKEN`. Related: #1640 (BLOB token wrongly marked required on fresh R2 deploys).

## Action (when unblocked)

1. Delete `vercel-blob-provider.ts` and its tests.
2. Remove the Blob branch from `storage/index.ts` and `apps/server/src/lib/storage.ts`.
3. Drop `BLOB_READ_WRITE_TOKEN` from the config schema and env templates.

PR #2224, opened 2026-07-27
fix(admin): serve R2 media in CSP + next/image, allow blob: previews, drop Cloudinary/Blob remnants (GAP-208)

RevealUIStudio:
## Summary

The GAP-208 R2 smoke test (live admin, media upload) surfaced three serving-side defects; this fixes all three in the admin CSP + next/image allowlist:

- **`img-src` gains `blob:`** — the upload UI's local preview thumbnails are `URL.createObjectURL` blob URLs, which the browser was blocking (seen in the live console during the smoke test).
- **`img-src` + `images.remotePatterns` gain the canonical R2 public origin** (derived from `R2_PUBLIC_BASE_URL` via the URL parser, no regex; both hosted and fleet modes) — previously R2-served originals were absent from both allowlists while the **retired Vercel Blob CDN** ("originals stored here" — false since #1644) and the **never-canonical Cloudinary host** were still allowlisted. Those remnants are removed.
- **`object-src` becomes an explicit `'none'`** — the old hosted-mode `object-src https://res.cloudinary.com` was dead config; admin renders no `<object>`/`<embed>`. Cloudinary is also dropped from `script-src` (no usage anywhere in `apps/admin/src`).

## Test plan

- [x] Prove-red: with the source files reverted to `origin/test`, the three new proxy-csp assertions (blob:, R2 origin via stubbed env, no-Cloudinary + object-src 'none') and the updated fleet-mode assertion all FAIL (4 failed / 19 passed); with the fix they pass 23/23.
- [x] `biome check` on the changed files — clean.
- [x] `pnpm --filter admin typecheck` — clean (after building the workspace dependency closure in the worktree).

## Notes

- Security-classed (CSP surface): needs a non-author guardrail-2 verdict before merge — review-pending.
- The remaining GAP-208 smoke-test question (whether the form-save 400 was just the required `alt` field) is being verified live and is independent of this fix.

### DIFF
--- a/apps/admin/next.config.mjs
+++ b/apps/admin/next.config.mjs
@@ -119,18 +119,13 @@ const nextConfig = {
       },
       {
         protocol: 'https',
-        // Vercel Blob CDN — originals stored here, next/image resizes on demand
-        hostname: '*.blob.vercel-storage.com',
-      },
-      {
-        protocol: 'https',
-        hostname: 'res.cloudinary.com',
-      },
-      {
-        protocol: 'https',
         hostname: 'www.gravatar.com',
       },
-      ...[process.env.NEXT_PUBLIC_SERVER_URL?.trim()]
+      // Cloudflare R2 public bucket — the canonical (and sole) object-storage
+      // backend for media originals (GAP-208); next/image resizes on demand.
+      // The retired Vercel Blob + never-canonical Cloudinary entries were
+      // removed with the R2 cutover.
+      ...[process.env.R2_PUBLIC_BASE_URL?.trim(), process.env.NEXT_PUBLIC_SERVER_URL?.trim()]
         .filter(Boolean)
         .map((item) => {
           try {
--- a/apps/admin/src/__tests__/proxy-csp.test.ts
+++ b/apps/admin/src/__tests__/proxy-csp.test.ts
@@ -57,6 +57,27 @@ describe('admin proxy — CSP nonce (GAP-219)', () => {
     expect(directive(csp, 'img-src')).toContain('https://*.stripe.com');
   });
 
+  it('allows blob: in img-src so media upload previews can render (GAP-208)', async () => {
+    const res = await proxy(new NextRequest('https://admin.example.com/login'));
+    const imgSrc = directive(res.headers.get('content-security-policy') ?? '', 'img-src');
+    expect(imgSrc).toContain('blob:');
+  });
+
+  it('adds the R2 public origin to img-src when R2_PUBLIC_BASE_URL is set (GAP-208)', async () => {
+    vi.stubEnv('R2_PUBLIC_BASE_URL', 'https://pub-abc123.r2.dev/media');
+    const res = await proxy(new NextRequest('https://admin.example.com/login'));
+    const imgSrc = directive(res.headers.get('content-security-policy') ?? '', 'img-src');
+    expect(imgSrc).toContain('https://pub-abc123.r2.dev');
+    vi.unstubAllEnvs();
+  });
+
+  it('carries no Cloudinary sources and locks object-src to none (R2 is the sole backend)', async () => {
+    const res = await proxy(new NextRequest('https://admin.example.com/login'));
+    const csp = res.headers.get('content-security-policy') ?? '';
+    expect(csp).not.toContain('cloudinary.com');
+    expect(directive(csp, 'object-src')).toContain("'none'");
+  });
+
   it('allows the marketing origin in frame-src so the edit-session canvas can frame the preview', async () => {
     const res = await proxy(new NextRequest('https://admin.example.com/login'));
     const frameSrc = directive(res.headers.get('content-security-policy') ?? '', 'frame-src');
@@ -151,6 +172,8 @@ describe('admin proxy — CSP fleet mode (GAP-290)', () => {
     const imgSrc = directive(res.headers.get('content-security-policy') ?? '', 'img-src');
     expect(imgSrc).not.toContain('stripe.com');
     expect(imgSrc).not.toContain('cloudinary.com');
+    // blob: previews are origin-local; fleet kits keep them (GAP-208).
+    expect(imgSrc).toContain('blob:');
   });
 
   it('strips Stripe from frame-src and connect-src in fleet mode', async () => {
--- a/apps/admin/src/lib/security/csp.ts
+++ b/apps/admin/src/lib/security/csp.ts
@@ -15,7 +15,7 @@
  *     `Content-Security-Policy` header and applies it to its framework, bundle,
  *     and inline bootstrap scripts (plus `<Script>` components).
  *   - We deliberately KEEP the external-script host allowlist (Stripe, Maps,
- *     Cloudinary, Vercel insights) in non-fleet mode and do NOT use
+ *     Vercel insights) in non-fleet mode and do NOT use
  *     `'strict-dynamic'`, which would make those host-sources be ignored. In
  *     fleet mode (`REVEALUI_FLEET_MODE=true`) the hosted third-party SDK domains
  *     are omitted — fleet kits are self-hosted and must not phone home to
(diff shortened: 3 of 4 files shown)


RevealUIStudio:

Non-author review of the admin CSP / `next/image` change. I read the full diff, tried to refute the change on loosening, breakage, and correctness grounds, and independently re-ran the red/green proof rather than trusting the PR's test-plan claim.

---

### Independent red/green evidence

Reverted only the two source files to `origin/test` (tests left at HEAD), then re-ran the suite:

```
git restore --source=origin/test -- apps/admin/src/lib/security/csp.ts apps/admin/src/proxy.ts
pnpm --filter admin exec vitest run src/__tests__/proxy-csp.test.ts
```

**RED — 4 failed / 19 passed (23).** The four failures are exactly the new/updated assertions, each failing for the right reason (not an import or setup error):

| Failing assertion | Observed on `origin/test` |
|---|---|
| `img-src` contains `blob:` (hosted) | `img-src 'self' https://res.cloudinary.com https://*.stripe.com data: https://www.gravatar.com` |
| `img-src` contains the R2 origin | same string — no `https://pub-abc123.r2.dev` |
| CSP contains no `cloudinary.com` + `object-src 'none'` | `script-src … https://res.cloudinary.com …; img-src 'self' https://res.cloudinary.com …; object-src https://res.cloudinary.com` |
| `img-src` contains `blob:` (fleet) | `img-src 'self' data: https://www.gravatar.com` |

Restored to HEAD and re-ran: **GREEN — 23 passed (23).** Worktree verified clean at HEAD afterwards (`git status --short` and `git diff HEAD --stat` both empty). The PR's 4-red / 23-green claim is accurate.

---

### Loosening risk: net tightening, not loosening

- **`blob:` in `img-src`** is not a meaningful XSS or exfil surface. `img-src` governs image fetches only, it never enables script execution, and `blob:` URLs are minted by `URL.createObjectURL` against the document's own origin, so they are origin-local with no outbound channel. An attacker who could mint a blob URL in the admin document already has script execution, at which point this directive is irrelevant.
- **The env-derived R2 origin in `img-src`** adds one operator-controlled image host. It is theoretically an exfil channel in the generic `` sense, but the operator's own R2 public bucket gives an attacker no request log to read back, and the origin is not attacker-controlled. This is the minimum grant that makes the feature work.
- **`script-src` loses `https://res.cloudinary.com`** — strictly tighter.
- **`object-src 'none'`** is tighter in *both* modes. Hosted previously allowed `https://res.cloudinary.com`; fleet mode previously emitted no `object-src` at all and therefore fell back to `default-src 'self'`, permitting same-origin plugin content. Both now resolve to `'none'`.
- **`remotePatterns` loses `*.blob.vercel-storage.com`** — a real SSRF-surface reduction, not just cleanup. That wildcard let `/_next/image` fetch from *any* subdomain of a shared multi-tenant CDN, i.e. any Vercel Blob customer's bucket. The replacement is a single operator-pinned host. I confirmed there is no `unoptimized`, no `loaderFile`, and no `dangerouslyAllowSVG` in `apps/admin/next.config.mjs`, so `remotePatterns` is genuinely enforced.

### Breakage risk: no surviving consumers

Greps across `apps/` + `packages/` (excluding `node_modules` and `dist`):

- **Cloudinary** — zero consumers. The only three hits in the tree are the negative assertions in `apps/admin/src/__tests__/proxy-csp.test.ts` (lines 77, 159, 174). Nothing loads a Cloudinary script or image, so both the `script-src` and `img-src` removals are dead-config deletions.
- **`<object>` / `<embed>`** — zero occurrences in `apps/admin`, `packages/core`, `packages/presentation`, `packages/editor`. The only textual hit is the new explanatory comment in `csp.ts:176`. `object-src 'none'` breaks nothing.
- **Vercel Blob** — no source consumer. The surviving hits are a server-route test fixture (`apps/server/src/routes/__tests__/content-media-upload.test.ts:105,142,313`) and a doc-currency gate rule string. See the legacy-data caveat below, which is about stored rows rather than code.

The fix also targets the correct code path. The media library renders stored media through **raw ``**, not `next/image` (`apps/admin/src/app/(backend)/media/page.tsx:194` grid thumbnail and `:267` preview modal, both carrying a `biome-ignore lint/performance/noImgElement` for exactly this reason). That path is governed solely by `img-src`, which is why the R2 origin had to be added there and why `remotePatterns` alone would not have fixed the reported breakage.

### Correctness: `originOf` is safe on hostile input

The security-relevant question is whether an operator-supplied `R2_PUBLIC_BASE_URL` can break out of its `img-src` token and inject a CSP directive. I probed `originOf` directly against hostile values. It cannot:

| Input | `origin` |
|---|---|
| `https://evil.com; script-src *` | `""` (parse fails) |
| `https://evil.com' 'unsafe-inline` | `""` |
| `https://evil.com\n; script-src *` | `""` |
| `https://evil.com/ ; script-src *` | `https://evil.com` (path discarded) |
| `not a url` / `""` / `"   "` | `""` |
| `https://Ünicode.example.com/media` | `https://xn--nicode-2ya.example.com` (punycoded) |

No output ever contained a space, `;`, `'`, or newline — the URL parser normalises the host and discards everything after it, so CSP-delimiter injection is not reachable. The `try/catch` means malformed input yields `''` and the entry is dropped, so the Edge proxy cannot throw on bad env; that matters because a throw in `proxy.ts` would fail every admin request. The implementation also honours the no-regex posture by using the `URL` parser.

Fleet mode behaves correctly: `blob:` and the R2 origin are retained (both are self-hosted-appropriate), Stripe is stripped, and `object-src 'none'` now applies where no `object-src` existed before.

The new `r2PublicBaseUrl` option is additive with a `''` default, so the exported `buildAdminCsp` contract stays backwards-compatible.

### CI

Every substantive check is green (Quality, Typecheck, Build, CodeQL, Gitleaks, Dependency Review, Security Gate, Integration Tests). The one red check, **Security review gate**, fails with `HOLD — PR #2224 touches a security-sensitive surface with NO recorded reviewer verdict` — that is this gate awaiting this comment, not a finding.

---

### One non-security caveat the owner should decide on before merging

This is **not** a security blocker and it does not change the verdict, but it is the one real behavioural regression risk in the diff and it deserves an explicit decision rather than silence.

Removing `*.blob.vercel-storage.com` from `remotePatterns` can turn a surviving legacy media row from "renders" into "broken image":

- **The rows can still exist.** No migration or backfill ever rewrote `media.url`. The `media` DDL (`packages/db/migrations/0000_init.sql:242-259`) is a plain `url text NOT NULL`, never altered, and the only data-mutating statements across all 31 migrations are search-vector backfills. The schema comment is still pre-cutover: `packages/db/src/schema/admin.ts:102` reads `// Storage URL (Vercel Blob or other)`. No seed or fixture inserts media rows.
- **The cutover retired the backend, not the data.** `packages/core/CHANGELOG.md:62-69,116-121` and `packages/core/src/storage/index.ts:7-8` describe removing the provider, the `'vercel-blob'` tag, and `BLOB_READ_WRITE_TOKEN`. None of them claims existing objects were copied or existing URLs rewritten.
- **The render path is live.** `apps/admin/src/lib/components/RichText/index.tsx:157,165` (`renderUpload`) passes the stored absolute URL straight to `next/image` with no host normalisation. Note this path was *not* previously gated by `img-src` — `/_next/image` is same-origin, so `'self'` covered it and `remotePatterns` was the only allow-list in play. Removing the pattern makes the optimizer return `400 hostname is not configured under images` for those rows.

I still think the removal is the right call, and not something to revert: `*.blob.vercel-storage.com` is a wildcard across a **shared multi-tenant CDN**, so keeping it allow-listed leaves `/_next/image` able to fetch any Vercel Blob customer's bucket. That is a genuine SSRF surface, and the correct remedy for legacy rows is a data migration, not re-widening the allow-list.

The authoritative check is one query against prod, which nothing in the repo can answer:

```sql
SELECT count(*) FROM media WHERE url LIKE '%blob.vercel-storage.com%';
```

If that returns `0`, this caveat is moot and the diff is clean as-is. If it returns non-zero, the follow-up is a URL-rewrite migration (worth its own GAP), not a change to this PR.

Unrelated pre-existing bug spotted while tracing that path, for whatever backlog it belongs in: `apps/admin/src/lib/components/Media/ImageMedia/index.tsx:39` builds `` `${process.env.NEXT_PUBLIC_SERVER_URL}${url}` ``, but `media.url` is stored as the provider's **absolute** URL (`apps/server/src/routes/content/media.ts:136-142`, `packages/core/src/storage/object-storage.ts:99`). That concatenation mangles every absolute URL, R2 and legacy alike. Out of scope here.

### Blockers

None.

### Suggestions (non-blocking, follow-up is fine)

1. `apps/admin/src/lib/security/csp.ts:73` — `new URL('javascript:alert(1)').origin` and `new URL('data:…').origin` both return the **literal string `"null"`**, which is truthy and gets emitted into `img-src` as a bogus `null` source. This is not exploitable (CSP3 ignores source expressions that fail the grammar, and it does not widen the policy), but filtering `origin === 'null'` alongside `''` would keep the header free of garbage tokens on a misconfigured env.
2. No test covers the malformed-env branch. A case asserting that a junk `R2_PUBLIC_BASE_URL` neither throws nor adds a source would lock in the `try/catch` behaviour that keeps the Edge proxy alive — currently that safety property is only proven by inspection.

### Nits

3. `apps/admin/src/__tests__/proxy-csp.test.ts:67-71` — the R2 test calls `vi.stubEnv` and then `vi.unstubAllEnvs()` inline as the last statement, but the enclosing `describe` has no `afterEach` (unlike the fleet-mode block at line 149, which does). If that assertion ever fails, the stub leaks into every later test in the file. Impact is low, but moving the cleanup into an `afterEach` matches the file's own existing pattern.

PR #2225, opened 2026-07-27
fix(auth): coerce createdAt on sign-in + seed bootstrap admin emailVerified (GAP-446)

RevealUIStudio:
## Problem (the SECOND fresh-deploy login bug, production-Postgres only)
After #2223 fixed the swallowed-error path, a fresh unlicensed self-host still 500'd on the correct-credentials login. Root cause: an unverified account within the 24h email-verification grace window computes `accountAge` from `user.createdAt`, but the Postgres storage path returns `createdAt` as a **string**, so `createdAt.getTime()` threw a TypeError → outer catch → "unexpected error". The bootstrap admin is seeded with `emailVerified` defaulting to false, so this hit the very first login every time. The prior regression test used an in-memory DB that returned a Date, so it never reproduced this.

## Fix
- `packages/auth/src/server/auth.ts`: `new Date(user.createdAt).getTime()` — tolerates a string or a Date.
- `apps/admin/revealui.config.ts`: seed the operator's env-driven bootstrap admin with `emailVerified: true` (trusted account, no mailbox to verify on a self-host), so sign-in never gates the operator out after the 24h grace window.
- Regression test: `signIn` with `createdAt` as an ISO string succeeds — proven red before the coerce, green after.

## Notes
- Pre-existing on `test` (not this PR): 3 `signUp` unit-test failures (signUp source untouched here) — worth a separate look.
- **Crown-jewel auth surface. REVIEW-PENDING under guardrail 2 — do not merge until a non-author APPROVE verdict is recorded.**

### DIFF
--- a/apps/admin/revealui.config.ts
+++ b/apps/admin/revealui.config.ts
@@ -353,31 +353,36 @@ export default buildConfig({
             // — 'super-admin' here violates users_role_check and the create fails.
             role: 'owner',
             roles: ['super-admin'],
-            // The operator sets this account via REVEALUI_ADMIN_EMAIL/PASSWORD;
-            // it is trusted by definition and there is no mailbox to verify on a
-            // self-host. Mark it verified so sign-in does not gate the operator
-            // out after the 24h email-verification grace window.
-            emailVerified: true,
           },
         });
 
         revealui.logger.info(`First admin user created: ${adminEmail}`);
 
-        // Stamp TOS acceptance on the typed tos_accepted_at / tos_version
-        // columns: revealui.create() above can't persist them (the engine's
-        // dynamic-SQL adapter rejects camelCase column identifiers), so an
-        // onInit-created owner would otherwise have a NULL tos_accepted_at.
-        // Record them via a typed Drizzle write, mirroring the web setup route
-        // (api/setup) and the CLI (scripts/admin/bootstrap). Non-fatal — a
+        // Stamp TOS acceptance AND email verification on their typed columns.
+        // revealui.create() above cannot persist these (the engine's dynamic-SQL
+        // adapter rejects camelCase column identifiers), so an onInit-created
+        // owner would otherwise have NULL tos_accepted_at and email_verified
+        // stuck at its `false` default. The operator sets this account via
+        // REVEALUI_ADMIN_EMAIL/PASSWORD, so it is trusted by definition and there
+        // is no mailbox to verify on a self-host; marking it verified keeps
+        // sign-in from gating the operator out after the 24h grace window.
+        // Both are typed Drizzle writes (which DO map camelCase correctly),
+        // mirroring the web setup route (api/setup) and the CLI. Non-fatal — a
         // failure here is a backfillable gap, not a reason to abort startup.
         // Imports are lazy because pulling @revealui/db/client into the
         // top-level module graph causes Turbopack async-module-init issues
         // (see the poolFactory note near the top of this file).
         try {
           const { getClient } = await import('@revealui/db/client');
+          const { eq } = await import('@revealui/db/orm');
+          const { users } = await import('@revealui/db/schema');
           const { stampTosAcceptanceByEmail } = await import('@/lib/auth/tos');
           const db = getClient('rest');
           await stampTosAcceptanceByEmail(db, adminEmail);
+          await db
+            .update(users)
+            .set({ emailVerified: true, emailVerifiedAt: new Date() })
+            .where(eq(users.email, adminEmail));
         } catch (tosError) {
           const message = tosError instanceof Error ? tosError.message : String(tosError);
           revealui.logger.error(
--- a/apps/admin/revealui.config.ts
+++ b/apps/admin/revealui.config.ts
@@ -353,6 +353,11 @@ export default buildConfig({
             // — 'super-admin' here violates users_role_check and the create fails.
             role: 'owner',
             roles: ['super-admin'],
+            // The operator sets this account via REVEALUI_ADMIN_EMAIL/PASSWORD;
+            // it is trusted by definition and there is no mailbox to verify on a
+            // self-host. Mark it verified so sign-in does not gate the operator
+            // out after the 24h email-verification grace window.
+            emailVerified: true,
           },
         });
 
--- a/packages/auth/src/server/__tests__/auth.test.ts
+++ b/packages/auth/src/server/__tests__/auth.test.ts
@@ -168,6 +168,27 @@ describe('auth', () => {
       expect(result.sessionToken).toBe('session-token-abc');
     });
 
+    it('signs in when createdAt is a string, not a Date (Postgres storage path)', async () => {
+      // Regression (GAP-446): the Postgres storage path returns createdAt as an
+      // ISO string, not a Date object. With emailVerified:false the sign-in reads
+      // createdAt to compute account age; calling .getTime() on a string threw a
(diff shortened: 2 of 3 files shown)


RevealUIStudio:

Adversarial review by a non-author reviewer. I read the full diff, the surrounding `signIn` path, the seed `onInit` guard, and the `users` schema, and I ran the touched test file on both the branch and `origin/test` in scratch worktrees. The change is correct and does not weaken the auth surface.

### 1. `emailVerified: true` on the bootstrap admin seed — SAFE, not a weakness
The seed runs only inside the `existingUsers.totalDocs === 0` block in `apps/admin/revealui.config.ts` (`onInit`), with credentials read from server env (`REVEALUI_ADMIN_EMAIL` / `REVEALUI_ADMIN_PASSWORD`) and `overrideAccess: true`. There is no user-input path into it: it fires exactly once, on a fresh empty DB, from operator-controlled process env. An attacker cannot drive this to bootstrap a verified account — they control neither the env nor the "zero users" precondition. On a self-host there is no mailbox to verify against, so marking the operator-trusted first admin verified is the correct posture and prevents the operator locking themselves out after the 24h grace window. No security concern.

### 2. `new Date(user.createdAt).getTime()` — correct; NaN fail-open path is unreachable
- Legitimate `Date`: `new Date(dateObj).getTime()` is identical to the old behavior. No change.
- Postgres ISO string (the bug): parses correctly. Fixes the production-only TypeError that fell through to the outer catch and turned a valid login into `unexpected_error`.
- I checked the theoretical fail-open: `new Date(undefined)` → Invalid Date → `getTime()` NaN → `NaN > grace` is `false` → the verification gate is skipped and login proceeds. This is a genuine direction change (the old code *threw* on a missing `createdAt`, which failed closed). **However it is unreachable:** `users.createdAt` is `timestamp(...).defaultNow().notNull()` (`packages/db/src/schema/users.ts:91`), so a persisted user always carries a valid server-assigned timestamp, and `createdAt` is never attacker-controllable. Not exploitable.
- Suggestion (defense-in-depth, non-blocking): guard the coercion so an unparseable `createdAt` fails safe rather than open, e.g. treat `Number.isNaN(accountAge)` as "grace expired" (block) instead of relying on the NOT NULL invariant. Cheap insurance against a future schema/read-path change.

### 3. Regression test genuinely reproduces
Proven empirically. Reverting only the `auth.ts` coercion and re-running the new test goes red with exactly `AssertionError: expected 'unexpected_error' not to be 'unexpected_error'` — i.e. the string `.getTime()` throws and is caught as `unexpected_error`. With the fix it is green. Real red→green, not a tautology.

### 4. No other unguarded date coercion on the sign-in path
The only `.getTime()` on a DB-string field in `signIn` is the fixed line. The other `new Date(...)` uses in `auth.ts` (lines 393/464+) are in `signUp`, constructing fresh server-side dates, not coercing stored/user-supplied values. `signIn`'s remaining reads are booleans (`mfaEnabled`, `mustRotatePassword`). Clean.

### 5. Tests: new test green; the 3 signUp failures are pre-existing
Branch: 41 passed / 3 failed (44). `origin/test`: 40 passed / **the same 3** failed (43) — `signUp > provisions a personal account...`, `signUp > creates user and returns session token...`, `signUp > returns raw email verification token...`. The diff (`--stat`: 3 files, +32/-2) does not touch `signUp` at all, so these are pre-existing and unrelated. The PR adds exactly one passing `signIn` test.

### Non-blocking notes
- The seed's intended benefit (no lockout after 24h) depends on `revealui.create()` actually persisting `emailVerified`. The adjacent TOS comment in the same file documents that this engine's dynamic-SQL adapter *drops* camelCase column writes (`tos_accepted_at` had to be stamped via a separate typed Drizzle write). Worth verifying `email_verified` isn't silently dropped the same way — if it is, the operator still gets grace-window fixed by change #1 but could hit the post-grace lockout this seed is meant to prevent. This is a functional-efficacy question, not a security one, and it fails closed either way.

Approving the security surface. Change is narrow, correct, and fails safe on every reachable input.

RevealUIStudio:
## Guardrail-2 security verdict: APPROVE (updated — supersedes my prior APPROVE, covers the complete fix through 256ea8a76)

Re-reviewed the delta `89d4d776d..256ea8a76` (`apps/admin/revealui.config.ts`) that addresses the functional flag from my first pass — the engine's raw-SQL `create()` drops camelCase columns, so the inline `emailVerified: true` in the create data was a no-op. The delta removes it and persists verification via a typed Drizzle write in `onInit`. Clean on all three checks.

### 1. The typed write actually persists `email_verified` — confirmed
```ts
await db.update(users).set({ emailVerified: true, emailVerifiedAt: new Date() })
        .where(eq(users.email, adminEmail));
```
- **Table/columns correct:** `users.emailVerified` → `email_verified` (`boolean('email_verified').default(false).notNull()`, `packages/db/src/schema/users.ts:56`) and `users.emailVerifiedAt` → `email_verified_at` (`:62`). Drizzle maps camelCase fields to snake_case columns, which is exactly why the typed path persists where `create()` cannot.
- **Imports resolve:** `eq` from `@revealui/db/orm` and `users` from `@revealui/db/schema` are both real subpath exports (`packages/db/package.json`), and this is the identical import shape the existing TOS stamp uses.
- **Same working precedent:** `stampTosAcceptanceByEmail` (`apps/admin/src/lib/auth/tos.ts`) does `db.update(users).set({...}).where(eq(users.email, email))` on this same `getClient('rest')` Database. The new write is byte-for-byte the same pattern on the same client — so it is proven to persist, not hoped to.
- **Where target correct:** `adminEmail` is the email the row was just created with, and `users.email` is unique; the update hits exactly the bootstrap admin row.

### 2. Still guarded, non-fatal, and not attacker-reachable — confirmed
- Unchanged precondition: the whole block runs only inside `existingUsers.totalDocs === 0`, with creds from server env (`REVEALUI_ADMIN_EMAIL`/`REVEALUI_ADMIN_PASSWORD`). No user-input path; `adminEmail` in the `where` is server-env, not attacker-supplied. An attacker controls neither the "zero users" precondition nor the env, so this cannot be driven to verify an arbitrary account.
- Non-fatal: the write sits in the existing `try { ... } catch (tosError)` that logs and continues ("backfillable gap, not a reason to abort startup"). A DB hiccup here does not abort startup and does not fail open into any auth decision.

### 3. No new issue
- If `create()` rolled back and no row exists, the update is a safe no-op (same as the TOS stamp). Ordering after `stampTosAcceptanceByEmail` is fine — both are best-effort typed writes; there's no security-relevant coupling between them.
- Setting `emailVerifiedAt` alongside the flag is consistent with the read model.

### Carry-over (unchanged, already approved and empirically verified)
- The `new Date(user.createdAt)` coercion in `packages/auth/src/server/auth.ts` and its regression test are untouched by this delta. I previously proved that test genuinely goes red without the fix (`unexpected_error`) and green with it, and that `createdAt` is a NOT NULL server-assigned column so the theoretical NaN fail-open is unreachable. The optional `Number.isNaN` defense-in-depth suggestion still stands as non-blocking.
- Noted: coordinator reports the 3 `signUp` failures I saw were an unbuilt-workspace artifact (537 passing once built). Consistent with what I observed — they were identical on `origin/test` and the diff never touches `signUp`, so they are not introduced here regardless; the security-load-bearing `signIn` test passes.

The functional flag is now correctly closed using the proven typed-write path, with no security regression. Approving the complete fix.

PR #2226, opened 2026-07-27
fix(server): anchor null-tenant audit rows under the __system__ scope (GAP-427)

RevealUIStudio:
## Summary

Owner ruling 2026-07-27: null-tenant (system) audit rows MUST be anchored,
under a dedicated system scope. Before this change the anchor sweep skipped
null-tenant rows entirely, so prod (1725 signed null-tenant rows, zero tenant
rows) could never anchor a single root.

Adds `SYSTEM_ANCHOR_SCOPE = '__system__'` as the `audit_anchors.tenant` value
for anchors over null-tenant `audit_log` rows. `audit_anchors.tenant` is
`NOT NULL`, so the sentinel avoids a schema migration; the underlying
`audit_log.tenant` column stays null on the rows themselves — only the anchor
row's `tenant` column carries the sentinel.

- `runAuditAnchorSweep` runs one additional system-scope pass after the
  per-tenant loop. It reuses the same `anchorTenantBatch` code path (floor
  derivation, contiguity, batch readiness, Merkle root, signing, insert),
  generalized to filter `audit_log` by `isNull(tenant)` instead of
  `eq(tenant, tenant)` when in system mode.
- The system pass is **not** entitlement-gated (`canAnchorTenant` is
  per-account; system events are the operator's own audit trail) and
  **never** writes a usage meter (no account FK backs a null tenant) —
  `doMeter` is forced `false` at the call site.
- The tenant enumeration loop defensively skips any row whose `tenant`
  literally equals `'__system__'`, so a hostile or buggy writer can't
  collide with the system scope's anchor bookkeeping.
- Legacy-floor semantics are identical to the per-tenant case: while the
  system scope has zero anchors, the floor is the highest **unsigned**
  null-tenant `seq`; rows at or below it are pre-enforcement legacy and are
  never retro-signed.
- `AnchorSweepResult` gains `systemOutcome: 'inserted' | 'waiting' |
  'skipped'`; a system insert also counts into `anchorsInserted`. The tick
  log line gains ` system=<outcome>`.
- Removed the old `logger.warn(... "(skipped)")` for null-tenant rows (the
  doctrine changed — they're no longer skipped). `nullTenantSignedRows` and
  the `revealui_audit_null_tenant_signed_rows` gauge are kept, with the gauge
  help text updated to describe the system scope.
- `packages/db/src/schema/audit-anchors.ts`: comment-only change on the
  `tenant` column describing the sentinel. No migration — the column type
  and constraints are unchanged.
- `docs/security/AUDIT_RECEIPTS.md`: new "System scope (GAP-427 ruling
  2026-07-27)" subsection under the legacy-floor documentation.

### Deviations from spec / design notes

- `result.tenantsFloorEngaged` also counts the system pass when it floor-
  engages (per spec point 6, "the floorEngaged metric/log applies" to the
  system scope identically to tenants). The field name still says "tenants"
  since no new field was requested for this.
- The `mWaiting` / `mGaps` / `mErrors` Prometheus counters (help text still
  says "Tenants"/"tenant") are **not** incremented for the system pass, to
  avoid silently redefining what those metrics count. The system pass's
  outcome is fully observable via `result.systemOutcome` and the extended
  tick log line instead. `mAnchorsCreated` and `mFloorEngaged` **are** shared
  (generic wording / explicitly requested).
- `GET /api/audit/anchors` (`apps/server/src/routes/audit.ts`) was **not**
  changed. Its `tenant` is derived entirely server-side via
  `getUserAccountId` (an `accountMemberships` lookup) — there is no request
  parameter a client can set to `'__system__'`. The existing tenant-isolation
  check on `GET /anchors/:id/proof` (`anchor.tenant !== tenant → 404`)
  already covers a system-scope anchor the same way it covers any other
  tenant's anchor. Added one test (`audit-anchors.test.ts`) locking this in
  by inserting a `'__system__'`-tenant anchor and asserting the proof
  endpoint 404s for a normal tenant caller.

## Prove-red

After committing, restored the pre-fix `audit-anchor-sweep.ts` from `HEAD`
(the branch's own prior state, i.e. `origin/test`) and re-ran the sweep test
file:

- **Red** (pre-fix `audit-anchor-sweep.ts`): 4 failed / 13 passed — the 4
  new/rewritten GAP-427 system-scope tests failed (`systemOutcome` was
  `undefined`, `anchorsInserted` missing the system anchor); the 13
  pre-existing tests were unaffected.
- **Green** (fix restored from `HEAD`): 17/17 passed (sweep) + 9/9 passed
  (routes) = 26/26.

## Test plan

- [x] `pnpm --filter server exec vitest run src/jobs/__tests__/audit-anchor-sweep.test.ts` — 17/17 passed
- [x] `pnpm --filter server exec vitest run src/routes/__tests__/audit-anchors.test.ts` — 9/9 passed
- [x] Prove-red: 4/17 failed on base code, 17/17 green with the fix
- [x] `biome check` clean on all 5 changed files
- [x] `pnpm --filter server typecheck` — zero errors in any of the 5 changed
      files (remaining repo-wide errors are pre-existing, from workspace
      packages whose dists weren't built in this worktree — unrelated to
      this change)
- [x] Pre-push gate (security + CI phases) passed

**Security-classed (audit anchoring surface): needs a non-author guardrail-2 verdict before merge — review-pending.**

### DIFF
--- a/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
+++ b/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
@@ -11,7 +11,7 @@ import {
 } from '@revealui/core/security';
 import { type AuditEntry, type AuditRowSignerFn, DrizzleAuditStore } from '@revealui/db';
 import type { Database } from '@revealui/db/client';
-import { auditAnchors } from '@revealui/db/schema';
+import { auditAnchors, usageMeters } from '@revealui/db/schema';
 import { eq } from 'drizzle-orm';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
@@ -19,7 +19,7 @@ import {
   type TestDb,
 } from '../../../../../packages/test/src/utils/drizzle-test-db.js';
 import { planContiguousBatch, type SignedAuditRow } from '../audit-anchor-batch.js';
-import { runAuditAnchorSweep } from '../audit-anchor-sweep.js';
+import { runAuditAnchorSweep, SYSTEM_ANCHOR_SCOPE } from '../audit-anchor-sweep.js';
 
 const KID = 'kid-s4-3';
 
@@ -214,7 +214,7 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     expect(await db.select().from(auditAnchors)).toHaveLength(0);
   });
 
-  it('counts null-tenant signed rows and never anchors them', async () => {
+  it('GAP-427: counts null-tenant signed rows and anchors them under the system scope, not per-tenant', async () => {
     const store = new DrizzleAuditStore(db, canonicalSignerFn(priv));
     await store.append(makeEntry({ tenant: null, id: 'null-tenant-1' }));
     await store.append(makeEntry({ tenant: 'acct_max', id: 't1' }));
@@ -231,10 +231,117 @@ describe('runAuditAnchorSweep (PGlite)', () => {
     });
 
     expect(result.nullTenantSignedRows).toBe(1);
-    const anchors = await db.select().from(auditAnchors);
-    expect(anchors.every((a) => a.tenant !== null && a.tenant.length > 0)).toBe(true);
-    // null-tenant row is not in any anchor range as a tenant key
-    expect(anchors.some((a) => a.tenant === '')).toBe(false);
+    // The per-tenant loop never treats the null-tenant row as a "tenant" key.
+    const tenantAnchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, 'acct_max'));
+    expect(tenantAnchors).toHaveLength(1);
+    // It anchors separately, via the system-scope pass.
+    expect(result.systemOutcome).toBe('inserted');
+    const systemAnchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, SYSTEM_ANCHOR_SCOPE));
+    expect(systemAnchors).toHaveLength(1);
+    expect(systemAnchors[0]?.leafCount).toBe(1);
+  });
+
+  it('GAP-427: system scope anchors signed null-tenant rows above the null-tenant unsigned floor', async () => {
+    const signedStore = new DrizzleAuditStore(db, canonicalSignerFn(priv));
+    const unsignedStore = new DrizzleAuditStore(db); // pre-enforcement legacy rows, no signer
+
+    // Legacy era (closed): unsigned null-tenant seq 1. Signed era: seq 2..3.
+    await unsignedStore.append(makeEntry({ id: 'sys-u1', tenant: null }));
+    await signedStore.append(makeEntry({ id: 'sys-s2', tenant: null }));
+    await signedStore.append(makeEntry({ id: 'sys-s3', tenant: null }));
+
+    const result = await runAuditAnchorSweep({
+      db,
+      signer: cryptoSigner,
+      batchSize: 2,
+      canAnchorTenant: async () => true,
+      recordMeter: false,
+      now: () => new Date('2026-07-23T12:00:10.000Z'),
+    });
+
+    expect(result.systemOutcome).toBe('inserted');
+    expect(result.tenantsFloorEngaged).toBe(1);
+    expect(result.anchorsInserted).toBe(1);
+
+    const anchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, SYSTEM_ANCHOR_SCOPE));
+    expect(anchors).toHaveLength(1);
+    expect(anchors[0]?.seqFrom).toBe(2);
+    expect(anchors[0]?.seqTo).toBe(3);
+
+    // No account FK backs a null tenant — the system pass must never meter.
+    const meters = await db.select().from(usageMeters);
+    expect(meters).toHaveLength(0);
+  });
+
+  it('GAP-427: systemOutcome is skipped when there are no null-tenant rows', async () => {
+    await appendSigned(3, 'acct_max', new Date('2026-07-23T12:00:00.000Z'));
+
+    const result = await runAuditAnchorSweep({
+      db,
+      signer: cryptoSigner,
+      batchSize: 3,
+      canAnchorTenant: async () => true,
+      recordMeter: false,
+      now: () => new Date('2026-07-23T12:00:10.000Z'),
+    });
+
+    expect(result.systemOutcome).toBe('skipped');
+    expect(result.nullTenantSignedRows).toBe(0);
+    const systemAnchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, SYSTEM_ANCHOR_SCOPE));
+    expect(systemAnchors).toHaveLength(0);
+  });
+
+  it('GAP-427: a tenant anchor and the system anchor both land in one sweep', async () => {
+    // Non-interleaved seq runs (tenant rows first, then null-tenant rows) —
+    // the global-seq interleave limitation is pre-existing and out of scope.
+    await appendSigned(2, 'acct_max', new Date('2026-07-23T12:00:00.000Z'));
+    const store = new DrizzleAuditStore(db, canonicalSignerFn(priv));
+    await store.append(
+      makeEntry({ id: 'sys-1', tenant: null, timestamp: new Date('2026-07-23T12:00:01.000Z') }),
+    );
+    await store.append(
+      makeEntry({ id: 'sys-2', tenant: null, timestamp: new Date('2026-07-23T12:00:02.000Z') }),
+    );
+
+    const result = await runAuditAnchorSweep({
+      db,
+      signer: cryptoSigner,
+      batchSize: 2,
+      canAnchorTenant: async () => true,
+      recordMeter: false,
+      now: () => new Date('2026-07-23T12:00:10.000Z'),
+    });
+
+    expect(result.anchorsInserted).toBe(2);
+    expect(result.systemOutcome).toBe('inserted');
+
+    const tenantAnchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, 'acct_max'));
+    expect(tenantAnchors).toHaveLength(1);
+    expect(tenantAnchors[0]?.seqFrom).toBe(1);
+    expect(tenantAnchors[0]?.seqTo).toBe(2);
+
+    const systemAnchors = await db
+      .select()
+      .from(auditAnchors)
+      .where(eq(auditAnchors.tenant, SYSTEM_ANCHOR_SCOPE));
+    expect(systemAnchors).toHaveLength(1);
+    expect(systemAnchors[0]?.seqFrom).toBe(3);
+    expect(systemAnchors[0]?.seqTo).toBe(4);
   });
 
   it('skips when an unsigned hole appears above an existing anchor (strict contiguity)', async () => {
--- a/apps/server/src/jobs/audit-anchor-sweep.ts
+++ b/apps/server/src/jobs/audit-anchor-sweep.ts
@@ -6,7 +6,15 @@
  * age ≥ max lag), build a contiguous batch, Merkle-root the signature
  * leaves, sign the root (Stage 3 Ed25519), insert audit_anchors, and
  * meter `audit_anchor`. Failures never delete audit rows (append-only).
- * Null-tenant rows are never anchored (§9); volume is gauged each tick.
+ *
+ * Null-tenant (system) rows anchor too, via one additional pass per sweep
+ * (GAP-427 ruling 2026-07-27): they land in `audit_anchors` under the
+ * reserved `SYSTEM_ANCHOR_SCOPE` ('__system__') sentinel tenant value.
+ * `audit_log.tenant` itself stays null on those rows (no schema migration);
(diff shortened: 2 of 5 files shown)


RevealUIStudio:

Non-author review of `fix/gap-427-system-anchor-scope` @ `194f19ac5`, read adversarially against the five integrity properties the PR claims. No blocker found. Findings below are hardening, not merge gates.

### Independent red/green (reproduced, not taken from the PR body)

Worktree `.wt/gap-427-system-scope`, base source restored via `git restore --source=origin/test` (no stash).

**RED** with the base `audit-anchor-sweep.ts`:

```
pnpm --filter server exec vitest run src/jobs/__tests__/audit-anchor-sweep.test.ts
Tests  4 failed | 13 passed (17)
```

The four reds are exactly the new GAP-427 assertions, and they fail on substance rather than on a collection error: `expected undefined to be 'inserted'` / `'skipped'` at `audit-anchor-sweep.test.ts:268` and `:297` (no `systemOutcome` on the base result), and `expected 1 to be 2` at `:327` (the system anchor missing from `anchorsInserted`). The 13 pre-existing tenant-scope tests stayed green under base code, which is the evidence that their meaning was not rewritten.

**GREEN** with the branch source restored:

```
pnpm --filter server exec vitest run \
  src/jobs/__tests__/audit-anchor-sweep.test.ts \
  src/routes/__tests__/audit-anchors.test.ts
Test Files  2 passed (2)
     Tests  26 passed (26)
```

Matches the claimed 17 + 9. Worktree left clean at `194f19ac5`. CI at review time: all required contexts pass except `Unit Tests`, still pending.

### 1. Sentinel safety (claim refuted? no)

The defensive skip at [audit-anchor-sweep.ts:207](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L207) does prevent both failure modes, and it does so structurally rather than by luck:

- **Double-anchoring:** the two row populations are provably disjoint. The tenant loop enumerates `isNotNull(auditLog.tenant)` ([:201](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L201)); the system pass matches `isNull(auditLog.tenant)` via `auditLogTenantMatch` ([:275-277](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L275-L277)). A hostile row carrying the literal string `'__system__'` is in neither set, so it can never reach a Merkle leaf in the system root. It is silently never anchored, which is a coverage loss, not a poisoning.
- **`lastAnchoredSeq` confusion:** `lastAnchoredSeq` keys on `audit_anchors.tenant` ([:279-286](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L279-L286)). Because the tenant loop `continue`s before calling `anchorTenantBatch`, no tenant pass can ever insert an `audit_anchors` row keyed `'__system__'`, so the system scope's contiguity baseline is derived only from its own writes. The unique index `audit_anchors_tenant_seq_range_uidx` on `(tenant, seq_from, seq_to)` ([audit-anchors.ts:63](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/db/src/schema/audit-anchors.ts#L63)) is a second net against a concurrent double-insert.

I also traced the write path rather than assuming it. The single door `DrizzleAuditStore.append` passes `tenant` through verbatim (`tenant: entry.tenant ?? null`, [audit-store.ts:190](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/db/src/audit-store.ts#L190)) with no validation, and `audit_log` carries no CHECK on `tenant`. Of all call sites, exactly one is client-influenced: `agent-dispatcher.ts:152` takes `resolution.workspaceId`, which originates in the raw `X-Tenant-ID` header (`middleware/tenant.ts:70`). Its format gate accepts `__system__` (underscore is inside `\w`). What actually stops it is the membership validator (`middleware/tenant.ts:122-147`), which admits only a value matching an `accounts.id` or `accounts.slug` the caller has an active membership on, and all three account-creation sites generate `id` with `crypto.randomUUID()` (`packages/auth/src/server/auth.ts:526`, `apps/server/src/routes/webhooks.ts:525`, `scripts/admin/grant-entitlement.ts:128`) while `buildAccountSlug` strips everything outside `[a-z0-9]` to dashes. Every other audit call site passes no tenant at all, and the `@revealui/ai` `AuditEntry` type has no `tenant` field, so those stores can only write NULL. **Bottom line: not reachable today.** See suggestion 1 for the residual.

### 2. Cross-tenant isolation (holds)

`GET /anchors` and `/anchors/:id/proof` both derive `tenant` server-side from `getUserAccountId` ([audit.ts:114](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/audit.ts#L114), [:203](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/audit.ts#L203)). There is no request parameter of any kind that sets the tenant. The list filters `eq(auditAnchors.tenant, tenant)` ([:128](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/audit.ts#L128)) and the proof route 404s on `anchor.tenant !== tenant` ([:221](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/audit.ts#L221)). The one attack I wanted to land was a collision: `accounts.id` is a free-form `text` column, not a `uuid`, so an account literally named `__system__` would be served every system anchor. All three creation paths use `crypto.randomUUID()` and none accepts a caller-supplied id, so it is unreachable. The new test at [audit-anchors.test.ts:327](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/__tests__/audit-anchors.test.ts#L327) is a real test, not a tautology: it inserts a genuine sentinel-keyed anchor over the caller's own seq range and asserts 404, so it would catch a future regression that dropped the tenant check.

### 3. Integrity semantics (preserved)

Strict contiguity is untouched: the system pass reuses `anchorTenantBatch` unchanged, so `planContiguousBatch` still returns `null` on `first.seq !== lastAnchoredSeq + 1` and `assertContiguousSeq` still guards the batch before signing. A genuine hole above an existing system anchor stalls the sweep exactly as it does for a tenant, and it stays observable, since `mGaps.inc()` and the gap warn at [:382-387](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L382-L387) fire for the system scope too.

Floor derivation matches tenant semantics: `legacyUnsignedFloor(db, tenant, isSystem)` swaps only the predicate to `isNull(auditLog.tenant)` ([:313](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L313)), and it is consulted only while `anchored === 0` ([:339](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L339)), so the floor can never span live history. The added `isSystem = false` default keeps the exported signature backwards compatible for the GAP-429 lag consumer at [audit.ts:161](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/routes/audit.ts#L161), which correctly keeps tenant semantics.

No retro-signing anywhere: nothing in the diff writes to `audit_log`, the store never issues an UPDATE, and migration 0026's `BEFORE UPDATE OR DELETE` trigger forbids one regardless.

Worth stating plainly for the record: `seq` is a table-wide `bigserial`, not per-scope, so strict last+1 contiguity only holds for a scope that owns an unbroken seq run. That is a pre-existing property of the whole anchoring design, it is not introduced or worsened here (null-tenant rows already consumed seq numbers before this change), the test at `:311` names it explicitly, and it is a non-issue for the deployment this PR targets (1725 signed null-tenant rows, zero tenant rows). Flagging it only because the system scope becomes load-bearing for the operator's own trail the day the first hosted account is provisioned. It fails visibly via `revealui_audit_anchor_seq_gaps_total`, so it is a follow-up gap, not a merge gate.

### 4. Signing payload (round-trips)

`auditAnchorSignableBytes` treats `tenant` as an opaque string: the only constraint is non-empty ([audit-merkle.ts:193](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/security/src/audit-merkle.ts#L193)), and it goes into the JCS canonicalization verbatim ([:214-221](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/security/src/audit-merkle.ts#L214-L221)). Nothing parses it, and nothing assumes an account id. Both offline paths agree: `verifyAuditAnchorOffline` copies `anchor.tenant` straight through ([audit-anchor-verify.ts:73](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/security/src/audit-anchor-verify.ts#L73)) and the CLI parses it as `String(o.tenant ?? '')` with only a non-empty check ([scripts/security/verify-audit-anchor.ts:131](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/scripts/security/verify-audit-anchor.ts#L131)). A `'__system__'` anchor verifies offline with no code change.

### 5. Regression (clean)

The 13 pre-existing sweep tests are byte-identical in meaning and pass under base code. The removed `"(skipped)"` warn is not load-bearing: a repo-wide grep for that line, for `nullTenantSignedRows`, and for `revealui_audit_null_tenant_signed_rows` across `.ts`/`.md`/`.json`/`.yml` returns only the sweep module, its tests, and the docs page. No dashboard, alert rule, or log-based consumer reads either. The gauge itself is retained with updated help text.

---

## Suggestions (none blocking)

1. **[audit-store.ts:190](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/packages/db/src/audit-store.ts#L190)** The sentinel is not rejected at the ONE DOOR. Today it is unreachable only because account ids are UUIDs and slugs cannot contain underscores, which means the property is enforced by mount-site config and slug shape rather than by the door. Any future `tenantMiddleware` mount without `validateTenant`, or an operator-created account slugged `__system__`, makes it writable. Anchoring already fails safe (finding 1), so this is defense in depth: a `throw` in `unsignedValues` when `entry.tenant === SYSTEM_ANCHOR_SCOPE`, or a CHECK constraint on `audit_log.tenant`, moves the guarantee into the schema. Would need the constant to live somewhere `@revealui/db` can import.
2. **[audit-anchor-sweep.ts:258-265](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L258-L265)** The system pass's catch pushes to `result.errors` and logs, but increments no counter. `mErrors` is deliberately left tenant-scoped per the PR body, which is defensible, but the net effect is that a persistently throwing system pass is invisible to Prometheus alerting on a crown-jewel surface. Either generalize `mErrors` help text and increment it, or add `revealui_audit_anchor_system_errors_total`. (Stalls are fine already: `mGaps` does fire for the system scope.)
3. **Test gap.** There is no test for the defensive skip at `:207`, which is the one hostile-writer scenario the PR body explicitly claims to defend. A test that appends a signed row with `tenant: '__system__'` plus genuine null-tenant rows, then asserts no `audit_anchors` row was produced from the tenant loop and that the system anchor's leaves cover only the null-tenant rows, would lock the disjointness property that finding 1 rests on. Cheap, and it is the assertion a future refactor is most likely to break.
4. **[audit-anchor-sweep.ts:279-286](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L279-L286)** `lastAnchoredSeq` cannot distinguish a sentinel-keyed anchor written by the new system pass from one the *old* tenant loop could have written over non-null `'__system__'` rows. If any such row pre-exists, the system scope would adopt a baseline derived from a different row population. Verified theoretical for the target deployment (zero tenant rows), but a one-time preflight `SELECT count(*) FROM audit_anchors WHERE tenant = '__system__'` before enabling the sweep is worth a line in the rollout note.
5. **[audit-anchor-sweep.test.ts:280](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts#L280)** The `expect(meters).toHaveLength(0)` assertion is weaker than it reads: the test passes `recordMeter: false`, so the tenant loop is not metering either. It would pass even if the system pass metered through the normal `doMeter` flag. Running that case with `recordMeter` defaulted to true and only null-tenant rows present would actually prove the forced `false` at `:252`.
6. **[AUDIT_RECEIPTS.md:65-84](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/docs/security/AUDIT_RECEIPTS.md#L65-L84)** The §Lag section directly above presents `GET /api/audit/anchors` as the retrieval surface, so a reader can reasonably infer system anchors appear there. They deliberately do not, and there is no operator route at all. One sentence saying system anchors are intentionally not exposed on the tenant API, and naming how an operator obtains the record to feed `pnpm verify:audit-anchor`, closes the loop.

## Nits

7. **[audit-anchor-sweep.ts:255-257](https://github.com/RevealUIStudio/revealui/blob/194f19ac5/apps/server/src/jobs/audit-anchor-sweep.ts#L255-L257)** Accounting is asymmetric: the system pass counts into `anchorsInserted` and `tenantsFloorEngaged` but not `tenantsConsidered` / `tenantsWaiting` / `tenantsSkipped`. Documented as intentional and `systemOutcome` covers it, but anyone summing the tick log will find `inserted` exceeding what the tenant counters explain.
8. No system-scope analogue of the "unsigned hole above an existing anchor" contiguity test at `:347`. Same code path, so low value, but it is the assertion that proves the system scope stalls rather than jumps.

PR #2229, opened 2026-07-27
fix(auth): scope required SESSION_COOKIE_DOMAIN to hosted deployments

RevealUIStudio:
## Root cause (the real bug, real Postgres only)

Every prior GAP-446 reproduction (#2223 DB-guard, #2225 createdAt coercion +
emailVerified persistence) ran against PGlite and passed. On a **real
Postgres** self-host (Railway, `pgvector/pgvector:pg16`), a fresh admin
still 500s on login with the **correct** credentials, even with those fixes
on `origin/test` (HEAD `ecc8cb751`):

- Wrong email → clean `401 INVALID_CREDENTIALS` (fails before the buggy code runs)
- Correct email → `500`, admin logs both `"Error signing in"` (the route's own
  catch) and `"Unexpected error"` (the generic `handleApiError` branch)

`signIn()` in `packages/auth/src/server/auth.ts` is **not** the throw site —
every DB-touching step in it is already guarded (rate limit, brute force,
user lookup, bcrypt, session rotation, and the audit-write chain all catch
internally, verified by driving them against real Postgres with a real
Ed25519 `REVEALUI_AUDIT_SIGNING_KEY` installed exactly like
`instrumentation-node.ts` does at boot).

The actual throw is in `apps/admin/src/app/api/auth/sign-in/route.ts:142`,
**after** `signIn()` already returned `success: true`:

```ts
response.cookies.set('revealui-session', result.sessionToken, {
  ...
  domain: requireSessionCookieDomain(),   // <-- throws here
});
```

[`requireSessionCookieDomain()`](apps/admin/src/lib/utils/session-cookies.ts:59)
threw whenever `SESSION_COOKIE_DOMAIN` was unset **unless** `REVEALUI_FLEET_MODE`
was also set. A plain, unlicensed self-host (no license key, no Fleet flag —
exactly the Railway marketplace-template shape) has neither, so it throws on
**every successful login**, which the outer `catch` turns into the reported
500. This only fires on success (wrong email never reaches the cookie-set
code), which is exactly the asymmetry reported.

This was drift, not a new mistake: `apps/admin/src/lib/utils/env-validation.ts`
already has an `isSaasHosted` check (via `isHostedDeployment()`) that
correctly exempts this exact deployment shape from requiring
`SESSION_COOKIE_DOMAIN` **at boot** (GAP-430/GAP-440, see the comment at
env-validation.ts:158-162 and its test file). `requireSessionCookieDomain()`
was never updated to use the same signal — it still checked only for
`REVEALUI_FLEET_MODE`, so boot-time validation passed but the request-time
check disagreed and threw anyway.

## The "missing id" clue (red herring, resolved)

`Database row missing required id field - skipping`
([safe-parse.ts](packages/core/src/database/safe-parse.ts)) fires "at startup
and during each login" because
[`universal-postgres.ts`](packages/core/src/database/universal-postgres.ts)'s
`connect()` runs a raw `SELECT 1` health check and pipes its result through
`safeParseRevealDocuments()`, which — correctly — drops the id-less
`{ "?column?": 1 }` row and logs a warning. The return value of that query is
discarded, so this is harmless noise, not the crash. Confirmed by isolating
the call: the warning fires with or without the bug present. Not fixed here
(no behavior change from it) — filing a follow-up gap for the raw-query
signature would be reasonable but out of scope for this login-breaking fix.

## Fix

`requireSessionCookieDomain()` now gates the throw on `isHostedDeployment()`
(`@revealui/core/deployment-mode`) — the **same** signal
`env-validation.ts`'s `isSaasHosted` already uses — instead of the narrower,
unrelated `REVEALUI_FLEET_MODE`-only exemption. A RevForge Fleet kit and a
plain OSS/unlicensed self-host both classify as non-hosted (`isHostedDeployment()`
is false whenever `REVEALUI_LICENSE_PRIVATE_KEY` is absent) and now correctly
fall back to a host-only cookie, matching `sessionCookieDomain()`'s existing
behavior. Only a real RevealUI Studio hosted-SaaS deployment (which carries
the private license-signing key) still requires `SESSION_COOKIE_DOMAIN`.

## Red -> green evidence (real Postgres, `pgvector/pgvector:pg16`, not PGlite)

Stood up a real `pgvector/pgvector:pg16` container, ran `drizzle-kit migrate`
against it, seeded a bootstrap admin via the **actual** onInit path (the
dynamic-SQL `create()` collections-engine operation + the typed TOS/emailVerified
Drizzle writes), installed real Postgres-backed signed audit storage exactly
like `instrumentation-node.ts` does at boot, then drove the real
`apps/admin/src/app/api/auth/sign-in/route.ts` `POST` handler end to end
(NODE_ENV=production, no `SESSION_COOKIE_DOMAIN`, no `REVEALUI_FLEET_MODE`,
no `REVEALUI_LICENSE_PRIVATE_KEY` — the exact env shape of a fresh Railway
self-host):

**Before fix (origin/test HEAD `ecc8cb751`):**
```
{"level":"error","message":"Error signing in", ...
  "stack":"Error: SESSION_COOKIE_DOMAIN env var is required in production for
  cross-subdomain auth\n    at requireSessionCookieDomain
  (.../session-cookies.ts:64:11)\n    at signInHandler (.../route.ts:142:15)..."}
{"level":"error","message":"Unexpected error", ...}
status: 500
body: {"error":"INTERNAL_ERROR","message":"An unexpected error occurred",...}
```
Wrong email in the same run: `status: 401` (`INVALID_CREDENTIALS`) — clean,
matching the report.

**After fix (this branch):**
```
status: 200
body: {"user":{"id":"rvl_...","email":"founder@revealui.com","name":"Admin User",...}}
```
Wrong email: unchanged, `status: 401`.

Also ran the full `apps/admin` auth-route + `session-cookies` Vitest suites
(174 tests, 12 test files) green, and `pnpm build` for the whole workspace
(admin app builds clean, including the touched route).

## Files

- [apps/admin/src/lib/utils/session-cookies.ts](apps/admin/src/lib/utils/session-cookies.ts) — the fix
- [apps/admin/src/__tests__/utils/session-cookies.test.ts](apps/admin/src/__tests__/utils/session-cookies.test.ts) — updated + new regression test for the plain-self-host shape

## Security note

**REVIEW-PENDING under guardrail 2 — do not merge until a non-author APPROVE
verdict is recorded.** This touches the sign-in cookie-domain gate on the
crown-jewel auth path. The fleet security-review pre-PR self-check
(`--diff origin/test`) currently reports no match against the fleet
sensitive-path list for this specific file, but the change is auth-adjacent
(session cookie scoping on the login response) and should get the same
non-author review discipline as the rest of the GAP-446 arc (#2223, #2225).

## Test plan

- [x] Reproduced red on real Postgres (`pgvector/pgvector:pg16`), not PGlite
- [x] Root-caused to `requireSessionCookieDomain()` throwing post-success
- [x] Fixed by reusing `isHostedDeployment()` (same signal as `env-validation.ts`)
- [x] Verified green on the same real-Postgres harness (200 on correct creds, 401 unchanged on wrong email)
- [x] `apps/admin` auth + session-cookies Vitest suites green (174 tests)
- [x] `pnpm build` green for the full workspace
- [ ] Owner/non-author guardrail-2 review before merge

### DIFF
--- a/apps/admin/src/__tests__/utils/session-cookies.test.ts
+++ b/apps/admin/src/__tests__/utils/session-cookies.test.ts
@@ -81,11 +81,14 @@ describe('requireSessionCookieDomain', () => {
     expect(requireSessionCookieDomain()).toBe('.revealui.com');
   });
 
-  it('throws in production when the variable is missing', () => {
+  it('throws in production on a hosted deployment when the variable is missing', () => {
     vi.stubEnv('NODE_ENV', 'production');
     vi.stubEnv('SESSION_COOKIE_DOMAIN', '');
+    // Hosted SaaS signal: REVEALUI_LICENSE_PRIVATE_KEY present (see
+    // @revealui/core/deployment-mode isHostedDeployment).
+    vi.stubEnv('REVEALUI_LICENSE_PRIVATE_KEY', 'a-private-key');
     expect(() => requireSessionCookieDomain()).toThrow(
-      'SESSION_COOKIE_DOMAIN env var is required in production for cross-subdomain auth',
+      'SESSION_COOKIE_DOMAIN env var is required in production for cross-subdomain auth on a RevealUI Studio hosted deployment',
     );
   });
 
@@ -93,6 +96,17 @@ describe('requireSessionCookieDomain', () => {
     vi.stubEnv('NODE_ENV', 'production');
     vi.stubEnv('SESSION_COOKIE_DOMAIN', '');
     vi.stubEnv('REVEALUI_FLEET_MODE', 'true');
+    expect(requireSessionCookieDomain()).toBeUndefined();
+  });
+
+  // GAP-446: a fresh, unlicensed self-host (e.g. the Railway marketplace
+  // template) has no REVEALUI_LICENSE_PRIVATE_KEY, no REVEALUI_DEPLOYMENT_MODE,
+  // and no REVEALUI_FLEET_MODE — exactly the env shape env-validation.ts's
+  // `isSaasHosted` already treats as "SESSION_COOKIE_DOMAIN not required" at
+  // boot. This must agree at request time too, or a correct login 500s.
+  it('falls back to host-only on a plain unlicensed self-host instead of throwing', () => {
+    vi.stubEnv('NODE_ENV', 'production');
+    vi.stubEnv('SESSION_COOKIE_DOMAIN', '');
     expect(requireSessionCookieDomain()).toBeUndefined();
   });
 });
--- a/apps/admin/src/lib/utils/session-cookies.ts
+++ b/apps/admin/src/lib/utils/session-cookies.ts
@@ -11,6 +11,7 @@
  * host-only delete is a browser-level no-op and the user stays signed in.
  */
 
+import { isHostedDeployment } from '@revealui/core/deployment-mode';
 import { logger } from '@revealui/utils/logger';
 import type { NextResponse } from 'next/server';
 
@@ -53,16 +54,29 @@ export function sessionCookieDomain(options?: SessionCookieDomainOptions): strin
 
 /**
  * Strict domain derivation for the sign-in session cookie: in production a
- * missing `SESSION_COOKIE_DOMAIN` throws  -  unless `REVEALUI_FLEET_MODE` is
- * set (fleet kits run host-only cookies by design).
+ * missing `SESSION_COOKIE_DOMAIN` throws, but ONLY on a RevealUI Studio
+ * hosted-SaaS deployment (`isHostedDeployment()` — admin.revealui.com +
+ * api.revealui.com sharing one auth cookie across subdomains). Every other
+ * posture — a RevForge Fleet-branded kit or a plain OSS/unlicensed self-host
+ * (GAP-430/GAP-440, no `REVEALUI_LICENSE_PRIVATE_KEY`) — runs single-host and
+ * falls back to a host-only cookie, matching `sessionCookieDomain()`.
+ *
+ * This MUST use the same `isHostedDeployment()` signal as
+ * `env-validation.ts`'s `isSaasHosted` check (which already exempts every
+ * self-host posture from requiring `SESSION_COOKIE_DOMAIN` at boot). Before
+ * GAP-446 this function used a narrower, unrelated `REVEALUI_FLEET_MODE`-only
+ * exemption that never recognized a plain self-host — so a stranger's fresh
+ * Railway deploy (no license key, no Fleet flag) passed boot-time validation
+ * but threw on every successful sign-in at request time, turning a correct
+ * login into a 500.
  */
 export function requireSessionCookieDomain(): string | undefined {
   if (process.env.NODE_ENV !== 'production') {
     return undefined;
   }
-  if (!(process.env.SESSION_COOKIE_DOMAIN || process.env.REVEALUI_FLEET_MODE)) {
+  if (isHostedDeployment(process.env) && !process.env.SESSION_COOKIE_DOMAIN) {
     throw new Error(
-      'SESSION_COOKIE_DOMAIN env var is required in production for cross-subdomain auth',
+      'SESSION_COOKIE_DOMAIN env var is required in production for cross-subdomain auth on a RevealUI Studio hosted deployment',
     );
   }
   return process.env.SESSION_COOKIE_DOMAIN || undefined;


RevealUIStudio:

Non-author cross-session review, dispatched by session `agent-system-p190995` as guardrail-2 coverage for the authoring peer. Reviewed adversarially in an independent detached worktree at the PR head (`3a558af`), not in the author's tree.

### Scope confirmed

PR head `3a558af` already contains the `origin/test` tip `ecc8cb751` (verified with `git merge-base --is-ancestor`), so this was reviewed against the branch it will actually merge into, including today's promotes. Both prior GAP-446 landings are in ancestry (`6b65286` = #2223, `fa05065` = #2225). The diff is 2 files, and `apps/admin/src/lib/utils/env-validation.ts` is byte-identical to `origin/test`, so the "same signal" claim below is checked against the real file, not the PR description.

### Attack direction 1: can a hosted production deploy now silently run without a cookie domain?

Functionally yes in one misconfigured corner, but this is **not** a session-scoping or fixation regression, because the relaxation can only ever produce a **narrower** cookie.

Both branches of `requireSessionCookieDomain()` return `process.env.SESSION_COOKIE_DOMAIN || undefined` (`apps/admin/src/lib/utils/session-cookies.ts:82`). The change only converts a `throw` into `undefined`, and `undefined` means a host-only cookie. There is no code path where this PR causes a cookie to be scoped **more broadly** than before, so session fixation, cross-subdomain leakage, and cross-tenant cookie reuse are all unreachable from this diff. The worst outcome on a hosted deploy that lost `SESSION_COOKIE_DOMAIN` is that `admin.` and `api.` stop sharing a session, which is an availability and UX failure, not a confidentiality one.

Two further mitigations that made this an approve rather than a request-changes:

1. **The throw was never a security boundary.** `requireSessionCookieDomain()` has exactly one call site in the tree, `apps/admin/src/app/api/auth/sign-in/route.ts:142`. Every other session-cookie SET site already used the non-throwing `sessionCookieDomain()`: sign-up (`sign-up/route.ts:272`), OAuth callback (`callback/[provider]/route.ts:166`), MFA verify (`mfa/verify/route.ts:113`), MFA backup (`mfa/backup/route.ts:116`), passkey authenticate (`passkey/authenticate-verify/route.ts:154`), recovery verify (`recovery/verify/route.ts:93`), verify-email (`verify-email/route.ts:134`). Six of seven session-issuing paths never had the strict gate, so it was a single-site boot-config canary, not an enforced invariant.
2. **The set/clear domain symmetry, which is the file's actual stated security invariant (`session-cookies.ts:7-11`), is preserved.** Sign-in sets with `requireSessionCookieDomain()` and `clearSessionCookies()` deletes with `sessionCookieDomain()` (`session-cookies.ts:99`). In the non-throwing path both now return the identical expression, so a sign-out still matches and deletes what sign-in set. Before this PR a self-host could not sign in at all, so there is no prior state where a set/clear mismatch is introduced.

### Attack direction 2: is the hosted-vs-self-host discriminator spoofable or env-fragile?

Not remotely spoofable. `isHostedDeployment()` (`packages/core/src/deployment-mode.ts:54`) reads only `REVEALUI_DEPLOYMENT_MODE` and `REVEALUI_LICENSE_PRIVATE_KEY` from server-side env. There is no request-derived input (no `Host`, no header, no body), so an attacker cannot flip the classification. The route is `runtime = 'nodejs'` (`sign-in/route.ts:23`), so this is a real runtime env read, not a build-time inline.

It is env-fragile in the documented sense, but not newly so. `deployment-mode.ts:4-11` explicitly warns that hosted posture must not be inferred from key presence alone, and `deploymentModeKeyConsistencyError` allows "hosted without private key" as the signer-isolation target. So a post-signer-isolation hosted admin that does not set `REVEALUI_DEPLOYMENT_MODE=hosted` classifies as `forge`. Critically, that fragility is **pre-existing and shared**: `env-validation.ts:63` computes `isSaasHosted = isHostedDeployment(process.env)` with the same default `lenient: false`, and gates the boot-time `SESSION_COOKIE_DOMAIN` requirement on it at `env-validation.ts:163-167`. This PR removes a boot-vs-request disagreement rather than creating one. Empty-string handling also matches: `isPresent()` with `lenient: false` treats `''` as absent on both sides.

### Attack direction 3: leakage into fleet kits and preview deploys

Fleet kits and previews get a host-only cookie, which is the correct and narrower posture for a single-host or `*.vercel.app` host. Preview behavior is unchanged by this diff (previews were already governed by whatever `SESSION_COOKIE_DOMAIN` is set for that env, and the return value in the non-throwing path is unchanged).

### Independent red/green reproduction

I did not take the author's word on prove-red. In my own worktree I swapped **only** the source file back to base with `git restore --source origin/test -- apps/admin/src/lib/utils/session-cookies.ts`, keeping the PR's tests, and reran:

```
× throws in production on a hosted deployment when the variable is missing
× falls back to host-only on a plain unlicensed self-host instead of throwing
  Error: SESSION_COOKIE_DOMAIN env var is required in production for cross-subdomain auth
   ❯ Module.requireSessionCookieDomain src/lib/utils/session-cookies.ts:64:11
Tests  2 failed | 10 passed (12)
```

That is the exact reported production error, reproduced from the new regression test. Restoring the fix: `Tests 12 passed (12)`. Wider run of `src/__tests__/utils`, `src/__tests__/auth`, and `src/lib/utils/__tests__/env-validation.test.ts`: **201 passed, 11 files**, so nothing else in the auth suites depended on the old throw. Worktree confirmed clean after the swap (no source modified). CI on the head commit is green across Typecheck, Build admin, Unit Tests, Integration Tests, CodeQL, and Security Gate.

### Convention compliance

Extend-before-create is satisfied: the PR reuses the `isHostedDeployment()` discriminator that #2223 already introduced in `env-validation.ts` rather than inventing a second hosted-vs-self-host signal. No new env var, no parallel predicate.

## Suggestions (non-blocking, do not hold the merge)

1. **`session-cookies.ts:77` narrows the Fleet-mode exemption from unconditional to indirect.** Boot-time requires `production && !isFleetMode && isSaasHosted` (`env-validation.ts:143` then `:163`), while request-time is now `production && isHostedDeployment`. These disagree for exactly one shape: a kit with `REVEALUI_FLEET_MODE=true` that *also* carries `REVEALUI_LICENSE_PRIVATE_KEY` or `REVEALUI_DEPLOYMENT_MODE=hosted`. Boot exempts it, request-time throws it. That shape is already declared invalid by `deploymentModeKeyConsistencyError` (`deployment-mode.ts:75`) and fails closed with a loud actionable error, which is why it is not a blocker. A belt-and-braces `|| process.env.REVEALUI_FLEET_MODE` in the condition would restore exact parity with the boot-time gate.

2. **`session-cookies.test.ts:95` is now a tautology.** "falls back to host-only under `REVEALUI_FLEET_MODE` instead of throwing" passes because no license key is stubbed, not because of the Fleet flag. Its env shape is otherwise identical to the new test at `:107`, so it adds zero coverage and would not catch the divergence in point 1. Either stub `REVEALUI_LICENSE_PRIVATE_KEY` in it (which makes it a real assertion, and red today) or retitle it to reflect what it actually exercises.

Neither suggestion is a security defect and neither should block. Merge is owner disposition.

PR #2230, opened 2026-07-27
fix(core): richtext DecoratorBlockNode createDOM + required media alt field

RevealUIStudio:
## Root causes

**Bug 1 — Lexical invariant #70 on richtext decorator nodes**
[`packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts:29-55`](packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts) — the abstract `DecoratorBlockNode` class extends Lexical's `DecoratorNode<JSX.Element>` but implemented neither `createDOM` nor `updateDOM`. Lexical's `DecoratorNode` itself does not override these either, so calls fell through to `LexicalNode`'s base implementations, which throw `createDOM: base method not extended` / `updateDOM: base method not extended` (Lexical invariant #70) the first time any subclass is reconciled to the DOM. Confirmed by reading `node_modules/.pnpm/lexical@0.46.0.../Lexical.dev.mjs` (the throwing base methods) and `@lexical/react`'s `LexicalDecoratorBlockNode.ts` (the upstream reference implementation this PR mirrors). Affects [`ImageNode`](packages/core/src/client/richtext/nodes/ImageNode.tsx) directly, and any admin custom feature node built on the same base.

**Bug 2 — POST /api/collections/media 400s on image upload**
[`packages/core/src/client/richtext/components/ImageUploadButton.tsx:96-97`](packages/core/src/client/richtext/components/ImageUploadButton.tsx) (pre-fix) built upload `FormData` with only a `file` entry. [`apps/admin/src/lib/collections/Media/index.ts:14-18`](apps/admin/src/lib/collections/Media/index.ts) declares `alt` as `required: true` on the Media collection, so every upload validation-failed with 400.

## What each fix does

- **DecoratorBlockNode**: adds `createDOM()` (returns a `div`) and `updateDOM()` (returns `false`, i.e. never recreate the element), matching `@lexical/react`'s `LexicalDecoratorBlockNode` minimal behavior exactly (no extra theme/format handling added).
- **ImageUploadButton**: derives `altText` from the filename without regex (`lastIndexOf('.')` + `slice`, extracted as the exported `deriveAltTextFromFilename` helper in `upload.ts` so it's unit-testable and shared), appends it as `formData.append('alt', altText)` before upload, and reuses the same `altText` for the `INSERT_IMAGE_COMMAND` payload — replacing the prior `file.name.replace(/\.[^/.]+$/, '')` regex (no-regex rule). Also improves the upload failure message: on `!response.ok`, tries to read a `message` field from the JSON response body and use it in the thrown `Error`, falling back to the original generic `Upload failed: ${response.statusText}` message if the body isn't JSON.

## Red → green proof

**Bug 1** — [`packages/core/src/client/richtext/nodes/__tests__/ImageNode.test.ts`](packages/core/src/client/richtext/nodes/__tests__/ImageNode.test.ts): constructs a real (headless) Lexical editor with `ImageNode` registered, creates a node via `$createImageNode`, and calls `createDOM`/`updateDOM` directly.

Before the fix, both assertions threw:
```
Error: createDOM: base method not extended
 ❯ ImageNode.createDOM .../lexical/dist/Lexical.dev.mjs:6286:7
```
After the fix: `2 passed (2)`.

**Bug 2** — extended [`packages/core/src/client/richtext/components/__tests__/upload.test.ts`](packages/core/src/client/richtext/components/__tests__/upload.test.ts) with `deriveAltTextFromFilename` unit tests and FormData-construction tests for three filename shapes (`photo.png` → `photo`, `photo` → `photo`, `.env` → `.env`).

Before the fix (function didn't exist yet):
```
TypeError: deriveAltTextFromFilename is not a function
```
6 of 12 tests in the file failed. After the fix: `12 passed (12)`.

Full suite after both fixes (via `pnpm turbo run test --filter=@revealui/core`, which builds workspace dependencies first): `115 test files passed, 2778 tests passed`.

`pnpm turbo run typecheck --filter=@revealui/core` — clean, no errors.
Biome (`biome check` on all changed files) — clean, no fixes needed.

## Security checker output

```
$ node .jv/scripts/security-pr-review-check.js --diff origin/test
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: packages/core/src/client/richtext/components/ImageUploadButton.tsx, packages/core/src/client/richtext/components/__tests__/upload.test.ts, packages/core/src/client/richtext/components/upload.ts, packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts, packages/core/src/client/richtext/nodes/__tests__/ImageNode.test.ts
(exit code 1)
```

**SECURITY-CLASSED (richtext sensitive path) — review-pending, do NOT merge until a non-author guardrail-2 verdict is recorded.**

## Known follow-ups (out of scope)

- The Content-Security-Policy `script-src` blocks `https://vercel.live` (the Vercel toolbar feedback script on deployments). That's a separate CSP-policy decision, not addressed here.
- Admin custom feature nodes (e.g. `EmbedNode`, `LabelNode`) are registered in their `feature.client.ts` files, but the core `RichTextEditor`'s `getNodesFromFeatures` never reads `feature.nodes`. This looks like a possible latent gap where custom feature nodes may not actually get registered on the editor — out of scope for this PR, flagging for follow-up investigation.

---

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -36,6 +36,7 @@
     "@types/react-dom": "catalog:",
     "@vitest/coverage-v8": "catalog:",
     "expect-type": "^1.4.0",
+    "jsdom": "catalog:",
     "tsup": "catalog:",
     "vitest": "catalog:"
   },
--- a/packages/core/src/client/richtext/components/ImageUploadButton.tsx
+++ b/packages/core/src/client/richtext/components/ImageUploadButton.tsx
@@ -11,7 +11,7 @@ import { logger } from '@revealui/core/utils/logger';
 import type React from 'react';
 import { useCallback, useRef, useState } from 'react';
 import { INSERT_IMAGE_COMMAND } from '../nodes/ImageNode.js';
-import { postMediaUpload } from './upload.js';
+import { deriveAltTextFromFilename, postMediaUpload } from './upload.js';
 
 /** Default maximum file size: 10 MB */
 const DEFAULT_MAX_FILE_SIZE_BYTES = 10 * 1024 * 1024;
@@ -92,9 +92,12 @@ export const ImageUploadButton: React.FC<ImageUploadButtonProps> = ({
           img.src = URL.createObjectURL(file);
         });
 
-        // Upload file
+        // Upload file. The admin Media collection requires `alt`, so it is
+        // derived from the filename and sent alongside the file.
+        const altText = deriveAltTextFromFilename(file.name);
         const formData = new FormData();
         formData.append('file', file);
+        formData.append('alt', altText);
 
         // Scale timeout with file size: 30s base + 10s per MB
         const timeoutMs = 30_000 + Math.ceil(file.size / (1024 * 1024)) * 10_000;
@@ -109,7 +112,16 @@ export const ImageUploadButton: React.FC<ImageUploadButtonProps> = ({
         }
 
         if (!response.ok) {
-          throw new Error(`Upload failed: ${response.statusText}`);
+          let message = `Upload failed: ${response.statusText}`;
+          try {
+            const body = (await response.json()) as { message?: unknown };
+            if (typeof body.message === 'string' && body.message.length > 0) {
+              message = body.message;
+            }
+          } catch {
+            // Body wasn't JSON - keep the generic message.
+          }
+          throw new Error(message);
         }
 
         const result = (await response.json()) as unknown;
@@ -123,7 +135,7 @@ export const ImageUploadButton: React.FC<ImageUploadButtonProps> = ({
         // Insert image into editor
         editor.dispatchCommand(INSERT_IMAGE_COMMAND, {
           src: imageUrl,
-          alt: file.name.replace(/\.[^/.]+$/, ''), // Remove extension for alt text
+          alt: altText,
           width: imageDimensions.width,
           height: imageDimensions.height,
         });
--- a/packages/core/src/client/richtext/components/__tests__/upload.test.ts
+++ b/packages/core/src/client/richtext/components/__tests__/upload.test.ts
@@ -11,7 +11,7 @@
  */
 
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import { postMediaUpload } from '../upload.js';
+import { deriveAltTextFromFilename, postMediaUpload } from '../upload.js';
 
 afterEach(() => {
   vi.unstubAllGlobals();
@@ -113,5 +113,45 @@ describe('postMediaUpload - CSRF token attach', () => {
       '/custom/upload',
       expect.objectContaining({ headers: { 'X-CSRF-Token': 'token-rotated' } }),
     );
+  });
+});
+
+describe('deriveAltTextFromFilename', () => {
+  it('strips the extension from a normal filename', () => {
+    expect(deriveAltTextFromFilename('photo.png')).toBe('photo');
+  });
+
+  it('keeps the whole name when there is no extension', () => {
+    expect(deriveAltTextFromFilename('photo')).toBe('photo');
+  });
+
+  it('keeps the whole name for a dotfile (leading dot is not an extension)', () => {
+    expect(deriveAltTextFromFilename('.env')).toBe('.env');
+  });
+});
+
+describe('image upload FormData - required alt field', () => {
+  function formDataFor(fileName: string): FormData {
+    const file = new Blob(['x'], { type: 'image/png' }) as unknown as File;
+    Object.defineProperty(file, 'name', { value: fileName });
+    const formData = new FormData();
+    formData.append('file', file);
+    formData.append('alt', deriveAltTextFromFilename(fileName));
+    return formData;
+  }
+
+  it('includes an alt entry derived from a normal filename', () => {
+    const formData = formDataFor('photo.png');
+    expect(formData.get('alt')).toBe('photo');
+  });
+
+  it('includes an alt entry equal to the whole name when there is no extension', () => {
+    const formData = formDataFor('photo');
+    expect(formData.get('alt')).toBe('photo');
+  });
+
+  it('includes an alt entry equal to the whole name for a dotfile', () => {
+    const formData = formDataFor('.env');
+    expect(formData.get('alt')).toBe('.env');
   });
 });
--- a/packages/core/src/client/richtext/components/upload.ts
+++ b/packages/core/src/client/richtext/components/upload.ts
@@ -14,6 +14,17 @@
 import { readCsrfToken } from '../../admin/utils/csrf.js';
 
 /**
+ * Derive alt text from a filename by stripping its extension. Falls back to
+ * the whole name when there is no extension (`lastIndexOf` returns -1) or
+ * the only dot is a leading dotfile marker (e.g. `.env`), where `dot > 0` is
+ * false.
+ */
+export function deriveAltTextFromFilename(fileName: string): string {
+  const dot = fileName.lastIndexOf('.');
+  return dot > 0 ? fileName.slice(0, dot) : fileName;
+}
+
+/**
  * POST `body` to `uploadEndpoint`, echoing the CSRF cookie as `X-CSRF-Token`
  * when one is readable. `Content-Type` is never set here - `fetch` derives
  * the multipart boundary from the FormData body, and an explicit value would
--- a/packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts
+++ b/packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts
@@ -45,6 +45,14 @@ export abstract class DecoratorBlockNode<
     writable.__format = format;
   }
 
+  createDOM(): HTMLElement {
+    return document.createElement('div');
+  }
+
+  updateDOM(): false {
+    return false;
+  }
+
   exportJSON(): SerializedDecoratorBlockNode<T> {
     return {
       ...super.exportJSON(),
--- a/packages/core/src/client/richtext/nodes/__tests__/ImageNode.test.ts
+++ b/packages/core/src/client/richtext/nodes/__tests__/ImageNode.test.ts
@@ -1,0 +1,62 @@
+// @vitest-environment jsdom
+/**
+ * ImageNode createDOM/updateDOM tests
+ *
+ * DecoratorBlockNode (the shared base for ImageNode and the admin app's
+ * custom feature nodes) extends Lexical's DecoratorNode without overriding
+ * createDOM/updateDOM. Lexical's DecoratorNode inherits the throwing base
+ * implementations from LexicalNode, so any reconciliation of a
+ * DecoratorBlockNode subclass throws Lexical invariant #70
+ * ("createDOM: base method not extended") the first time it is rendered.
+ * This package's default vitest environment is node (see
+ * packages/dev/src/vitest/create-config.ts), so `document` is stubbed via
+ * jsdom for this file only.
+ */
+
+import type { EditorConfig, LexicalEditor } from 'lexical';
+import { createEditor } from 'lexical';
+import { describe, expect, it } from 'vitest';
+import { $createImageNode, ImageNode } from '../ImageNode.js';
+
+const editorConfig: EditorConfig = { namespace: 'test', theme: {} };
+
+function createTestEditor(): LexicalEditor {
+  return createEditor({
+    nodes: [ImageNode],
+    onError: (error) => {
+      throw error;
+    },
+  });
+}
+
+describe('ImageNode createDOM/updateDOM', () => {
+  it('createDOM returns an HTMLElement instead of throwing invariant #70', () => {
+    const editor = createTestEditor();
+    let node: ImageNode | undefined;
+    editor.update(
+      () => {
+        node = $createImageNode({ src: 'https://example.com/photo.png', alt: 'photo' });
+      },
+      { discrete: true },
+    );
+
+    expect(node).toBeDefined();
+    const dom = (node as ImageNode).createDOM(editorConfig, editor);
+    expect(dom).toBeInstanceOf(HTMLElement);
+  });
+
+  it('updateDOM returns false (never recreates the DOM element)', () => {
+    const editor = createTestEditor();
+    let node: ImageNode | undefined;
+    editor.update(
+      () => {
+        node = $createImageNode({ src: 'https://example.com/photo.png', alt: 'photo' });
+      },
+      { discrete: true },
+    );
+
+    const imageNode = node as ImageNode;
+    const dom = imageNode.createDOM(editorConfig, editor);
+    expect(imageNode.updateDOM(imageNode, dom, editorConfig)).toBe(false);
+  });
+});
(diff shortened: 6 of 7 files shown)


RevealUIStudio:

Non-author adversarial review. Everything below was re-run independently in a detached worktree at the PR head (`c9e6dcd`), fresh `pnpm install --frozen-lockfile` (exit 0). I did not reuse the author's worktree; the one turbo result that replayed from a shared cache (`typecheck`) was re-run with `--force`.

Bug 1's root cause and fix are correct as far as they go. Bug 2's fix is correct but unguarded. One confirmed defect blocks: the base class is still missing the upstream member that makes "block" mean block, and with `createDOM` now working, that missing member starts producing structurally invalid published HTML on the exact path this PR unblocks.

---

### Independent red -> green rerun

Branch is two commits on `ecc8cb7` (`git merge-base HEAD origin/test` == `HEAD~2`), so I reverted the *source* files to the merge base and left the tests at head.

**Green at head** (`vitest run` on both new/extended files):
```
Test Files  2 passed (2)
Tests  14 passed (14)
```

**Bug 1 red** — `DecoratorBlockNode.ts` reverted to `ecc8cb7`, tests at head:
```
FAIL  ImageNode.test.ts > createDOM returns an HTMLElement instead of throwing invariant #70
Error: createDOM: base method not extended
 ❯ ImageNode.createDOM .../lexical/dist/Lexical.dev.mjs:6286:7
Test Files  1 failed (1)   Tests  2 failed (2)
```
Exact error claimed in the PR body. Confirmed.

**Bug 2 red** — `upload.ts` + `ImageUploadButton.tsx` reverted to `ecc8cb7`:
```
TypeError: deriveAltTextFromFilename is not a function
Tests  6 failed | 6 passed (12)
```
Matches the claimed 6-of-12. Confirmed.

**Adversarial mutation the PR body does not cover** — I reverted *only* `ImageUploadButton.tsx` to the merge base (i.e. deleted the actual production fix, `formData.append('alt', altText)`), keeping `upload.ts` at head:
```
grep -c "formData.append('alt'" ImageUploadButton.tsx  ->  0
Test Files  2 passed (2)
Tests  14 passed (14)
```
The bug-2 regression is **not guarded by any test**. See Blocker 2.

**Full suites** (via turbo so workspace deps build):
- `@revealui/core` — `115 test files passed, 2778 tests passed` (matches the PR body)
- `admin` (contains `RichText-serializer-xss.test.tsx`) — `159 passed | 2 skipped (161)`, `2058 passed | 13 skipped`
- `packages/core/src/richtext` + `src/__tests__/security` — `3 files, 180 tests passed`
- `typecheck --force` in my worktree — clean; `biome check` on the 5 changed source files — clean

---

## Blockers

### 1. Missing `isInline(): false` — every editor-inserted image now serializes as `paragraph > image` and publishes invalid HTML

[`packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts:48-54`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts#L48-L54)

The PR body states the added methods match `@lexical/react`'s `LexicalDecoratorBlockNode` "minimal behavior exactly (no extra theme/format handling added)". I read `@lexical/react@0.46.0`'s `src/LexicalDecoratorBlockNode.ts` in my worktree. `createDOM`/`updateDOM` are byte-equivalent, so that half of the claim holds. But upstream's class also defines `canIndent(): false` and, critically, `isInline(): false`. The local base defines neither, so it inherits `DecoratorNode.isInline()`, which returns `true` (`lexical/dist/Lexical.dev.mjs:13196-13198`).

That inherited `true` is load-bearing here. `ImagePlugin` inserts via `$insertNodeToNearestRoot` on both branches ([`ImagePlugin.tsx:46`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/plugins/ImagePlugin.tsx#L46), [`:58`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/plugins/ImagePlugin.tsx#L58)), which bottoms out in `lexical/dist/Lexical.dev.mjs:19276`:

```js
insertCaret.insert(node.isInline() ? $createParagraphNode().append(node) : node);
```

So the image is wrapped in a `ParagraphNode` instead of landing at root. I proved this with a throwaway headless probe at PR head (jsdom, real editor, real `$insertNodeToNearestRoot`, probe deleted afterwards):

```
ImageNode.isInline() = true
root first child type = paragraph
JSON = {"root":{"children":[{"type":"paragraph","children":[{"type":"image", ... }]}, ...]}}
editor DOM = <p dir="auto"><div data-lexical-decorator="true"></div>...</p>
```

Then I fed exactly that saved state to the published-render path. [`rsc.tsx:251`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/richtext/exports/server/rsc.tsx#L251) renders `<p>{children}</p>` and [`rsc.tsx:358`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/richtext/exports/server/rsc.tsx#L358) renders the image as `<figure>`. The admin serializer passes only `renderBlock`, `renderLink`, `renderUpload` ([`apps/admin/src/lib/components/RichText/index.tsx:199-203`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/apps/admin/src/lib/components/RichText/index.tsx#L199-L203)) and no `renderImage`, so the core default applies. `renderToStaticMarkup` output:

```html
<p><figure style="margin:16px 0;text-align:center"><img src="https://example.com/a.png" alt="a" .../></figure></p>
```

`<figure>` inside `<p>` is invalid HTML. The browser's parser implies a `</p>` before `<figure>`, so the client DOM is `<p></p><figure>...</figure>` while React's server tree nests them. Concrete failure scenario: an author inserts an image in the richtext editor, saves, and the published page hits a React hydration mismatch on that subtree and loses the paragraph wrapper's layout. This did not exist before because `createDOM` threw invariant #70 and images could not render at all. This PR is what makes the wrapping shippable, and every image inserted after merge persists in the wrong shape in the database.

Fix: add `isInline(): false` (and, for full upstream parity, `canIndent(): false`) to the base class, and extend `ImageNode.test.ts` with an assertion that `$insertNodeToNearestRoot($createImageNode(...))` leaves the image as a direct child of root. Please also correct the "matches upstream exactly" line in the PR body, since it is the claim that made this look already-verified.

### 2. The bug-2 fix has no regression test; the stated red -> green proof does not cover the production change

[`ImageUploadButton.tsx:97-100`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/ImageUploadButton.tsx#L97-L100) is the line that actually fixes the 400. The new tests never execute it. [`upload.test.ts:134-141`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/__tests__/upload.test.ts#L134-L141) builds its own `FormData` inside the test file and asserts on that, which is a copy of the production logic rather than a check of it.

Empirically (mutation run above): deleting `formData.append('alt', altText)` from the component leaves all 14 tests green. So the regression this PR exists to prevent is unguarded, and the PR body's "Bug 2 — red -> green proof" proves only that the helper was absent, not that the upload carries `alt`.

`jsdom` is now a devDependency of `packages/core` and `@testing-library/react` is already in the catalog, so a component-level test that renders `ImageUploadButton`, stubs `fetch`, fires a file selection, and asserts `formData.get('alt') === 'photo'` on the captured request is cheap. Given this is the security-classed richtext path and the base contract is a `required: true` field on the Media collection ([`apps/admin/src/lib/collections/Media/index.ts:14-18`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/apps/admin/src/lib/collections/Media/index.ts#L14-L18), which I verified), I am treating the missing guard as blocking rather than advisory.

---

## Adversarial angles: results

**createDOM/updateDOM parity with `@lexical/react`.** Verified against `@lexical/react@0.46.0` source in my worktree. Both methods are byte-equivalent to upstream, and upstream genuinely does no theme/format handling in `createDOM` (alignment is applied by `BlockWithAlignableContents`, not the node). No regression from omitting theme handling. The parity gap that does matter is `isInline`/`canIndent` -> Blocker 1.

**Regression risk for `EmbedNode` / `LabelNode` / other admin feature nodes.** Checked: neither `apps/admin/src/lib/features/embed/nodes/EmbedNode.tsx` nor `.../label/nodes/LabelNode.tsx` overrides `createDOM`, `updateDOM`, or `isInline`, so both inherit the new base. For them the change is a strict improvement (they would have thrown invariant #70 too). They inherit the Blocker-1 `isInline` problem identically, which is another reason to fix it in the base rather than per-node. No `LargeBodyNode` exists in the tree at this head.

**`updateDOM(): false` and stale DOM on `setData`.** Not stale, verified in the reconciler. `lexical/dist/Lexical.dev.mjs:3459` replaces the element only when `$updateDOM` returns true; returning false falls through, and at `:3544-3548` the decorator branch calls `nextNode.decorate(...)` + `reconcileDecorator(key, decorator)` on **every** dirty reconcile regardless of the `updateDOM` return. `ImageNode.setData` goes through `getWritable()`, which marks the node dirty, so `decorate()` produces a fresh `<ImageComponent data={next}/>` and React re-renders. `updateDOM(): false` is the correct choice here.

**XSS via filename-derived alt.** No injection path found; this angle is clean.
- `exportDOM` uses `img.setAttribute('alt', ...)` on a real element ([`ImageNode.tsx`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/nodes/ImageNode.tsx)) — no parsing, safe by construction.
- Every render sink passes alt as a JSX prop: `ImageNodeComponent.tsx` (`alt={data.alt || ''}`), `rsc.tsx:346/362`, `rcc.tsx:226/240`. React escapes.
- Repo-wide grep for `dangerouslySetInnerHTML` returns three hits, all inline theme/flag scripts, none touching richtext content.
- Empirical: I ran a filename-shaped payload `"><span data-x="` through `serializeLexicalState` + `renderToStaticMarkup` at PR head. Output: `alt="&quot;&gt;&lt;img src=x onerror=alert(1)&gt;&lt;span data-x=&quot;"` — fully entity-escaped, no element break-out.
- `RichText-serializer-xss.test.tsx` passes at head. Note for the record: that suite covers `src`/tag/link sanitization and does **not** assert on alt. That is acceptable given React handles it, but it means alt escaping is guarded by React, not by a repo test.

**`isSafeUrl()` protections on image src.** Untouched, confirmed. The diff is 7 files, none under `packages/security` or the serializer exports. `url-sanitization.test.ts`, `tag-and-upload-sanitization.test.ts`, and `src/__tests__/security/injection-prevention.test.ts` all pass at head (180 tests). My hostile-alt probe also shows `sanitizeUrl` still applied to `src` on the same render.

**Error-message improvement on `!response.ok`.** [`ImageUploadButton.tsx:114-124`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/ImageUploadButton.tsx#L114-L124) is safe. `response.json()` on a non-JSON or empty body rejects and is caught, preserving the generic message; the body is read once, so no double-consume. Only a `string` `message` is interpolated, so an object/array `message` cannot leak a serialized payload. The value reaches `logger.error` (client-side) and the host app's `onUploadError` callback, and `ImageUploadButton` has no other in-repo consumer, so nothing in this repo renders it as HTML. The endpoint is authenticated-only. No sensitive leak.

**`deriveAltTextFromFilename` edge cases.** [`upload.ts:22-25`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/upload.ts#L22-L25). Walked through: `photo.png`->`photo`, `photo`->`photo`, `.env`->`.env`, `photo.`->`photo`, `.`->`.`, `..`->`.`, `archive.tar.gz`->`archive.tar`, unicode `фото.png`->`фото` (`lastIndexOf('.')` has no surrogate hazard). All non-empty. The one input that returns `''` is `''` itself, which cannot reintroduce the 400 through this component: the only caller is `event.target.files?.[0]` from a native `<input type="file">`, whose `File.name` is never empty. So no reachable regression. But `deriveAltTextFromFilename` is exported from a public package surface with no floor, so a library consumer calling it with `''` gets `''` and a 400 with no signal. See Suggestions.

**jsdom devDependency, vitest pragma, lockfile.** Clean, no production dependency change.
- `jsdom` lands in `devDependencies` only, as `catalog:`, matching the 8 workspace packages already on it (`apps/docs`, `apps/marketing`, `packages/ai`, `packages/presentation`, `packages/sync`, `packages/editor`, `packages/test`).
- The lockfile diff is 3 lines resolving to `jsdom@29.0.1`, already present in the store for those 8 packages, so no new package enters the tree.
- `pnpm install --frozen-lockfile --prefer-offline` exits 0 in a clean worktree, so the lockfile is consistent with `package.json`.
- `// @vitest-environment jsdom` is file-scoped to the one new test; the package default stays `node`.

---

## Suggestions

- [`upload.ts:22-25`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/upload.ts#L22-L25) — give the exported helper a non-empty floor (`return derived || fileName || 'image'`) so no caller of this public API can produce an empty `alt` and a 400 with no diagnostic. Not reachable from the current caller, but the helper is exported precisely so others can call it.
- [`upload.test.ts:118-130`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/__tests__/upload.test.ts#L118-L130) — the helper's edge-case table covers three shapes. The interesting ones are absent: trailing dot (`photo.`), bare dot, double extension (`archive.tar.gz`), and empty string. Cheap to add alongside the Blocker-2 component test.
- The PR body's "Known follow-ups" note about `getNodesFromFeatures` never reading `feature.nodes` is a good catch and worth a gap. If custom feature nodes really are not registered, `EmbedNode`/`LabelNode` never reconcile and the Blocker-1 `isInline` consequence is currently latent for them but live for `ImageNode`. Worth confirming before that gap is scoped, because it changes the blast radius.

## Nits

- `pnpm-workspace.yaml:29` pins `jsdom: 29.1.1` while the lockfile resolves `29.0.1` for every importer, including the one this PR adds. Pre-existing repo-wide drift, not caused by this PR, but this diff is the moment it becomes visible in `packages/core`.
- [`ImageUploadButton.tsx:88`](https://github.com/RevealUIStudio/revealui/blob/c9e6dcdce/packages/core/src/client/richtext/components/ImageUploadButton.tsx#L88) — `URL.createObjectURL(file)` is never revoked. Pre-existing, out of scope, noting it since the surrounding lines were touched.

---

**Summary:** the diagnosis in this PR is accurate and the reproduction is real, but the fix is one member short of the upstream contract it claims parity with, and that shortfall is what turns a now-rendering decorator into malformed published markup. Add `isInline(): false` (plus `canIndent(): false`) with a test that pins the insertion shape, and add a component-level test that pins `alt` onto the outgoing `FormData`. I will re-review on push.

PR #2231, opened 2026-07-27
feat(server): worker liveness cron probe with private env-driven target (GAP-443)

RevealUIStudio:
## Ruling context

Owner ruling 2026-07-27 (in-session): private Vercel cron chosen over a public status-page service (Upptime) specifically to keep the Fly worker's hostname out of the public `revealui` repo. The probe target is read from `REVEALUI_WORKER_HEALTH_URL` at runtime and is never hardcoded, logged, or included in alert text.

## Design

- New route `apps/server/src/routes/cron/worker-liveness.ts`: `GET /api/cron/worker-liveness` (and `/api/v1/cron/worker-liveness`). Auth matches every other cron route in this file (`X-Cron-Secret` header, timing-safe compare against `REVEALUI_CRON_SECRET`, 401 otherwise) — same pattern as `billing-readiness.ts`.
- `REVEALUI_WORKER_HEALTH_URL` unset/empty → 200 `{skipped:true}`, one `logger.info` line, no alert. No fallback URL anywhere in code (public repo; hostname disclosure).
- Set → fetches through `createSafeFetch()` (`@revealui/security/server`, the same SSRF-guarded fetch `routes/marketplace.ts` uses: rejects private/reserved IPs, pins the resolved address, refuses redirects) with a 10s `AbortSignal.timeout`.
  - HTTP 200 → `{healthy:true, status}`, 200, no alert.
  - Non-200 / timeout / network error → fires `sendCronFailureAlert` (the same log+Sentry+email path `billing-readiness.ts` uses), `severity: 'error'`, and responds `{healthy:false, status|error}` with 503 (mirrors `billing-readiness`'s ok/503 convention). The alert `Error` message and metadata are built from a fixed classification (`timeout` / `network-error` / observed status code) — the caught error's own message is never forwarded, because both the SSRF guard and undici's network errors can embed the target hostname in their text.

## Schedule (amended mid-session)

`apps/server/vercel.json` currently has exactly **one** `crons` entry (`/api/cron/dispatch`, daily `0 6 * * *`), because this Vercel project is on the **Hobby plan**, which allows only one cron/day — `dispatch.ts` exists specifically to consolidate ~14 jobs into that single slot. Adding a second entry, or changing the schedule to sub-daily, isn't possible on this plan.

So this PR does **not** touch `vercel.json` and does **not** wire the route into `dispatch.ts` (which invokes every job via `POST`; this route is `GET`-only per spec, so it can't compose with the dispatcher's loop without changing its method-per-job).

Instead, cadence comes from `.github/workflows/worker-liveness.yml`: `schedule: '*/5 * * * *'` (GitHub's minimum granularity) + `workflow_dispatch`. One job, one `curl -H "X-Cron-Secret: ..." https://api.revealui.com/api/cron/worker-liveness` step, fails the job (`exit 1`) on any non-200 (200 only happens on `healthy:true` or `skipped:true`, so this is equivalent to failing on `healthy:false`). Mirrors the existing GAP-142 `reconciliation-crons.yml` precedent (same secret, same `vars.REVEALUI_API_URL || 'https://api.revealui.com'` pattern, same `permissions: {}`). The workflow never references the worker's hostname — it only knows the public `api.revealui.com` base and the probe's status.

## Docs

- `docs/ENVIRONMENT-VARIABLES-GUIDE.md`: added the `REVEALUI_WORKER_HEALTH_URL` row (App-Specific section).
- `scripts/sync/revvault-vercel.toml`: added `REVEALUI_WORKER_HEALTH_URL` to the `revealui-api` project's `skip` list with a comment, mirroring how `STRIPE_TAX_ENABLED` is documented as Vercel/Fly-direct (plain value, not a secret, no vault path) in `revvault-fly.toml`.

## Tests

`apps/server/src/routes/cron/worker-liveness.test.ts`, 8 cases: auth (missing secret, wrong secret, unset `REVEALUI_CRON_SECRET`) → 401; unset health URL → skipped, no alert; healthy 200 → no alert; non-200 → alert fired, hostname not in alert text; timeout → alert fired; network error → alert fired.

- **Red:** swapped in a stub implementation (bare `Hono` app returning `{healthy:true}` unconditionally, no auth/skip/alert logic) and ran the suite against it: **8/8 failed**.
- **Green:** restored the real implementation: **8/8 passed**.

Also ran the full `apps/server/src/routes/cron/` suite after the change: **114/114 passed** (no regressions). `pnpm --filter server typecheck` and `biome check` on all touched files are clean.

## Security-classed surface

Security-classed (cron route surface): needs a non-author guardrail-2 verdict before merge — review-pending.

## Owner actions (post-merge)

- Confirm the `REVEALUI_CRON_SECRET` repo Actions secret is set (it should already exist — `reconciliation-crons.yml` depends on the same secret; this workflow reuses it, no new secret is introduced).
- Set `REVEALUI_WORKER_HEALTH_URL` on the Vercel `api` project (value = the worker's public health URL), Vercel-direct.
- Redeploy `api`.


RevealUIStudio:

Non-author review of `fix/gap-443-worker-liveness-cron` @ `e14b13d`, reviewed adversarially against the six refutation targets below. Full diff read (6 files, +339/-0). No blockers found. Four suggestions and three nits, none merge-blocking.

### 1. Auth: no bypass found

`apps/server/src/routes/cron/worker-liveness.ts:45-60` is byte-for-byte the auth block of `apps/server/src/routes/cron/billing-readiness.ts:81-96` (same header read, same length pre-check, same `timingSafeEqual`, same `catch` -> 401). Verified line by line, not by eyeball similarity.

- **Fail-closed on missing config**: `!(cronSecret && provided)` -> 401. An empty-string `REVEALUI_CRON_SECRET` is falsy, so a blank secret does not become a universal key.
- **No middleware ordering to get wrong**: the check lives *inside* the handler, not in a mounted middleware, so neither mount path can be reached with the check skipped. Both mounts (`apps/server/src/index.ts:1287` `/api/cron`, `:1361` `/api/v1/cron`) point at the same Hono sub-app and therefore the same in-handler gate.
- **No competing route**: no other cron sub-app registers `/worker-liveness`, so there is no shadowing at either prefix.
- Header lookup is case-insensitive in Hono, so `X-Cron-Secret` and `x-cron-secret` both land; the explicit `|| c.req.header('x-cron-secret')` is redundant but harmless (precedent parity).

### 2. SSRF: `createSafeFetch` contract honored

Checked against the implementation at `packages/security/src/ssrf.ts:271-309`, not against the docstring.

- `safeFetch(healthUrl, { method, signal })` -> `createSafeFetch` spreads `...init` at `ssrf.ts:293`, so **the 10s `AbortSignal.timeout` is not dropped**. This was the specific failure mode I went looking for.
- Pre-flight `assertPublicUrl` (`ssrf.ts:290`) plus the connect-time revalidating dispatcher (`ssrf.ts:318-339`) means a mis-set `REVEALUI_WORKER_HEALTH_URL` pointing at `127.0.0.1`, `169.254.169.254`, `10.x`, or a rebinding host is rejected, not fetched. Non-http(s) schemes rejected at `ssrf.ts:234`.
- `redirect: 'manual'` plus the explicit throw at `ssrf.ts:300-304` means no redirect chasing.
- **The worker's response body is never read and never echoed.** The route returns only `response.status` (`worker-liveness.ts:81`, `:93`). No info leak of worker internals to the caller.

### 3. Leak paths: worker hostname and cron secret both contained

This is the constraint the PR exists to satisfy, so I traced every egress:

- **Route responses**: `{skipped}` / `{healthy,status,checkedAt}` / `{healthy,error:'timeout'|'network-error',checkedAt}`. Fixed vocabulary, no interpolation of `healthUrl`.
- **Logs**: `worker-liveness.ts:69`, `:80`, `:84`, `:101` log only literals plus `status` / `reason`.
- **Alert text**: both `new Error(...)` constructions (`:89`, `:104-108`) are built from constants and the observed status code. The caught `err.message` is never forwarded, which matters concretely: `ssrf.ts:302` embeds `target.href` **and** the redirect `Location` in its message, and `ssrf.ts:208-211` embeds the resolved hostname. Forwarding would have leaked. It doesn't.
- **Email body**: `apps/server/src/lib/cron-alerts.ts:44` renders `error.stack`. Safe here only because the `Error` is constructed inside the route, so the stack is frames and file paths, not the target. Worth knowing that this file renders stacks if anyone later passes a caught error through.
- **Audit trail**: `auditMiddleware` (`apps/server/src/middleware/audit.ts:47-80`) records method, path, status, IP, and user-agent. It does **not** record request headers, so `X-Cron-Secret` does not land in the audit log on the 401 path.
- **Workflow**: `grep` over the four touched non-test files for `fly.dev` / `fly.io` / `worker.` returns nothing. The workflow only ever names `api.revealui.com` (`worker-liveness.yml:44`); the worker hostname does not transit the runner at all. No `-v`, no `set -x`, no `--trace`; `curl -sS` only.

### 4. Workflow correctness

`.github/workflows/worker-liveness.yml` is a faithful mirror of the GAP-142 precedent (`reconciliation-crons.yml:18-63`): same `schedule` + `workflow_dispatch` shape, same `permissions: {}` (no `GITHUB_TOKEN`, no checkout step), same secret, same status-capture idiom.

- **Fails red on unhealthy**: `:59-62` exits 1 on any non-200. Since the route returns 200 only for `healthy:true` or `skipped:true`, `healthy:false` -> 503 -> red run. As intended.
- **Fails safe on missing secret**: `:47-50` exits 1 before the request is made.
- **Fails safe on curl failure**: `|| true` at `:54` yields an empty `$response`, so `$status` is empty, `!= "200"`, exit 1. A network failure to the API is not silently green.
- **Scheduled runs will actually fire**: `schedule` only runs on the default branch, and this repo's default branch is `test` (verified via the API), which is this PR's base. No dead-schedule trap.

### 5. DoS / abuse: acceptable

The unauthenticated path costs one `process.env` read, one header read, two small `Buffer.from`s, and a `timingSafeEqual`. No DB call, no I/O. The outbound probe is unreachable before auth (`healthUrl` is read at `:63`, strictly after both 401 returns), so there is no unauthenticated amplification primitive. `/api/*` is additionally covered by `tieredRateLimit` (`index.ts:567-568`).

### 6. Red/green independently reproduced

- **Green, verified locally**: `worker-liveness.test.ts` 8/8 passed; full `src/routes/cron/` suite **114/114 passed**, matching the PR's claim exactly.
- **Non-tautological, verified by mutation** (untracked `*.mut.*` copies, tracked source never modified, cleaned up on exit; worktree confirmed clean with `git status --short`):

| Mutant | Result |
|---|---|
| M1 auth block removed | RED: `returns 401 with wrong cron secret` |
| M2 skip branch removed | RED: `skips the probe ... when REVEALUI_WORKER_HEALTH_URL is unset` |
| M3 non-200 treated as healthy | RED: `fires an alert and returns 503 on a non-200 response` |
| M4 `healthUrl` interpolated into the alert `Error` | RED: `fires an alert and returns 503 on a network error` |

M4 is the one that matters for this PR's whole premise: it proves the `expect(call.error.message).not.toContain('worker.example.internal')` assertions at `worker-liveness.test.ts:118`, `:135`, `:149` are load-bearing and would actually catch a hostname regression, rather than passing vacuously.

One observation from M1: with the auth block fully removed, the missing-secret and unset-`REVEALUI_CRON_SECRET` tests still pass, because `Buffer.from(undefined)` throws into the `catch` at `:58` which returns 401 anyway. That is defense in depth working, not a weak test, but it means those two cases are covered by the `catch` rather than by the explicit guard.

---

## Suggestions (non-blocking)

- [`worker-liveness.ts:87`, `:102`](apps/server/src/routes/cron/worker-liveness.ts#L87) — `void sendCronFailureAlert(...)` is fire-and-forget and the handler returns the response on the next line. On Vercel the instance can be frozen once the response is written, so the Sentry capture and the alert email may never flush. `sendCronFailureAlert` is documented and implemented as never-throwing (`cron-alerts.ts:64-117` wraps every step in its own try/catch), so `await`ing it costs only latency on the already-failed path and cannot turn an alert failure into a 500. Note this is materially less serious than it looks because the GHA job goes red independently, so the outage is still surfaced through a second channel. Still worth the one-word change.
- [`worker-liveness.ts:98-99`](apps/server/src/routes/cron/worker-liveness.ts#L98) — every non-timeout throw collapses into `'network-error'`, including the two guard rejections (`ssrf.ts:300-304` refused redirect, `ssrf.ts:186-212` private-IP or no-public-IP). A misconfigured `REVEALUI_WORKER_HEALTH_URL` (say a plain-`http` value that 301s to https) would then page as an indistinguishable worker outage, forever, with the explanatory message deliberately suppressed. A third bucket derived from a constant prefix keeps the leak property intact: `err.message.startsWith('SSRF:') ? 'blocked-by-guard' : 'network-error'` reads only a fixed literal, never the hostname.
- [`worker-liveness.ts:81`](apps/server/src/routes/cron/worker-liveness.ts#L81) — consider `c.header('Cache-Control', 'no-store')` on the 200 path. Nothing in `index.ts` currently attaches cache headers to `/api/cron/*` and Vercel will not cache a function response without them, so this is belt-and-braces, but this is a `GET` liveness endpoint polled every 5 minutes and a cached 200 would mask exactly the outage it exists to catch.
- [`worker-liveness.yml:58`](.github/workflows/worker-liveness.yml#L58) — `echo "Body: $body"` is safe against the current response vocabulary, but this is a public repo, so every red run publicly advertises an outage window in real time. Presumably an accepted trade for free scheduling; flagging it so it is a decision rather than a side effect.

## Nits

- [`worker-liveness.yml:52`](.github/workflows/worker-liveness.yml#L52) — the secret lands in `curl`'s argv on the runner. Ephemeral single-tenant runner, and identical to the `reconciliation-crons.yml` precedent, so consistency wins here; `curl -H @-` with the header on stdin would avoid it if this idiom is ever revisited fleet-wide.
- [`worker-liveness.ts:46`](apps/server/src/routes/cron/worker-liveness.ts#L46) — `c.req.header()` is already case-insensitive, so `|| c.req.header('x-cron-secret')` is unreachable. Matches every other cron route; not worth diverging.
- [`worker-liveness.ts:55`](apps/server/src/routes/cron/worker-liveness.ts#L55) — the length pre-check short-circuits before `timingSafeEqual`, so secret length is observable. Required to avoid `timingSafeEqual`'s throw on unequal buffers, negligible in practice, and identical to `billing-readiness.ts:91`.

## Conventions

Extend-before-create satisfied: auth, alerting, and workflow scheduling all reuse existing primitives (`billing-readiness` auth block, `lib/cron-alerts.ts`, the GAP-142 workflow pattern) rather than re-authoring them. No hardcoded fallback URL anywhere, which is the owner constraint this PR is built around. Biome clean on both new files. Verdict recorded by a non-author session.

PR #2232, opened 2026-07-27
fix(server): traverse classified anchor holes (burned seqs + foreign scopes) (GAP-447)

RevealUIStudio:
## Summary

Countersigned design (owner, 2026-07-27). The audit anchor sweep's strict
+1 contiguity stalled in prod at a burned bigserial (seq 2618 absent,
owner-verified). `audit_log` is append-only (migrations 0002 + 0026:
`BEFORE UPDATE OR DELETE` raises), and `previous_signature` was abandoned
(never written — [audit-store.ts:179](packages/db/src/audit-store.ts),
spec D7.5). So absence of a row PROVES a burned seq (`nextval()` consumed,
INSERT never landed), never a deletion, and a present row scoped to a
different tenant proves non-deletion too. Both are now traversable. Only a
present, **same-scope, UNSIGNED** row still fails closed — post-GAP-417
rails this should be impossible, so it remains the one real anomaly signal.

### What changed

- **[audit-anchor-batch.ts](apps/server/src/jobs/audit-anchor-batch.ts)** —
  new pure `planTraversableBatch` (existence-classified hole traversal over
  a single classification query), `isConsecutiveFromStart` (skips that
  query in the common gapless case), `assertTraversalIntegrity` (span ==
  leaves + holes self-check, replaces `assertContiguousSeq` for this path).
  `planContiguousBatch` stays exported unchanged for compatibility.
- **[audit-anchor-sweep.ts](apps/server/src/jobs/audit-anchor-sweep.ts)** —
  switches to the new planner; fetches `SELECT seq, tenant, (signature IS
  NULL)` over `(lastAnchored, maxCandidateSeq]` only when candidates aren't
  already contiguous; records traversed holes on the anchor; logs + a new
  `revealui_audit_anchor_holes_traversed_total` counter. `SYSTEM_ANCHOR_SCOPE`
  moved to `@revealui/db/schema` (re-exported here for compatibility).
- **[schema/audit-anchors.ts](packages/db/src/schema/audit-anchors.ts)** —
  nullable `holes` jsonb column (`{ burned: number[], foreign: number }`).
  NULL means no holes traversed — every anchor written before this column
  existed reads back NULL and verifies unchanged.
- **[0031_gap447_audit_anchors_holes.sql](packages/db/migrations/0031_gap447_audit_anchors_holes.sql)** —
  hand-written `ADD COLUMN IF NOT EXISTS`, following the 0011/0015/0030
  precedent (`_custom.json` entry added, migration-journal validator green).
- **[audit-store.ts](packages/db/src/audit-store.ts)** — `DrizzleAuditStore`
  rejects any write whose `entry.tenant` is the `SYSTEM_ANCHOR_SCOPE`
  sentinel, both signed and unsigned append paths — the one door GAP-427's
  system-scope bookkeeping depends on never colliding with a real tenant.
- **[audit-merkle.ts](packages/security/src/audit-merkle.ts)** —
  `AuditAnchorSignable` gains an OPTIONAL `holes` field, covered by the root
  signature only when present. `undefined` serializes to **identical bytes**
  as before (proven by test: `auditAnchorSignableBytes({...no holes})` ===
  `auditAnchorSignableBytes({...holes: undefined})`), so every pre-existing
  anchor still verifies unchanged. `leafCount` validation now accounts for
  holes (`span - holesCount`).
- **[audit-anchor-verify.ts](packages/security/src/audit-anchor-verify.ts)** —
  `verifyAuditAnchorOffline` is now async and accepts an optional injected
  `checkBurnedSeqAbsent` callback to re-verify at verify time that recorded-
  burned seqs are **still** absent. Since `audit_log.seq` is written
  explicitly (not purely DB-generated per row — see `nextSeqValues`), a
  forged row could later fill a recorded-burned seq; that fails verification
  with a distinct "possible tamper" reason. Omitted (the true no-network/
  no-database CLI mode) skips the recheck — the pure root+inclusion-proof
  verify still runs.
- **[lib/audit-anchor-proof.ts](apps/server/src/lib/audit-anchor-proof.ts)
  + [routes/audit.ts](apps/server/src/routes/audit.ts)** — found while
  auditing every consumer of the anchor row shape (not itself in the
  original design bullets, but a real bug the change would otherwise ship
  broken): `buildAnchorInclusionProof` assumed a contiguous range
  (`leafIndex = seq - seqFrom`); a holes-carrying anchor would have
  spuriously thrown `MISSING_LEAF` or returned the wrong leaf for every
  proof request. Fixed to validate strictly-increasing, `leafCount`-length
  rows (holes are legitimate gaps, not a contiguity violation) and compute
  `leafIndex` by the row's position among the actual leaves. `serializeAnchor`
  now includes `holes` in the API response so a downloading client can
  reproduce the exact signed payload.

### Trust-statement change

Anchors now cover **every committed signed row of the scope in the
attested range**, not just a contiguous prefix. Deletion-resistance is the
append-only trigger (migrations 0002 + 0026); a burned seq is provably
non-deletion, and the traversed holes are themselves covered by the root
signature — tampering with the recorded holes invalidates the signature.

### Prove-red

```
git restore --source=origin/test -- apps/server/src/jobs/audit-anchor-sweep.ts apps/server/src/jobs/audit-anchor-batch.ts
```
→ **19 tests failed** in `audit-anchor-sweep.test.ts` (new `planTraversableBatch`
/ `isConsecutiveFromStart` / `assertTraversalIntegrity` unit tests +
GAP-447 PGlite integration tests), confirming the new tests actually
exercise new behavior, not something the base code already did.

```
git restore --source=HEAD --staged --worktree -- <same files>
```
→ full affected set green:

| Suite | Tests |
|---|---|
| `audit-anchor-sweep.test.ts` | 36 |
| `audit-anchor-proof.test.ts` | 7 |
| `audit-anchors.test.ts` | 10 |
| `audit-store.test.ts` | 11 |
| `audit-store.integration.test.ts` | 8 |
| `packages/security` (full suite) | 767 |

### Test plan

- [x] Burned seq: sign a row, `SELECT nextval(pg_get_serial_sequence('audit_log','seq'))` with no INSERT (the exact production mechanism), sign another row → sweep anchors across it, `holes.burned=[thatSeq]`, root verifies.
- [x] Cross-scope interleave: system rows with a tenant row in the middle → system scope anchors across it (`holes.foreign=1`), tenant scope still anchors its own row.
- [x] Fail closed: a same-scope UNSIGNED row inside the hole stalls the batch at that point (nothing anchored past it), warn logged naming the seq, `mGaps` incremented.
- [x] Sentinel rejection at `audit-store` (signed + unsigned writes, single + batch).
- [x] Back-compat: an anchor created WITHOUT holes (pre-change shape, and even a byte-identical `holes: undefined` vs. omitted-key comparison) still verifies; a row appearing later at a recorded-burned seq fails verification with a distinct error via the injected live-recheck callback.
- [x] Migration journal validator green (32 SQL files, 32 journal entries).
- [x] Biome clean on all changed files; `tsc --noEmit` clean on `apps/server`, `packages/db`, `packages/security`.
- [x] Pre-push gate green, including the "audit_log ONE DOOR" hard-fail check.

### Spec deviations

None from the countersigned design. One addition beyond the explicit
bullets: `audit-anchor-proof.ts` + `routes/audit.ts` needed fixing (found
via a dedicated consumer audit) so the anchors download/proof API doesn't
silently break for any anchor that actually traversed a hole — without it,
GAP-447 would ship a sweep that anchors correctly but a proof API that
can't serve inclusion proofs for those same anchors.

---

## Settlement amendment (guardrail-2 REQUEST-CHANGES, addressed in `80987d631`)

Non-author review ([comment 5095491027](https://github.com/RevealUIStudio/revealui/pull/2232#issuecomment-5095491027)) found a real defect in the original premise: **absence of a row at a seq proves non-DELETION (the append-only trigger). It does not prove non-PENDENCY.**

`DrizzleAuditStore.append`/`appendBatch` allocate `seq` via `nextval()` in a separate round trip *before* the INSERT ([audit-store.ts:159](packages/db/src/audit-store.ts), [audit-store.ts:249-266](packages/db/src/audit-store.ts)). Two concurrent writers can therefore commit out of seq order — writer A takes seq N, writer B takes seq N+1 and commits first. An unfixed sweep sees N as absent and signs it away as permanently burned the moment it queries, even though writer A's row is only in flight, not gone. Once that happens the row is orphaned forever: no future anchor ever revisits a covered range, and `buildAnchorInclusionProof`'s `row count != leafCount` check makes every proof in that anchor 409 — including for untouched leaves. The exposure is worst on the caught-up, low-volume max-lag path, which is exactly the `__system__` scope's steady state.

**Fix — settlement, not mere absence, before classifying a hole as burned** (the reviewer's option (a), the cheapest durable form, no new process state):

- New pure `settledMaxSeq` ([audit-anchor-batch.ts](apps/server/src/jobs/audit-anchor-batch.ts)): given seq-ordered rows with timestamps, returns the highest seq old enough (`>= AUDIT_ANCHOR_SETTLE_MS`, default 120s) to trust, stopping at the first row that isn't.
- `audit_log.seq` is one global sequence, so `nextval()` calls happen in strictly increasing return-value order: if seq(A) < seq(B), A's `nextval()` necessarily happened before B's. So once the highest-seq candidate in a prefix has been committed for at least the settle window, every lower seq in that prefix started its own `nextval()`→INSERT race at least that long ago too — if it still hasn't landed, it never will (burned) or it already has (and would be present, not absent).
- The sweep ([audit-anchor-sweep.ts](apps/server/src/jobs/audit-anchor-sweep.ts)) caps the working candidate set to this settled prefix *before* attempting any hole traversal. Candidates newer than the window — and any hole at or beyond them — defer to the next tick. This degrades exactly to the pre-GAP-447 self-healing stall for the narrow window where an in-flight write could still land, and has zero effect on the already-contiguous fast path (no absence is ever interpreted there, so there's nothing to misclassify).
- New env var `AUDIT_ANCHOR_SETTLE_MS` (default 120000ms), documented in the sweep header and a new `docs/ENVIRONMENT-VARIABLES-GUIDE.md` § "Audit Anchor Sweep" (which also closes a pre-existing gap: the other four `AUDIT_ANCHOR_*` vars were never documented there either).

**Regression test** reproduces the reviewer's exact hostile fixture in `audit-anchor-sweep.test.ts`: seq 1 commits, seq 2 is reserved via `nextval()` with its INSERT still in flight, seq 3 commits, a sweep pass runs in that window, and *then* seq 2 finally lands. Proved RED against the pre-fix code first (6 failures — 5 pure `settledMaxSeq` unit tests plus the PGlite fixture, which reproduced the defect log line verbatim: `anchored tenant=acct_max seq=1..3 leaves=2 holes(burned=[2] foreign=0)`), then GREEN with the fix (42/42 in the sweep test file — the sweep truncates the first pass at the settled prefix (`seq 1..1`, no holes recorded), and a later pass, once the window has passed, picks up seq 2 and seq 3 as real leaves with zero burn misattribution).

Corrected premise for the design record: the append-only trigger proves non-deletion; it does not prove non-pendency. The settlement window is what closes that gap. (Owner: amending the `.jv` spec's finding 2 wording separately to match.)

---

**Security-classed (audit anchoring surface): needs a non-author guardrail-2 verdict before merge — review-pending.**

### DIFF
--- a/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
+++ b/apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts
@@ -12,13 +12,20 @@ import {
 import { type AuditEntry, type AuditRowSignerFn, DrizzleAuditStore } from '@revealui/db';
 import type { Database } from '@revealui/db/client';
 import { auditAnchors, usageMeters } from '@revealui/db/schema';
-import { eq } from 'drizzle-orm';
+import { eq, sql } from 'drizzle-orm';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import {
   createTestDb,
   type TestDb,
 } from '../../../../../packages/test/src/utils/drizzle-test-db.js';
-import { planContiguousBatch, type SignedAuditRow } from '../audit-anchor-batch.js';
+import {
+  assertTraversalIntegrity,
+  isConsecutiveFromStart,
+  planContiguousBatch,
+  planTraversableBatch,
+  type ScopeRangeRow,
+  type SignedAuditRow,
+} from '../audit-anchor-batch.js';
 import { runAuditAnchorSweep, SYSTEM_ANCHOR_SCOPE } from '../audit-anchor-sweep.js';
 
 const KID = 'kid-s4-3';
@@ -88,6 +95,183 @@ describe('planContiguousBatch (GAP-355 S4-3)', () => {
   });
 });
 
+describe('planTraversableBatch (GAP-447 existence-classified hole traversal)', () => {
+  it('returns null for empty candidates', () => {
+    expect(planTraversableBatch(0, [], new Map())).toBeNull();
+  });
+
+  it('lastAnchoredSeq=0: permissive start — walk begins at the first candidate, not seq 1', () => {
+    const rows: SignedAuditRow[] = [{ seq: 5, signature: 's5' }];
+    const result = planTraversableBatch(0, rows, new Map());
+    expect(result).toEqual({
+      rows,
+      seqFrom: 5,
+      seqTo: 5,
+      holes: { burned: [], foreign: 0 },
+    });
+  });
+
+  it('lastAnchoredSeq=0 with internally non-contiguous candidates: gaps between them are still traversed', () => {
+    const rows: SignedAuditRow[] = [
+      { seq: 1, signature: 'a' },
+      { seq: 3, signature: 'c' },
+    ];
+    const rangeMap = new Map<number, ScopeRangeRow>([[2, { sameScope: false, signed: true }]]);
+    const result = planTraversableBatch(0, rows, rangeMap);
+    expect(result).toEqual({
+      rows,
+      seqFrom: 1,
+      seqTo: 3,
+      holes: { burned: [], foreign: 1 },
+    });
+  });
+
+  it('burned seq (absent from rangeMap) is traversable', () => {
+    const rows: SignedAuditRow[] = [
+      { seq: 1, signature: 'a' },
+      { seq: 3, signature: 'c' },
+    ];
+    const result = planTraversableBatch(0, rows, new Map()); // seq 2 absent entirely
+    expect(result).toEqual({
+      rows,
+      seqFrom: 1,
+      seqTo: 3,
+      holes: { burned: [2], foreign: 0 },
+    });
+  });
+
+  it('foreign-scope present row is traversable (count only, not enumerated)', () => {
+    const rows: SignedAuditRow[] = [
+      { seq: 1, signature: 'a' },
+      { seq: 3, signature: 'c' },
+    ];
+    const rangeMap = new Map<number, ScopeRangeRow>([[2, { sameScope: false, signed: true }]]);
+    const result = planTraversableBatch(0, rows, rangeMap);
+    expect(result?.holes).toEqual({ burned: [], foreign: 1 });
+  });
+
+  it('present + same-scope + unsigned row FAILS CLOSED and stalls before it', () => {
+    const rows: SignedAuditRow[] = [
+      { seq: 1, signature: 'a' },
+      { seq: 3, signature: 'c' },
+    ];
+    const rangeMap = new Map<number, ScopeRangeRow>([[2, { sameScope: true, signed: false }]]);
+    const result = planTraversableBatch(0, rows, rangeMap);
+    expect(result).toEqual({
+      rows: [{ seq: 1, signature: 'a' }],
+      seqFrom: 1,
+      seqTo: 1,
+      holes: { burned: [], foreign: 0 },
+      stalledAtSeq: 2,
+    });
+  });
+
+  it('stall right at the very start (no progress) still returns a result, not null', () => {
+    // lastAnchoredSeq=10 (an established anchor/floor): the walk must start
+    // exactly at seq 11, so the same-scope-unsigned row there stalls with
+    // zero rows collected, distinct from "no candidates at all" (null).
+    const rows: SignedAuditRow[] = [{ seq: 13, signature: 'c' }];
+    const rangeMap = new Map<number, ScopeRangeRow>([
+      [11, { sameScope: true, signed: false }],
+      [12, { sameScope: false, signed: true }],
+    ]);
+    const result = planTraversableBatch(10, rows, rangeMap);
+    expect(result).toEqual({
+      rows: [],
+      seqFrom: 11,
+      seqTo: 10,
+      holes: { burned: [], foreign: 0 },
+      stalledAtSeq: 11,
+    });
+  });
+
+  it('lastAnchoredSeq>0: an established anchor requires the full range attested, not just the first candidate', () => {
+    const rows: SignedAuditRow[] = [{ seq: 13, signature: 's13' }];
+    const rangeMap = new Map<number, ScopeRangeRow>([[12, { sameScope: false, signed: true }]]);
+    // seq 11 absent (burned), seq 12 foreign, seq 13 the candidate.
+    const result = planTraversableBatch(10, rows, rangeMap);
+    expect(result).toEqual({
+      rows,
+      seqFrom: 11,
+      seqTo: 13,
+      holes: { burned: [11], foreign: 1 },
+    });
+  });
+
+  it('no holes at all when candidates are already contiguous', () => {
+    const rows: SignedAuditRow[] = [
+      { seq: 11, signature: 'a' },
+      { seq: 12, signature: 'b' },
+      { seq: 13, signature: 'c' },
+    ];
+    const result = planTraversableBatch(10, rows, new Map());
+    expect(result).toEqual({ rows, seqFrom: 11, seqTo: 13, holes: { burned: [], foreign: 0 } });
+  });
+});
+
+describe('isConsecutiveFromStart', () => {
+  it('true for empty candidates', () => {
+    expect(isConsecutiveFromStart(0, [])).toBe(true);
+  });
+
+  it('true when contiguous with lastAnchoredSeq', () => {
+    expect(
+      isConsecutiveFromStart(10, [
+        { seq: 11, signature: 'a' },
+        { seq: 12, signature: 'b' },
+      ]),
+    ).toBe(true);
+  });
+
+  it('false when the first candidate does not follow lastAnchoredSeq', () => {
+    expect(isConsecutiveFromStart(10, [{ seq: 13, signature: 'a' }])).toBe(false);
+  });
+
+  it('false when candidates have an internal gap, even with lastAnchoredSeq=0', () => {
+    expect(
+      isConsecutiveFromStart(0, [
+        { seq: 1, signature: 'a' },
+        { seq: 3, signature: 'b' },
+      ]),
+    ).toBe(false);
+  });
+});
+
+describe('assertTraversalIntegrity', () => {
+  it('passes for a consistent result (span == leaves + holes)', () => {
+    expect(() =>
+      assertTraversalIntegrity({
+        rows: [{ seq: 11, signature: 'a' }],
+        seqFrom: 10,
+        seqTo: 12,
+        holes: { burned: [10], foreign: 1 },
+      }),
+    ).not.toThrow();
+  });
+
+  it('throws when the recorded holes do not account for the full span', () => {
+    expect(() =>
+      assertTraversalIntegrity({
+        rows: [{ seq: 11, signature: 'a' }],
+        seqFrom: 10,
+        seqTo: 12,
+        holes: { burned: [], foreign: 0 }, // missing 2 holes
+      }),
+    ).toThrow(/span/);
+  });
+
+  it('throws on an empty batch', () => {
+    expect(() =>
+      assertTraversalIntegrity({
+        rows: [],
+        seqFrom: 1,
+        seqTo: 1,
+        holes: { burned: [], foreign: 0 },
+      }),
+    ).toThrow(/empty batch/);
+  });
+});
+
 describe('runAuditAnchorSweep (PGlite)', () => {
   let testDb: TestDb;
   let db: Database;
@@ -119,6 +303,15 @@ describe('runAuditAnchorSweep (PGlite)', () => {
         }),
       );
     }
+  }
+
+  /**
+   * GAP-447: burn a seq value the same way production does — `nextval()` the
+   * bigserial sequence with no corresponding INSERT. The append-only trigger
+   * makes the row undeletable, so this is the only way a real seq gap forms.
+   */
+  async function burnSeq(): Promise<void> {
+    await db.execute(sql`SELECT nextval(pg_get_serial_sequence('audit_log', 'seq'))`);
   }
 
   it('inserts a verifiable anchor when batch size is reached', async () => {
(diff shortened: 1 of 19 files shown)


RevealUIStudio:

Non-author review of `5746d5cde` against the countersigned GAP-447 design. Worked from a clean worktree at the pushed commit; reproduced red/green independently; wrote a hostile fixture. Six of the seven properties I was asked to attack hold up well — the implementation is careful, the tests are substantive, and the deviation (`audit-anchor-proof.ts`) was the right call. **Property 1 does not hold**, and I can demonstrate it, so this is REQUEST-CHANGES.

---

## BLOCKER — a legitimately committed signed row can be signed away as a "burned" hole

### The premise gap

The design rests on: *absence of a row at a seq PROVES the value was burned, because the append-only trigger makes deletion impossible* (spec finding 2; [audit-anchor-batch.ts:92-95](apps/server/src/jobs/audit-anchor-batch.ts#L92-L95)).

The trigger argument rules out **deletion**. It does not rule out **pendency**. Absence at query time means "no *committed* row here *right now*", which is either *burned forever* or *allocated and still in flight*. The store makes that window structurally unavoidable:

`DrizzleAuditStore.append` calls `nextSeqValues` ([audit-store.ts:159](packages/db/src/audit-store.ts#L159), [audit-store.ts:249-266](packages/db/src/audit-store.ts#L249-L266)) as a **separate round trip**, signs, then INSERTs in a second round trip. Two concurrent appends therefore commit out of order routinely: writer A takes seq N, writer B takes seq N+1 and commits first. `appendBatch` widens this from one seq to N ([audit-store.ts:185](packages/db/src/audit-store.ts#L185)).

The classifier cannot see the difference. [audit-anchor-batch.ts:139-142](apps/server/src/jobs/audit-anchor-batch.ts#L139-L142):

```ts
const info = rangeMap.get(seq);
if (!info) {
  burned.push(seq);
  continue;
}
```

An in-flight seq is byte-identical to a burned one in `rangeMap`. It gets signed as permanently absent.

### Reproduction (hostile fixture, PGlite, run against HEAD)

Sequence: append → seq 1 commits; `nextval` takes seq 2 with the INSERT still in flight; append → seq 3 commits first; sweep runs in that window; then writer B's row commits at seq 2.

```
ANCHORS: [{"from":1,"to":3,"leaves":2,"holes":{"burned":[2],"foreign":0}}]  secondPass: 0

LIVE VERIFY: {"ok":false,"checks":[
  "root signature VALID (tenant=acct_max seq=1..3 leaves=2)",
  "no inclusion proof provided (root-only verify)",
  "burned seq 2 recorded as permanently absent, but a row now exists at that seq — anchor integrity violated (possible tamper)"]}

PROOF FOR SEQ 1: {"ok":false,"error":"row count 3 != leafCount 2","code":"MISSING_LEAF"}
```

Three consequences, all on an **honest, untampered database**:

1. **Evidence is silently omitted, permanently.** Seq 2 is a real, committed, signed, same-scope row. It sits inside `[seqFrom, seqTo]` but is not a leaf, and `lastAnchoredSeq` is now 3, so every future batch starts at 4 ([audit-anchor-sweep.ts:382-398](apps/server/src/jobs/audit-anchor-sweep.ts#L382-L398)). Second sweep pass: `anchorsInserted: 0`. That row is never rooted by any anchor, ever. This is exactly the "use holes to omit evidence" outcome, and it needs no attacker — ordinary concurrency produces it.
2. **The trust statement becomes false.** The PR claims anchors cover "every committed signed row of the scope in the attested range". Here they demonstrably do not.
3. **Verification and the proof API both break, and misattribute.** The live recheck fires "possible tamper" at an operator who has nothing wrong with their DB, and `buildAnchorInclusionProof` returns `MISSING_LEAF` for the *entire* anchor — including seq 1, an untouched leaf. Through the route that is a permanent HTTP 409 for every row in that anchor ([routes/audit.ts:277-341](apps/server/src/routes/audit.ts#L277)).

### Why this is a regression, not a pre-existing condition

The code being replaced handled this case **correctly and self-healingly**. `planContiguousBatch` returns `null` on a gap ([audit-anchor-batch.ts:70-72](apps/server/src/jobs/audit-anchor-batch.ts#L70-L72)) — the sweep stalls, retries a minute later, the in-flight row has landed, contiguity is restored, everything gets anchored. This PR converts a benign one-tick stall into a permanent, cryptographically signed misattestation. That is a net loss on the crown-jewel property.

### Exposure profile

Not uniform, but not exotic either. When a scope has a backlog larger than `batchSize`, `maxSeq` sits well behind the write frontier and the window is closed. The exposure is when a scope is **caught up** and anchors via the max-lag path ([audit-anchor-sweep.ts:469-474](apps/server/src/jobs/audit-anchor-sweep.ts#L469-L474)) — the batch tail then sits exactly at the write frontier. That is the steady state, and it is precisely the `__system__` scope's low-volume profile that motivated this gap.

### Suggested fix (preserves the design; small)

Make burned-ness require **settlement**, not just absence. Classify a seq as burned only when no transaction could still fill it. The cheapest durable form: when holes are present, cap the traversal's `maxSeq` at the highest candidate whose row is older than a settle window (env-overridable, default comfortably beyond any `nextval`→commit latency, e.g. 60s–5min), and leave anything newer for the next tick. Alternatives worth considering: re-run `fetchScopeRangeMap` once after the settle interval and require identical absence before recording `burned`; or gate on `pg_snapshot_xmin(pg_current_snapshot())` having advanced past the sweep's classification snapshot.

Either way the prod unblock still lands — seq 2618 is weeks old and settles trivially. And please amend the spec's finding 2 wording: the append-only trigger proves *non-deletion*, not *non-pendency*; the burned claim needs the settlement predicate spelled out.

Related asymmetry worth a second thought: the design fails **closed** on a same-scope unsigned row (which post-GAP-417 should be impossible) and fails **open** on an absent seq (which happens constantly). The permissive treatment landed on the riskier classification.

---

## What I verified and found sound

**2. Signature back-compat — PASS.** `holes: undefined` omits the key rather than serializing `null`, so the JCS bytes are byte-identical to the pre-change payload ([audit-merkle.ts:266-278](packages/security/src/audit-merkle.ts#L266-L278)), proven directly by the `auditAnchorSignableBytes` equality test. Existing prod anchors (2612..2617, NULL `holes`) verify unchanged. Tamper direction holds in both senses: an explicit empty `{burned:[],foreign:0}` produces different bytes than an omitted key and fails verification, and stripping/mutating a stored holes list changes the bytes *and* trips the `leafCount != span - holes` check ([audit-merkle.ts:254-262](packages/security/src/audit-merkle.ts#L254-L262)). Burned seqs are range-validated and sorted before canonicalization, so no ordering equivocation.

**3. Verify-time recheck — PASS.** `verifyBurnedSeqsStillAbsent` returns a distinct, non-generic reason and `ok:false` ([audit-anchor-verify.ts:83-118](packages/security/src/audit-anchor-verify.ts#L83-L118)); the no-checker path degrades honestly with an explicit "recheck skipped" note rather than a silent pass. My fixture exercised it against a live PGlite DB, not just a stub. (The property works — the blocker is that it will fire on honest data.)

**4. Fail closed — PASS, non-tautological.** The unsigned-row stall test builds real interleaved rows through `DrizzleAuditStore` on PGlite, establishes a genuine prior anchor first so the GAP-427 floor cannot mask the result, and asserts on the warn text, the metric path, and that `seqTo` did not advance. Verified red on base with a substantive assertion failure (`expected 1 to be 3`), not an import error.

**5. Sentinel — PASS.** `assertNotSentinelTenant` is called before any insert on all four write paths ([audit-store.ts:146-152](packages/db/src/audit-store.ts#L146), signed + unsigned, single + batch), tests assert `db.insert` was never called. `audit-log-single-door.ts` validator: PASS, 0 violations across 3478 files, door still solely `packages/db/src/audit-store.ts`. The sweep also skips a `__system__`-valued tenant defensively at [audit-anchor-sweep.ts:226](apps/server/src/jobs/audit-anchor-sweep.ts#L226).

**6. The deviation (`audit-anchor-proof.ts`) — PASS, and it was correct to fix.** The old `leafIndex = seq - seqFrom` really would have returned the wrong leaf for every holes-carrying anchor; shipping the sweep without this would have been the worse outcome. The replacement validates strictly-increasing, in-range, `leafCount`-length rows and derives the index by position ([audit-anchor-proof.ts:100-108](apps/server/src/lib/audit-anchor-proof.ts#L100-L108)). A hole seq errors cleanly as `MISSING_LEAF`, not a wrong proof. Cross-tenant isolation is intact: the proof route 404s on `anchor.tenant !== tenant` before any row fetch, and `signedRows` is re-scoped with `eq(auditLog.tenant, tenant)`, so system anchors remain unreachable (and `audit_log.tenant` can no longer be `'__system__'` at all now). `rows[built.leafIndex]` stays index-aligned with `signedRows`. `serializeAnchor` exposes only `holes`, which is already covered by the signature the client holds.

**7. Migration — PASS.** 0031 is a single `ADD COLUMN IF NOT EXISTS`, nullable, no backfill, no default. `_journal.json` idx 31 and `_custom.json` entry are consistent; `migration-journal.ts` validator: `32 SQL files, 32 journal entries, all checks passed`. Snapshot debt is pre-existing and honestly declared.

**8. Independent red/green — reproduced.**

| Run | Result |
|---|---|
| RED (`git restore --source=origin/test` on sweep + batch) | **19 failed** / 17 passed (36) — matches the PR claim; the 3 PGlite integration failures are substantive (`expected +0 to be 1`, `expected 1 to be 3`) |
| server (sweep + proof + anchors routes) | 53 passed (3 files) |
| `packages/db` (audit-store unit + integration) | 19 passed (2 files) |
| `packages/security` (full suite) | 767 passed (19 files) |
| `audit-log-single-door.ts` | PASS, 0 violations |
| `migration-journal.ts` | PASS |

Worktree restored and clean at `5746d5cde`; no `stash` used.

---

## Suggestions (non-blocking)

- **[packages/security/src/audit-merkle.ts:196-200](packages/security/src/audit-merkle.ts#L196) + [packages/db/src/schema/audit-anchors.ts:38-41](packages/db/src/schema/audit-anchors.ts#L38)** — `AuditAnchorHoles` is declared twice with subtly different types (`readonly number[]` vs `number[]`). For a field inside a signed payload, two structural definitions is one too many. Have `@revealui/db` import the security type (or vice versa) so the shape can't drift.
- **[apps/server/src/jobs/audit-anchor-sweep.ts:88-91](apps/server/src/jobs/audit-anchor-sweep.ts#L88)** — `revealui_audit_anchor_seq_gaps_total`'s help text still says "batches skipped because of a non-contiguous seq gap". Post-GAP-447 it means "same-scope unsigned row stalled traversal", which is a much stronger signal. Alert rules keyed on the old meaning will read differently now.

## Nits

- **[scripts/security/verify-audit-anchor.ts:224](scripts/security/verify-audit-anchor.ts#L224)** — `main().then((code) => process.exit(code))` has no `.catch`. `main` was synchronous before, so a throw exited via the normal path; now it becomes an unhandled rejection. Add `.catch(() => process.exit(1))` to keep the exit contract explicit. (Behaviour is still fail-closed on Node ≥15 — noting it, not blocking.)
- **[packages/security/src/audit-anchor-verify.ts:132-148](packages/security/src/audit-anchor-verify.ts#L132)** — the `try { anchorSignable = {...} } catch` block cannot throw; the comment "Validate shape via `auditAnchorSignableBytes` rules by calling verify path" describes work that actually happens in the `verifyAuditAnchorRoot` try/catch below. Pre-existing, but the block grew in this diff and the comment now actively misleads.
- **[apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts](apps/server/src/jobs/__tests__/audit-anchor-sweep.test.ts)** — the fail-closed test's own comment notes that on a scope's very first pass the GAP-427 legacy floor would jump past an unsigned row. Correctly scoped out here, but it deserves a gap of its own rather than only a test comment.

---

To be clear about the shape of this verdict: the crypto, the back-compat, the fail-closed path, the sentinel, the proof-math deviation, and the migration are all good work and I would not ask for changes to any of them. The single thing standing between this and an approve is the burned-vs-in-flight classification. A settlement predicate on the burned branch should be a contained change, and I am happy to re-review on that alone.

RevealUIStudio:

Re-review of `80987d631` (settlement amendment) as non-author. This supersedes my REQUEST-CHANGES on `5746d5cde` ([comment 5095491027](https://github.com/RevealUIStudio/revealui/pull/2232#issuecomment-5095491027)). **The round-1 blocker is fixed, correctly and minimally.** I re-ran my own hostile fixture against the fix, then spent the rest of the review trying to break the settlement premise itself. I did find two ways to defeat it, but an exhaustive trace of every audit-write call site shows neither is reachable in this codebase today. They are latent hazards, not live holes, so they belong in follow-up work rather than blocking this merge — documented in full below so the builder can act on them deliberately.

---

## 1. The round-1 blocker is closed — verified with my own fixture

I re-ran the exact round-1 scenario against the fix: seq 1 commits, seq 2 allocated via `nextval` with its INSERT still in flight, seq 3 commits first, sweep runs in that window, seq 2 commits afterward.

```
first pass  (T+10s, inside window): {"anchorsInserted":0,"tenantsWaiting":1,"errors":[]}  anchors: []
second pass (T+600s, seq 2 landed): {"anchorsInserted":1,"errors":[]}
            anchors: [{"f":1,"t":3,"l":3,"h":null}]
proof for seq 2: true
live verify ok:  true
```

Every round-1 consequence is gone. No `burned:[2]` is ever signed; the in-flight row is anchored as a **real leaf** on the next tick (`leafCount` 3, `holes` NULL); the inclusion proof for seq 2 succeeds; live verification is clean, with no false tamper. Exactly the old self-healing stall, scoped to precisely the window where an in-flight write could still land.

The mechanism is the right one. [`settledMaxSeq`](apps/server/src/jobs/audit-anchor-batch.ts#L207-L223) is a pure function, holds **zero process-local state** (so no restart caveat), and the `break`-on-first-unsettled makes it a true *prefix* bound rather than a filter — that distinction is load-bearing and it is correct. Capping happens **before** any hole traversal ([audit-anchor-sweep.ts:487-505](apps/server/src/jobs/audit-anchor-sweep.ts#L487-L505)), so no classification ever runs on unsettled data.

I also confirmed the fast-path carve-out is sound, not just convenient: if candidates are contiguous with `last`, every seq in the range is a committed visible row, so an in-flight seq inside the range is impossible by construction — it would break contiguity and route to the slow path. Skipping settlement there interprets no absence and is safe.

### Edge cases — all clean

| Case | Result |
|---|---|
| All candidates younger than the window | `waiting`, 0 anchors, `errors: []` — clean deferral, no throw |
| Same case, `maxLagMs` already satisfied | still defers — settlement correctly outranks max-lag, and **no deadlock**: the next pass past the window anchors `1..3` with `burned:[2]` |
| Hole exactly at the settled boundary (older candidates settled, newest young) | anchors `1..2` with `holes: null`, defers the hole — the conservative direction |

---

## 2. Attacking the settlement premise — two latent hazards, neither reachable today

The premise is: *`seq` is one global sequence, so if seq(A) < seq(B), A's `nextval` preceded B's; therefore once B has been settled for `settleMs`, A has had at least that long to land.* The sequence-ordering half is airtight. The half that carries the risk is the proxy for "how long ago": `settledMaxSeq` uses `audit_log.timestamp`, which is **app-supplied**, not DB-generated ([audit-store.ts](packages/db/src/audit-store.ts) writes `timestamp: entry.timestamp`).

There is a direction subtlety worth stating plainly. The entry object is constructed and *then* `append()` runs `nextval`, so `entry.timestamp ≤ nextval-instant`. The proof needs an **upper** bound on the nextval instant; the timestamp is a **lower** bound. In practice the gap is sub-millisecond against a 120s window (about six orders of magnitude of headroom), so the conclusion holds — but it holds by margin, not by construction. Both attacks below exploit that.

**ATTACK A — backdated row defeats the window.** A row stamped older than `settleMs` settles instantly. It cannot extend the prefix past a younger *lower* seq (the `break` blocks that, and my first attempt failed for exactly that reason). But when the lower seqs are already anchored, the backdated row stands alone in the candidate set:

```
anchors: [{"f":1,"t":1,"l":1,"h":null},{"f":2,"t":3,"l":1,"h":{"burned":[2],"foreign":0}}]
WINDOW DEFEATED (seq 2 signed as burned while in flight)? true
after victim row lands: totalLeaves = 2   (3 == anchored, 2 == ORPHANED)
```

That is the round-1 defect reproduced through a caller-controlled timestamp.

**ATTACK B — future-stamped row freezes a scope, silently and permanently.** `settledMaxSeq` breaks at the first row failing `nowMs - ts >= settleMs`. A timestamp in the future never satisfies it, so the prefix never advances past that seq:

```
rows: seq 1, [2 burned], 3 (stamped +10y), 4, 5
pass1: anchored [{"f":1,"t":1,"l":1,"h":null}]
pass2 (one full YEAR later): {"anchorsInserted":0,"tenantsWaiting":1,"errors":[]}
highest anchored seq: 1   → rows 3, 4, 5 and everything after are stranded forever
```

Integrity-wise this fails **closed** (nothing is misattested), which is the right direction. But it is the exact GAP-447 symptom this PR exists to cure — a permanently stalled sweep with rows stranded above it — and it is *quieter* than the original, because the deferral path deliberately emits no warn and no metric ([audit-anchor-sweep.ts:500-503](apps/server/src/jobs/audit-anchor-sweep.ts#L500)): `waiting` is indistinguishable from a normal not-ready batch.

### Why neither blocks this merge

I traced the provenance of `AuditEntry.timestamp` exhaustively across `apps/`, `packages/`, and `scripts/`. All **17** production call sites that reach the ONE DOOR are server-side `new Date()` at the append call, sub-millisecond gap. Concretely:

- **No client-influenced path.** Nothing from a request body, query, header, or Zod field reaches an audit timestamp. The main funnel is structurally immune: `AuditSystem.log()` and `logAuditEvent()` both take `Omit<AuditEvent, 'id' | 'timestamp'>`, which makes a caller-supplied timestamp *unrepresentable at the type level*.
- **No buffering, queue, outbox, batching, or retry-of-a-stale-entry** anywhere on the audit write path. `appendBatch` has zero production call sites. The one historical queue in `mcp-audit.ts` is documented as removed.
- **No backfill or importer** stamps historical timestamps.
- The nearest external value, Stripe's `event.created`, is correctly firewalled: `auditLicenseEvent` hardcodes `timestamp: new Date()` and never receives it.

So ATTACK A needs a writer that does not exist, and ATTACK B needs host clock skew above two minutes across the api/admin writer processes. Both are real hazards for the *future* of this code, not live vulnerabilities in it. That is a follow-up, not a gate.

---

## 3. Previously-approved surfaces are untouched

`80987d631` changes exactly four files: `audit-anchor-batch.ts`, `audit-anchor-sweep.ts`, the sweep test, and the env guide. `git diff 5746d5cde 80987d631 -- packages/security packages/db apps/server/src/lib/audit-anchor-proof.ts apps/server/src/routes/audit.ts` is **empty** — the JCS back-compat, tamper direction, verify-time recheck, sentinel rejection, proof-index math, and migration 0031 are byte-identical to what I verified in round 1. Re-ran anyway:

| Suite | Result |
|---|---|
| server (sweep 42 + proof 7 + anchors 10) | **59 passed** |
| `packages/db` (audit-store unit + integration) | **19 passed** |
| `packages/security` (full suite — back-compat + tamper + recheck) | **767 passed** |
| `audit-log-single-door.ts` | **PASS**, 0 violations |
| `migration-journal.ts` | **PASS**, 32/32 |

The builder's new regression fixture mirrors my round-1 scenario faithfully and was proven red against the pre-fix code. Scoping the pre-existing classification tests to `settleMs: 0` is the right call — they test classification, not timing — and the default window is covered by the new regression test plus five `settledMaxSeq` unit tests. Worktree left clean at `80987d631`; no `stash` used.

---

## Suggestions (follow-up, not blocking)

- **Enforce the invariant at the ONE DOOR.** The durable fix for both attacks is one guard beside `assertNotSentinelTenant` in [audit-store.ts](packages/db/src/audit-store.ts): reject an entry whose `timestamp` is more than a few minutes from now in either direction. Every writer already passes through that chokepoint, so a single assert closes the whole class — present writers, future writers, and clock drift — without touching 17 call sites. It also protects two other consumers that trust this field: the timestamp is inside the signed row bytes, and `packages/ai/src/audit/policy.ts` filters rate-limit windows on it. Worth a gap of its own; the settlement premise then rests on an *enforced* invariant instead of an observed convention. (The stronger variant, a DB-generated `committed_at DEFAULT now()` column settled on instead of `timestamp`, would make the proof correct by construction, since the INSERT's `now()` is strictly after the `nextval` round trip. Additive migration, same shape as 0031 — probably overkill unless the guard proves insufficient.)
- **Make the deferral observable.** [audit-anchor-sweep.ts:500-503](apps/server/src/jobs/audit-anchor-sweep.ts#L500) returns `waiting` with no warn and no metric. A scope deferring continuously for many multiples of `settleMs` is not "just needs more time" — it is stuck, and today nothing says so. A counter plus a warn once the oldest candidate exceeds, say, `10 × settleMs` would turn ATTACK B from silent to loud, and would have surfaced the original prod stall faster too.
- **State the timestamp direction in the docblock.** [`settledMaxSeq`](apps/server/src/jobs/audit-anchor-batch.ts#L184-L206) explains the sequence-ordering argument well but does not name its assumption about `row.timestamp` relative to the `nextval` instant. One sentence — "assumes the row timestamp is set within milliseconds of the `nextval()` call; a buffered or backdated writer would invalidate this bound" — protects the next person who adds an audit writer.

## Nits

- `settleMsFromEnv` accepts `n >= 0`, so `AUDIT_ANCHOR_SETTLE_MS=0` fully disables the protection and restores the round-1 vulnerability. The env guide rates it `Security: LOW`; it is the sole control preventing a permanent signed misattestation. Suggest MEDIUM, and a note that `0` is test-only.
- The env-guide row is otherwise excellent — documenting the four pre-existing sweep vars that had never been written down is a nice unprompted cleanup.
- My round-1 nits (`verify-audit-anchor.ts` missing `.catch`, the dead try/catch in `audit-anchor-verify.ts`, the duplicated `AuditAnchorHoles` interface, the stale `seq_gaps_total` help text) are unchanged and still non-blocking.

---

Good response to the review: the fix went to the root cause rather than the symptom, added no state, and came with a regression test proven red first. The two hazards I raise above are genuinely latent — I went looking for them specifically and had to reach outside the current codebase to trigger either one. Approving.

PR #2233, opened 2026-07-27
fix(core): make DecoratorBlockNode block-level, cover alt-field regression

RevealUIStudio:
## Summary

Follow-up to #2230, which merged with a **REQUEST-CHANGES verdict standing** (full evidence: https://github.com/RevealUIStudio/revealui/pull/2230#issuecomment-5095332430). This PR completes the two blockers that shipped to `test` unresolved.

### Defect 1 — `DecoratorBlockNode` reports `isInline() === true`

[`DecoratorBlockNode.ts`](packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts) extends Lexical's `DecoratorNode` without overriding `isInline()`, so it inherits the default `true`. `$insertNodeToNearestRoot` branches on `isInline()` and wraps inline nodes in a paragraph before insertion, so every editor-inserted image persisted as `root > paragraph > image` and rendered as `<p><figure></figure></p>`, invalid HTML (a `<figure>` cannot sit inside a `<p>`) that also caused a hydration mismatch.

Fix: added `isInline(): false` and `canIndent(): false`, matching `@lexical/react@0.46.0`'s `LexicalDecoratorBlockNode` (confirmed against `node_modules/@lexical/react/src/LexicalDecoratorBlockNode.ts` — those are the two methods our class was missing relative to the reference implementation).

### Defect 2 — the alt-upload fix had no mutation-sensitive test

The #2230 alt-upload fix (`formData.append('alt', altText)` in [`ImageUploadButton.tsx`](packages/core/src/client/richtext/components/ImageUploadButton.tsx)) shipped with no test that would catch its own regression. `upload.test.ts`'s `formDataFor()` helper (lines 134-141) rebuilds its own `FormData` rather than exercising the component, so the reviewer proved deleting the `append` line leaves the whole suite green.

Fix: added [`ImageUploadButton.test.tsx`](packages/core/src/client/richtext/components/__tests__/ImageUploadButton.test.tsx), which mounts the real component under `@testing-library/react` (added as a devDependency of `@revealui/core`, already in the workspace catalog), drives a real `fireEvent.change` on the hidden file input, and asserts on the `FormData` the component itself builds before it reaches the (mocked) `postMediaUpload`. `Image` and `URL.createObjectURL` are stubbed for jsdom's missing image-decode step.

### Defect 1 — red before fix, green after

Red (before adding `isInline()`/`canIndent()`):
```
FAIL  DecoratorBlockNode isInline/canIndent > inserts an ImageNode as a direct child of the root, not wrapped in a paragraph
AssertionError: expected undefined to be defined
FAIL  DecoratorBlockNode isInline/canIndent > reports isInline() as false and canIndent() as false
AssertionError: expected true to be false
```

Green (after the fix):
```
Test Files  2 passed (2)
     Tests  4 passed (4)
```
(includes the existing `ImageNode.test.ts` createDOM/updateDOM suite, unaffected.)

### Defect 2 — mutation check

With `formData.append('alt', altText)` temporarily deleted:
```
FAIL  ImageUploadButton - alt field mutation guard > sends the derived alt text in the FormData posted to postMediaUpload
AssertionError: expected null to be 'company-logo'
```

Restored, green:
```
Test Files  1 passed (1)
     Tests  1 passed (1)
```

### Full package suite / typecheck / lint

- `pnpm turbo run test --filter=@revealui/core` — 117 test files, 2781 tests, all green
- `pnpm --filter @revealui/core typecheck` — clean
- `biome check` on all changed/new files — clean

## Security review

**SECURITY-CLASSED (richtext sensitive path) — review-pending, do NOT merge until a non-author guardrail-2 verdict is recorded.**

Fleet security-path checker (internal tooling) output:
```
🔒 Current branch is SECURITY-SENSITIVE (vs origin/test). Touched: packages/core/src/client/richtext/components/__tests__/ImageUploadButton.test.tsx, packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts, packages/core/src/client/richtext/nodes/__tests__/DecoratorBlockNode.test.ts
   When you open the PR, it needs a recorded coordinator verdict before merge.
```

## Known follow-up (not migrated here)

Any image nodes saved to content between #2230's merge and this fix may already be persisted paragraph-wrapped (`root > paragraph > image`). This PR fixes new insertions only; migrating already-saved documents is a separate, deliberate follow-up decision, not bundled into this fix.

## Deviations from the task spec

- Pushing the branch required building `@revealui/harnesses` (`pnpm turbo run build --filter=@revealui/harnesses`) in addition to `@revealui/core`, because the repo's pre-push gate's "Project manager check" hard-fails when `packages/harnesses/dist/manager` is missing in a fresh worktree. Unrelated to the richtext fix; done to get a clean gate run.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@revealui/dev": "workspace:*",
+    "@testing-library/react": "catalog:",
     "@types/json-schema": "^7.0.15",
     "@types/pg": "^8.20.0",
     "@types/react": "catalog:",
--- a/packages/core/src/client/richtext/components/__tests__/ImageUploadButton.test.tsx
+++ b/packages/core/src/client/richtext/components/__tests__/ImageUploadButton.test.tsx
@@ -1,0 +1,95 @@
+// @vitest-environment jsdom
+/**
+ * ImageUploadButton mutation-sensitive regression test
+ *
+ * upload.test.ts's "required alt field" suite (see formDataFor() in that
+ * file) rebuilds its own FormData rather than exercising
+ * ImageUploadButton.tsx's actual handleFileSelect handler, so deleting the
+ * `formData.append('alt', altText)` line there leaves the whole suite green
+ * (reviewer finding on PR #2230). This test mounts the real component,
+ * drives a real file-input change event, and inspects the FormData the
+ * component itself builds and hands to postMediaUpload - so removing that
+ * line fails it.
+ */
+
+import { LexicalComposer } from '@lexical/react/LexicalComposer';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import type { Mock } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { ImageUploadButton } from '../ImageUploadButton.js';
+import { deriveAltTextFromFilename, postMediaUpload } from '../upload.js';
+
+vi.mock('../upload.js', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../upload.js')>();
+  return {
+    ...actual,
+    postMediaUpload: vi.fn(),
+  };
+});
+
+/** jsdom has no image decoder - fire `onload` as soon as `src` is set. */
+class StubImage {
+  width = 400;
+  height = 300;
+  onload: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+  #src = '';
+
+  get src(): string {
+    return this.#src;
+  }
+
+  set src(value: string) {
+    this.#src = value;
+    queueMicrotask(() => this.onload?.());
+  }
+}
+
+function renderUploadButton(): void {
+  render(
+    <LexicalComposer
+      initialConfig={{
+        namespace: 'test',
+        onError: (error) => {
+          throw error;
+        },
+      }}
+    >
+      <ImageUploadButton />
+    </LexicalComposer>,
+  );
+}
+
+describe('ImageUploadButton - alt field mutation guard', () => {
+  beforeEach(() => {
+    vi.stubGlobal('Image', StubImage);
+    vi.stubGlobal('URL', {
+      ...URL,
+      createObjectURL: vi.fn().mockReturnValue('blob:test'),
+      revokeObjectURL: vi.fn(),
+    });
+    (postMediaUpload as Mock).mockResolvedValue({
+      ok: true,
+      json: async () => ({ url: 'https://example.com/uploads/company-logo.png' }),
+    });
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  it('sends the derived alt text in the FormData posted to postMediaUpload', async () => {
+    renderUploadButton();
+
+    const input = screen.getByLabelText('Upload image') as HTMLInputElement;
+    const file = new File(['binary-image-data'], 'company-logo.png', { type: 'image/png' });
+
+    fireEvent.change(input, { target: { files: [file] } });
+
+    await waitFor(() => expect(postMediaUpload).toHaveBeenCalledTimes(1));
+
+    const formData = (postMediaUpload as Mock).mock.calls[0]?.[1] as FormData;
+    expect(formData.get('alt')).toBe(deriveAltTextFromFilename('company-logo.png'));
+  });
+});
--- a/packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts
+++ b/packages/core/src/client/richtext/nodes/DecoratorBlockNode.ts
@@ -53,6 +53,14 @@ export abstract class DecoratorBlockNode<
     return false;
   }
 
+  isInline(): false {
+    return false;
+  }
+
+  canIndent(): false {
+    return false;
+  }
+
   exportJSON(): SerializedDecoratorBlockNode<T> {
     return {
       ...super.exportJSON(),
--- a/packages/core/src/client/richtext/nodes/__tests__/DecoratorBlockNode.test.ts
+++ b/packages/core/src/client/richtext/nodes/__tests__/DecoratorBlockNode.test.ts
@@ -1,0 +1,69 @@
+// @vitest-environment jsdom
+/**
+ * DecoratorBlockNode isInline/canIndent tests
+ *
+ * DecoratorBlockNode extends Lexical's DecoratorNode without overriding
+ * isInline(), so it inherits the DecoratorNode default of `true`.
+ * $insertNodeToNearestRoot branches on isInline(): an inline node gets
+ * wrapped in a paragraph before insertion. That means every
+ * editor-inserted image persists as root > paragraph > image and renders
+ * as <p><figure><img/></figure></p>, which is invalid HTML (a <figure> is
+ * not permitted inside a <p>) and causes a hydration mismatch. Matching
+ * @lexical/react's LexicalDecoratorBlockNode (isInline(): false), the
+ * image node must land directly under the root.
+ */
+
+import { $insertNodeToNearestRoot } from '@lexical/utils';
+import type { LexicalEditor } from 'lexical';
+import { $getRoot, createEditor } from 'lexical';
+import { describe, expect, it } from 'vitest';
+import { $createImageNode, ImageNode } from '../ImageNode.js';
+
+function createTestEditor(): LexicalEditor {
+  return createEditor({
+    nodes: [ImageNode],
+    onError: (error) => {
+      throw error;
+    },
+  });
+}
+
+describe('DecoratorBlockNode isInline/canIndent', () => {
+  it('inserts an ImageNode as a direct child of the root, not wrapped in a paragraph', () => {
+    const editor = createTestEditor();
+
+    editor.update(
+      () => {
+        $getRoot().selectEnd();
+        const imageNode = $createImageNode({ src: 'x' });
+        $insertNodeToNearestRoot(imageNode);
+      },
+      { discrete: true },
+    );
+
+    editor.getEditorState().read(() => {
+      const root = $getRoot();
+      const imageNode = root
+        .getChildren()
+        .find((child): child is ImageNode => child instanceof ImageNode);
+
+      expect(imageNode).toBeDefined();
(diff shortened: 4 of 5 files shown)


RevealUIStudio:

Non-author adversarial review, same protocol as my [#2230 verdict](https://github.com/RevealUIStudio/revealui/pull/2230#issuecomment-5095332430). Fresh detached worktree at this head (`5d6e092`), clean `pnpm install --frozen-lockfile --prefer-offline` (exit 0). I did not touch the author's worktree. Every turbo result below was produced with `--force` so nothing replayed from a shared cache.

Both blockers I raised on #2230 are genuinely fixed, and I confirmed each by re-running my own probe and my own mutation rather than trusting the PR body. Scope is clean, no production dependency change, typecheck and biome clean independently. Two non-blocking items below, one of which is a required pre-merge action on the PR body rather than the code.

---

### Check 1 — my #2230 insertion probe, re-run at this head

Same probe as before: real headless Lexical editor, `ImageNode` registered, real `$insertNodeToNearestRoot`, then the resulting editor state fed through `serializeLexicalState` + `renderToStaticMarkup`.

| | #2230 head (broken) | this head |
|---|---|---|
| `ImageNode.isInline()` | `true` | **`false`** |
| root child types | `['paragraph','paragraph']` | **`['image']`** |
| image parent | `paragraph` | **`root`** |
| editor DOM | `<p dir="auto"><div data-lexical-decorator="true"></div>…</p>` | **`<div data-lexical-decorator="true"></div>`** |
| SSR html | `<p><figure …></figure></p>` | **`<figure …></figure>`** |

The `<figure>`-inside-`<p>` construction is gone, so the invalid nesting and the hydration mismatch it caused are both resolved. Blocker 1 closed.

### Check 2 — my #2230 mutation, inverted

On #2230 I deleted `formData.append('alt', altText)` from `ImageUploadButton.tsx` and the entire suite stayed green (14/14), which is what made the missing guard a blocker. I ran the identical mutation at this head:

```
FAIL  ImageUploadButton - alt field mutation guard > sends the derived alt text in the FormData posted to postMediaUpload
AssertionError: expected null to be 'company-logo'
Test Files  1 failed | 1 passed (2)
Tests  1 failed | 12 passed (13)
```

Restored, green. The guard is genuinely mutation-sensitive and it inspects the `FormData` the component itself builds (`mock.calls[0][1]`), not a test-local reconstruction. Blocker 2 closed.

### Check 2b — the PR's own defect-1 red claim, re-derived

Reverting only `DecoratorBlockNode.ts` to the merge base (`13187c5`) with the tests at head:

```
FAIL  DecoratorBlockNode isInline/canIndent > inserts an ImageNode as a direct child of the root, not wrapped in a paragraph
AssertionError: expected undefined to be defined
FAIL  DecoratorBlockNode isInline/canIndent > reports isInline() as false and canIndent() as false
AssertionError: expected true to be false
Test Files  1 failed | 1 passed (2)   Tests  2 failed | 2 passed (4)
```

Exactly the assertions quoted in the PR body, and `ImageNode.test.ts` (createDOM/updateDOM from #2230) is unaffected, as claimed.

### Check 3 — method-surface parity with `@lexical/react@0.46.0`

I diffed the full member list of the local base against `@lexical/react/src/LexicalDecoratorBlockNode.ts` in my worktree. `isInline(): false` and `canIndent(): false` were indeed the only two functionally relevant members missing, and both are now present and byte-equivalent. Remaining differences, all cosmetic:

- `afterCloneFrom(prevNode)` — upstream copies `__format` there. The local base omits it, but I checked all three subclasses and every one carries the format through its own `static clone()`: `ImageNode.tsx:68-74`, `apps/admin/.../embed/nodes/EmbedNode.tsx:59-65`, `apps/admin/.../label/nodes/LabelNode.tsx:67-73`. So alignment is preserved today. The only exposure is a future subclass whose `clone()` forgets the field. Suggestion, not a defect.
- `updateFromJSON()` — upstream's newer deserialization hook. The local nodes use `static importJSON`, which Lexical 0.46 still supports. Not a defect.
- `setFormat()` returns `void` locally vs `this` upstream. Nothing in the tree chains off it.
- `$isDecoratorBlockNode()` type guard is not exported locally.

Nothing else is missing. One honest footnote on `canIndent()`: I traced every call site across `lexical`, `@lexical/utils`, and `@lexical/rich-text`. The only one is `$handleIndentAndOutdent` (`LexicalUtils.dev.mjs:1108`), which calls it on a `parentBlock` resolved via `$findMatchingParent(node, t => $isElementNode(t) && !t.isInline())`. A `DecoratorNode` is not an `ElementNode`, so Lexical never invokes `canIndent()` on this class. Adding it is correct upstream-mirror parity and costs nothing, but it is inert in practice, which makes the second half of the new test's second assertion tautological. Nit only, called out so the coverage is not read as stronger than it is.

### Check 4 — scope creep

Clean. Single commit on `13187c5` (which is `origin/test` tip and does contain #2230), `176 insertions, 0 deletions` across exactly five files:

- `DecoratorBlockNode.ts` +8 (the two methods, nothing else)
- `DecoratorBlockNode.test.ts` +69 (new)
- `ImageUploadButton.test.tsx` +95 (new)
- `packages/core/package.json` +1 — `@testing-library/react: catalog:` under **devDependencies**
- `pnpm-lock.yaml` +3 — resolves to `16.3.2`, already in the store for other workspace packages, so no new package enters the tree

No production dependency change, no source file touched other than the 8-line base-class addition. `pnpm install --frozen-lockfile` exits 0, so the lockfile is consistent with `package.json`.

### Check 5 — everything else I flagged at this head

**`pnpm turbo run typecheck --filter=@revealui/core --force`** — clean, run in my own worktree (not a cache replay).
**`biome check`** on all four changed/new files — clean.
**Full package suite** — `117 test files passed, 2781 tests passed`, matching the PR body, but see the first suggestion below for the prerequisite the PR body omits.

---

## Required before merge (not a code defect)

- **The PR body leaks an internal repo path into a public repo.** The `## Security review` section pastes a command containing a private coordination-hub path. The repo's own scanner caught it: `Gitleaks flagged 1 match(es) across rule(s): revealui-private-repo-path`, and the `contains-sensitive-content` label is currently on this PR. Please edit the body to reference the checker generically, then drop the label. Per the bot's own note, editing does not scrub GitHub's edit history, so if the exactness matters, delete and recreate the body. I am not holding the code for this, but it should be cleared before the merge, and the fleet convention is that public-repo PR bodies never carry internal paths.

## Suggestions

- **The PR body's headline evidence command does not reproduce on a clean checkout.** `pnpm turbo run test --filter=@revealui/core` is quoted as "117 test files, 2781 tests, all green". In a fresh worktree it **fails**:

  ```
  Error: Failed to resolve import "@revealui/core/utils/logger" from
    "src/client/richtext/components/ImageUploadButton.tsx"
  Test Files  1 failed | 116 passed (117)
  Tests  2780 passed (2780)
  Failed: @revealui/core#test
  ```

  Cause: the new component test is the first test in this package to transitively pull a self-referential `@revealui/core/*` specifier, and that export maps to `./dist/utils/logger.js` (`packages/core/package.json:283-287`). `turbo.json:110-112` gives the `test` task `dependsOn: ["^build"]`, which builds *upstream* deps but never `@revealui/core` itself, and `packages/core/vitest.config.ts` aliases only `@revealui/contracts`. It passed for you because your worktree already had a `dist/` from the `@revealui/core` build you describe under "Deviations from the task spec".

  **CI is not affected** — I checked, and the Unit Tests job runs `pnpm turbo build --filter='./packages/*'` before `pnpm test` precisely for this reason (`ci.yml:381-382`, comment: "workspace tests resolve some deps from dist/"). After that build the suite is green for me too, `117 passed / 2781 passed`, exactly your numbers. So this is a local-DX and evidence-accuracy issue, not a merge risk. Either add a `@revealui/core` → `src` resolve alias in the package's vitest config, or restate the evidence line as "after `pnpm turbo build --filter='./packages/*'`".

- **Legacy paragraph-wrapped content, and why the deferral is fine.** I probed the interim shape (`root > paragraph > image`) at this head. Good news first: it loads without error (`loadError: null`), so nothing is corrupted or unopenable. But it still renders `<p><figure …></figure></p>`, so the invalid nesting persists for any document saved in that window, exactly as the "Known follow-up" section says.

  I then checked the blast radius, and it is smaller than the PR body implies: **`origin/main` does not carry #2230** (`git show origin/main:…/DecoratorBlockNode.ts` has no `createDOM`), so on production the insert path still throws invariant #70 and no production document can be paragraph-wrapped. The exposure is limited to content authored against `test`/staging between #2230's merge and this fix, and when `test` promotes to `main` both PRs travel together so production gets the correct behavior from the start. That makes deferring the migration clearly the right call. Worth recording that reasoning in the follow-up gap so a later reader does not re-litigate it.

- `afterCloneFrom` (see Check 3) — cheap to add for full upstream parity and it removes the future-subclass footgun.

## Nits

- `ImageUploadButton.test.tsx:93` asserts `formData.get('alt')` against `deriveAltTextFromFilename('company-logo.png')` rather than the literal `'company-logo'`. Both sides move together if the helper's behavior changes, so this assertion alone would not catch a derivation regression. It is adequately covered in practice because `upload.test.ts` pins the helper's outputs directly, but the literal would be strictly stronger here.
- `ImageUploadButton.test.tsx:73-77` — `vi.stubGlobal('URL', { ...URL, createObjectURL, revokeObjectURL })`. Static methods on `URL` are non-enumerable, so `{ ...URL }` spreads to `{}` and the stub replaces the global with a plain object that has no `URL` constructor. Nothing in this render path calls `new URL(...)`, so the test is correct today, but the spread reads as if it preserves the original and does not. `vi.stubGlobal('URL.createObjectURL', …)`-style targeting, or an explicit `Object.assign` onto the real `URL`, would be less fragile.
- `DecoratorBlockNode.test.ts:67` asserts `canIndent()` on a method Lexical never calls for decorator nodes (see Check 3). Harmless, just not load-bearing coverage.

---

**Summary:** this PR does exactly what it says. Both #2230 blockers are closed, and I confirmed each one by inverting my own original evidence rather than re-reading the author's. Parity with the upstream reference is complete on every member that affects behavior, scope is minimal and dev-only, and CI's build-then-test ordering means the one reproducibility gap I found never reaches the pipeline. Approving. Please redact the private path from the PR body and clear the `contains-sensitive-content` label before merging.

PR #2263, opened 2026-07-28
fix(security): the SSRF guard now says when it cannot build itself

RevealUIStudio:
⚠️ **SECURITY-SENSITIVE — review-pending, do not merge.** Touches `packages/security`. Needs a recorded guardrail-2 verdict from someone who did not author it. I wrote both this and the classifier that found it, so I must not clear it.

## What was actually wrong

A guard that fails to construct threw a **bare TypeError with no cause and no code** — indistinguishable from a network fault to every caller.

GAP-455 spent four rounds and three production deploys on this. The "worker unreachable" alarm was never about the worker. **The probe never contacted it**, which is exactly why the worker looked healthy throughout: it was.

## Located by shape, not by guessing

The deployed fingerprint returned `names:["TypeError"]` with no code. Local reproduction against the real guard showed only one site produces that signature:

| Simulated failure | Signature |
|---|---|
| **construction throws** | **`["TypeError"]`, no code** ← matches production |
| bogus dispatcher | `["TypeError","TypeError"]` |
| undefined dispatcher | no throw |
| any malformed URL | carries `ERR_INVALID_URL` |

## Changes

**1. Agent resolution handles both module shapes.** A bundler can return a CJS namespace whose exports sit under `default`, leaving `mod.Agent` undefined so `new Agent()` throws a bare TypeError. A genuinely missing module would carry `ERR_MODULE_NOT_FOUND`, so an import that *succeeds* while `Agent` is absent is the interop case, not an absent dependency.

**2. Construction failures are re-labelled** with a fixed `SSRF:`-prefixed message. Nothing from the caught error is interpolated, so no target hostname can reach a log or alert through this path. A test asserts the target does not appear in the message.

**3. Poisoned-promise fix**, found while reading: a rejected dispatcher promise was memoized and returned **forever**, so one transient construction failure broke every later call for the life of the process. It now clears and rebuilds.

## The security boundary is unchanged

It still **fails closed**. Requests do not proceed unpinned, they do not proceed at all. This makes an existing refusal *legible*; it does not relax it. The reviewer's question is whether the re-labelling can mask a real network fault — I believe not, since it wraps only the dispatcher factory, never the fetch call.

## Blast radius beyond the probe

`createSafeFetch` also backs `apps/server/src/routes/marketplace.ts` (the MCP proxy) and `packages/mcp/src/remote-client.ts`. If construction fails in that runtime, the marketplace returns 502 and marks transactions `failed` — **blaming the upstream operator for a fault on our side**. Still an inference; I have not driven those routes.

## Verification

63 tests pass, 3 new, **all proven red against the unfixed file** (reverting only `ssrf.ts` fails exactly those three).

### DIFF
--- a/packages/security/src/__tests__/ssrf.test.ts
+++ b/packages/security/src/__tests__/ssrf.test.ts
@@ -201,4 +201,59 @@ describe('createSafeFetch', () => {
     });
     await expect(safeFetch('https://1.1.1.1/')).rejects.toThrow('refusing to follow redirect');
   });
+
+  // GAP-455. A dispatcher that cannot be built used to surface as a bare
+  // TypeError with no `cause` and no `code` — indistinguishable from a network
+  // fault. A "worker unreachable" alarm was actually this, and it took four
+  // rounds and three production deploys to tell the two apart.
+  describe('when the connection guard cannot be built', () => {
+    it('reports a guard failure, not something that looks like a network fault', async () => {
+      const baseFetch = vi.fn();
+      const safeFetch = createSafeFetch({
+        baseFetch: baseFetch as unknown as typeof fetch,
+        dispatcherFactory: async () => {
+          throw new TypeError('Agent is not a constructor');
+        },
+      });
+
+      await expect(safeFetch('https://1.1.1.1/')).rejects.toThrow('SSRF:');
+      // Fails CLOSED: the request must not proceed unpinned.
+      expect(baseFetch).not.toHaveBeenCalled();
+    });
+
+    it('does not leak the target into the guard-failure message', async () => {
+      const safeFetch = createSafeFetch({
+        baseFetch: vi.fn() as unknown as typeof fetch,
+        dispatcherFactory: async () => {
+          throw new TypeError('boom');
+        },
+      });
+
+      const err = await safeFetch('https://1.1.1.1/secret-path').catch((e: Error) => e);
+      expect(err.message.startsWith('SSRF: connection guard unavailable')).toBe(true);
+      expect(err.message).not.toContain('1.1.1.1');
+      expect(err.message).not.toContain('secret-path');
+    });
+
+    it('retries construction instead of caching the rejection forever', async () => {
+      let attempts = 0;
+      const dispatcher = { sentinel: true };
+      const baseFetch = vi.fn(async () => new Response('ok', { status: 200 }));
+
+      const safeFetch = createSafeFetch({
+        baseFetch: baseFetch as unknown as typeof fetch,
+        dispatcherFactory: async () => {
+          attempts += 1;
+          if (attempts === 1) throw new TypeError('transient');
+          return dispatcher;
+        },
+      });
+
+      await expect(safeFetch('https://1.1.1.1/')).rejects.toThrow('SSRF:');
+      // Second call must rebuild. Memoizing the rejected promise previously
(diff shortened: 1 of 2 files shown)


RevealUIStudio:

Reviewed by a session that did not author this change. Read against the source, not only the tests.

### Security properties verified

**Fails closed — confirmed in source, not inferred from the test.** `await getDispatcher()` sits on the line immediately before `baseFetch`, there is exactly one call site, and no `try` wraps it. A construction failure cannot reach a socket. The test's `expect(baseFetch).not.toHaveBeenCalled()` is real but would not have been sufficient on its own; a second call site or a swallowing `catch` elsewhere would have made it a false comfort. There is neither.

**No target leak.** The thrown message is a fixed constant with nothing interpolated from `err`, so no hostname or path can reach a log or alert through it. `cause` still carries the original for diagnosis, and a construction error (`import('undici')` / `new Agent()`) does not itself contain the target.

**Concurrency is sound.** `dispatcherPromise` is assigned synchronously before any await, so concurrent callers share a single construction rather than racing to build several dispatchers. The `.catch` nulls the slot exactly once, at rejection time, and a replacement is only assigned after that — so there is no ordering in which the handler clears a *good* promise.

**The poisoned-cache fix is a genuine availability repair, not cosmetics.** Previously a single transient construction failure memoized a rejected promise and disabled every outbound `safeFetch` for the life of the process. The third test pins exactly that (`expect(attempts).toBe(2)`).

### One finding — non-blocking, but it undercuts this PR's own stated goal

The PR's rationale is that "the guard failing to build itself and the network failing are different diagnoses and must read differently." That is achieved *at the guard*. It is then collapsed again at the one consumer whose entire job is telling those apart.

`apps/server/src/routes/cron/worker-liveness.ts:126`:

```ts
if (message.startsWith('SSRF:')) sawGuardRejection = true;
```

The new message — `SSRF: connection guard unavailable — dispatcher initialization failed` — matches no entry in `GUARD_MESSAGE_TOKENS`, so it falls through to that line and classifies as **`blocked-by-guard`**.

To an operator triaging a liveness alarm, `blocked-by-guard` reads as *"the target was blocked"* and points them at the worker URL or the allowlist. The actual condition is *"the guard cannot construct itself"* — the target may be perfectly healthy. That is the same misdirection GAP-455 spent four rounds and three deploys on, moved up one level.

**No test catches this**, because `blocked-by-guard` is a valid enum value, so nothing fails — it degrades silently. That is precisely the failure mode `worker-liveness.ts`'s own header comment warns about.

Suggested follow-up, cheap and low-risk:

```ts
// ProbeFailureReason
| 'guard-unavailable'

// GUARD_MESSAGE_TOKENS
['connection guard unavailable', 'guard-unavailable'],
```

The fragment is already a fixed constant in this PR's new code, so the message coupling is no worse than what that file already accepts deliberately, and `worker-liveness.test.ts` — which drives the real guard — would lock it.

### Why this is APPROVE and not REQUEST-CHANGES

The finding is in a *different file* that this PR does not touch, and it is a diagnosability gap rather than a security regression. Every security property of the guard is preserved or improved by this change: still fails closed, still leaks nothing, and it removes a process-lifetime denial of its own outbound path. Holding a correct security fix hostage to a consumer-side classification improvement would be the wrong trade.

Recommend landing this, then the `guard-unavailable` token as a small follow-up so the diagnosis this PR creates actually survives to the operator.

PR #2353, opened 2026-08-02
fix(ci): E2E smoke API boots under NODE_ENV=development

RevealUIStudio:
## Summary

Follow-up to #2349. E2E Smoke still failed after the Prism fix because we set `NODE_ENV=production` for `pnpm --filter server start`.

**Why that breaks smoke:** `apps/server/src/index.ts` only calls `serve()` on the long-running **dev** path. Under `NODE_ENV=production`, this module is the **Vercel serverless** entry (no port bind). Fly long-running prod uses `dist/worker.js`.

Smoke needs a process listening on `:3004`, so restore `NODE_ENV=development` for the CI API start step (with the Prism-safe bundle from #2349).

**Production is unaffected** — Vercel still uses serverless `index.js`; worker still uses `worker.js`.

## Related

Unblocks promote #2344 E2E Smoke.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -780,18 +780,30 @@ jobs:
             echo "E2E_KEY_EOF"
           } >> "$GITHUB_ENV"
 
+      # Ephemeral CI secrets (not real credentials). Admin next start is
+      # production-mode and still requires the base secret trio even with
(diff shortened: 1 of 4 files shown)


RevealUIStudio:
Superseded by clean branch fix/e2e-smoke-node-env-dev (rebased onto current test; this PR was DIRTY after #2349 landed).

PR #2611, opened 2026-08-13
feat(harnesses): export runNativeSkillInvoke from content

RevealUIStudio:
## Summary

Export `runNativeSkillInvoke` from `@revealui/harnesses/content` so the AgentRuntime native-skill loop is a public SSOT. Companion revdev#402 switches the daemon RPC onto that loop class.

## Review

Harnesses export. Treat as review-pending with revdev#402.

GAP-293.


RevealUIStudio:
Companions: revdev#402 (daemon drops HTTP loop), jv#1327.

